PR #82, opened 2024-01-29
Improve wallet-api

Happy2C0de:
- Change POST to DELETE on credential endpoint

### DIFF
--- a/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/CredentialController.kt
+++ b/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/CredentialController.kt
@@ -59,7 +59,7 @@ fun Application.credentials() = walletRoute {
                 context.respond(getWalletService().getCredential(credentialId))
             }
 
-            post({
+            delete({
                 summary = "Delete a credential"
 
                 response {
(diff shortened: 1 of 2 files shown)


Happy2C0de:
This makes just no sense.

mikeplotean:
then we can revert to using the delete verb once the update to caddy in docker-compose is ready

Happy2C0de:
@waltkb @mikeplotean 
The problem isn't on nginx but on ktor CORS. See the changes. 
Can you try the setup on your side?

mikeplotean:
> @waltkb @mikeplotean The problem isn't on nginx but on ktor CORS. See the changes. Can you try the setup on your side?

this won't help, because the ktor cors plugin being used, supports these by default, please see https://ktor.io/docs/cors.html#methods
this is an nginx related issue

Happy2C0de:
> > @waltkb @mikeplotean The problem isn't on nginx but on ktor CORS. See the changes. Can you try the setup on your side?
> 
> unfortunately, this won't help, please see the link mentioned above this is an nginx related issue

The link [CORS - Methods](https://ktor.io/docs/cors.html#methods=) mentions 
> By default, the CORS plugin allows the GET, POST and HEAD HTTP methods. 

`DELETE` is not supported by default. For me, the provided PR fixed the local docker-compose setup and I can make delete requests.

mikeplotean:
yes, my mistake, you probably saw the original message
don't know how the provided pr was tested, but the record doesn't get deleted from the db

Happy2C0de:
> Hi @Happy2C0de, I'll close this PR, I've fixed this issue in above mentioned PR #122, and updated the docker ingress to also be able to utilize proper HTTP operations.

That's fine. You mitigated the problem by switching to Caddy. Thanks.

PR #95, opened 2024-01-31
fix: removing config related to ios

dinkar-jain:


dinkar-jain:
> You removed all native builds?

Only related to iPhones(IOS)

PR #103, opened 2024-01-31
Fix having multiple gradle wrappers and update the main gradle wrapper

waltkb:
Multiple gradle wrappers are still mixed in from previous non-module projects. They are removed and the main gradle wrapper is updated to 8.5.


mikeplotean (approved):
does it mean now there's a root-level (waltid-identity) gradle-wrapper.properties which is being referenced by the subprojects?

PR #273, opened 2024-03-28
Feat/ebsi issue to holder

chsavvaidis:


waltid-issuer-api/src/main/kotlin/id/walt/issuer/CIProvider.kt
```diff
@@ -108,14 +128,25 @@ open class CIProvider : OpenIDCredentialIssuer(
         return authSessions.remove(id)
     }
 
+
     // ------------------------------------------
     // Simple cryptographics operation interface implementations
-    override fun signToken(target: TokenTarget, payload: JsonObject, header: JsonObject?, keyId: String?) =
+    override fun signToken(target: TokenTarget, payload: JsonObject, header: JsonObject? , keyId: String?, privKeyJwk: String?) =
```
severinstampler, line 134: Why do we need to pass a private key as JWK string here?

waltid-openid4vc/src/commonMain/kotlin/id/walt/oid4vc/OpenID4VCI.kt
```diff
@@ -0,0 +1,161 @@
...
+        val supportedCredentials =
+            providerMetadata.credentialsSupported?.filter { !it.id.isNullOrEmpty() }?.associateBy { it.id!! } ?: mapOf()
+        return credentialOffer.credentialConfigurationIds.mapNotNull { c ->
+            supportedCredentials[c]?.let {
+                OfferedCredential.fromProviderMetadata(it)
+            }
+        }
+    }
+
+//    fun processTokenRequest(tokenRequest: TokenRequest): TokenResponse {
```
severinstampler, line 101: What's all this commented code?

PR #279, opened 2024-03-29
Passing through DidCreateOptions to the DID registration process.

alegomes:


alegomes:
If this update makes sense, we should also update the API documentation to include and explain the `"config": { /* options */ }` property.

PR #297, opened 2024-04-04
update docker compose to use postgres

mikeplotean:


ayoubagrebi (commented):
I noticed that the logs of the wallet-api container show a "Broken pipe" error

lanterno:
Thanks! I don't know why I keep running into issues with the sqlite version! I'll try this out.

PR #363, opened 2024-04-26
proposal replace mdoc kotest assertions

mikeplotean:


severinstampler (commented):
There's a TODO for replacing @BeforeAll, by the correct annotation.
Maybe @BeforeClass would be the corresponding replacement?

severinstampler (approved):

mikeplotean:
> There's a TODO for replacing @BeforeAll, by the correct annotation. Maybe @BeforeClass would be the corresponding replacement?

will use @BeforeAll from jUnit Platform. It's jUnit5's equivalent of @BeforeClass from jUnit4

mikeplotean:
obsoletes https://github.com/walt-id/waltid-identity/pull/390

severinstampler (approved):

PR #400, opened 2024-05-16
Fix/issuer api breaking changes

SuperBatata:


mikeplotean (dismissed):
please check the comments if they make sense to you, otherwise looks good

mikeplotean (approved):

PR #412, opened 2024-05-22
refactor: webwallet utils cleanup

waltkb:


waltid-wallet-api/src/main/kotlin/id/walt/webwallet/utils/JsonUtils.kt
```diff
@@ -1,7 +1,6 @@
 package id.walt.webwallet.utils
 
 import id.walt.crypto.utils.JsonUtils.toJsonElement
-import kotlinx.serialization.ExperimentalSerializationApi
```
mikeplotean, line 4: is it ok to use experimental api's (we do use them for libs, but still)?
cc @philpotisk @taminobaumann

mikeplotean (approved):

PR #437, opened 2024-05-29
Feat/DID selection during issuance in portal

dinkar-jain:


mikeplotean (approved):

mikeplotean (approved):
the did selection dropdown is also displayed in the "Verify" tab which is somewhat misleading, but otherwise looks good

PR #454, opened 2024-06-03
Major June refactoring & overhaul

waltkb:
# Major June refactoring & overhaul

This refactoring & overhaul release introduces major changes to all projects. The greater selection of them are listed here:

## Introduction of waltid-service-commons

A new module was added: "waltid-service-commons". This module holds common code shared between the walt.id services. It introduces a number improvements and new features:

### Feature system (& changed configuration system) - "No more configuration issues"
This change was introduced as many users face issues with the old configuration system, including:
- Occasionally missing configuration files (especially docker-compose) breaking the service,
- having to configure features that are not used,
- mental complexity with having such a large number of configurations,
- misconfigurations breaking production service (not being catched due to being lazy-loaded)

To resolve these issues, the feature system was introduced. The feature system operates on the basis of a number of principles and concepts:
- It applies to "Services" (runnable REST APIs) as it's part of the waltid-service-commons.
- Every service is initialized with the commons service initializer.
- The service initializer is passed one or multiple Feature Catalogs.
- A feature catalog is composed of a number of base and optional features.
    - A base feature is (basically) always enabled (not the case only in very specific niche use cases - in general one can expect a base feature to always be enabled).
    - An optional feature might be enabled or disabled. They can be:
        - An optional feature that is enabled by default: Features that do not need any configuration or have a working and useful default configuration, e.g.: Health checks (does not need configuration to be already useful)
        - An optional feature that is disabled by default: Features that do require configuration to be useful or work in the first place, e.g. Entra or OCI integration (needs access credentials, API urls, etc. - which cannot be provided by default)
- A feature can have none, one or multiple configurations associated with it.

The refactoring introducing the feature system improves upon above listed user issues by making the following changes:
- The service initializer will gather which features are explicitly enabled and which are explicitly disabled, and in addition gathers which features shall be enabled by default and which features shall be disabled by default. Accordingly, it will load the configurations of the non-disabled base features, optional features explicitly enabled, and of the optional features implicitly enabled by being feature defaulting to being enabled. This results in:
    - Less configurations required overall (as most users only use a small subset of features) and as such reducing the mental complexity of many configurations ("Where do I have to configure what now?")  and the risk of misconfigurations.
    - **No need to configure unused features**: As only configurations of enabled features are loaded, users no longer need to provide an Entra configuration even though they do not use Entra, or OCI configurations even though they do not use OCI.
- Lots of configurations (in particular the ones of base features and optional features that are enabled by default) are improved by having as many defaults as possible, making it as easy as possible for users to get started by needing them to only configure the absolute minimum.
- By having the service initializer parse all configurations required for the set of enabled features in the initialization steps, it can no longer happen that an invalid configuration is not catched before being put into production (as they mostly were lazy-loaded previously), as the mistake will already be catched when the service is started.

### CLI for services

Services get a command system. In fact, the initialization is now a command too. This can later be extended to have services have a set of commands, e.g. for migrations or configuration generation. Additionally, you can select the logging level and type from with a command line argument, see below.

### Advanced service logging

A new logging concept was introduced, as showcased on 5th of Junes dev connect meeting:
- The multiplatform libraries utilize the already used multiplatform logging library, which targets the platforms underlying common logging system (android.util.Log on Android, darwin log system on macOS / iOS, SLF4J on Java/JVM, console logger on JS, etc).
- The services receive a more advanced logging system based on the service commons module.
    - It allows for dynamic reconfiguration,
    - multiple log outputs and pushing the logs to a log aggregator system (stdout / stderr, [Seq log aggregator](https://datalust.co/seq), [Graylog log aggregator](https://graylog.org/), [Logstash](https://www.elastic.co/logstash) (part of [ELK stack](https://www.elastic.co/elastic-stack/)), [Splunk](https://www.splunk.com/)
    - multiple log output formats ([CLEF](https://clef-json.org) compact JSON format, [Elastic Common Schema](https://www.elastic.co/guide/en/ecs/current/index.html), [Elastic Common Schema for .NET](https://github.com/elastic/ecs-dotnet/blob/main/examples/aspnetcore-with-serilog/), [GELF](https://docs.graylog.org/en/latest/pages/gelf.html#gelf-payload-specification) JSON format, JSON format for Splunk HTTP Event Collector, simple & ISO 8601 rendering
    - custom logging configurations by the user (without recompiling the module)
    - structured logging (structured parcels of data instead of primitive message strings) and context logging (similar to MDC)

Configuration of the logging system can happen either with a flexible (but complex) logging configuration file (HOCON or JSON) supplied by the user, or users may choose from simplified pre-defined logging setups curated by walt.id by simply setting a command line argument. A sensible default logging setup is used if no user-provided logging configuration is detected.

### Service initialization & phase management

All services are started with a unified service initializer. This service initializer handles initializing and running the service in a standard way, instead of having this duplicated & different in each service. One can differentiate the phases: registration, loading, initialization, starting/running, ending/stopping.

### Debug endpoints & healthchecks

The service-commons provides all services with a unified set of debugging and healthcheck endpoints. The debugging endpoints provide a way of retrieving information (e.g. system properties, JVM arguments, version information, OS name / version, creating thread & heap dumps in the standard JVM format, etc.) from a running instance, which is especially useful for debugging on customer systems.

The health checks are Kuberentes compatible, and can be used for liveness checks, readiness checks, and startup checks. HTTP server and database checks are provided.

## Project hierarchy refactoring

The modules are now grouped in a project hierarchy, making the distinction between:
- waltid-libraries
    - Multiplatform libraries that developers can use to build SSI enabled use-cases
    - e.g. crypto, did, mdoc, openid4vc, sdjwt, VC
- waltid-services
    - Services with REST APIs to build SSI enabled use-cases and service-related modules
    - e.g. Issuer API, Verifier API, Wallet API, Service Commons, Service E2E Tests
- waltid-applications
    - runnable applications (not libraries) that are not services (no server or REST API character), Web frontends
    - e.g. Android demo, CLI, Web Wallet, Portal

## New E2E testing system

A new E2E testing system is introduced to resolve the shortcomings of the old one. It's changes include:
- Instead of mocking the HTTP calls, an actual HTTP server is started and the HTTP clients make real requests, which allows for catching some niche issues that would otherwise be bypassed by the old E2E testing system. This also allows for easier development, as internal HTTP clients do not have to be replaceable with a mock HTTP client anymore.
- All tests are run in the same context, which stops needing the re-contexting (recreation of the database, reregistration, reauthentication, ...) the previous system required. This makes the tests run much faster (as much less code has to be executed), and allows for much easier writing of new test cases.
- Adding new test cases is much easier than it previously was, requiring less boilerplate and unnecessary re-contexting.

## Configuration changes

 Some configurations were simplified or changed in ways to make it easier for users to grasp and configure the features associated with them. See [BREAKING_CHANGES.md](https://github.com/walt-id/waltid-identity/blob/commons/BREAKING_CHANGES.md) for more information.
 
 ## Dependency updates
 
 Dependencies of all modules were updated.
 
 #### Initial introduction of K2 / Kotlin 2.0.0
 
 The source code was updated to now be able to be compiled with Kotlin K2 compiler, which became the new default with Kotlin 2.0.0.
 
 #### Swagger improvements
 
Swagger UI library was updated to the latest release, which introduced a new JSON schema generator which is capable of producing cleaner and more predictable schemas in the spec.

### DIFF
--- a/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/policies/PresentationDefinitionPolicy.kt
+++ b/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/policies/PresentationDefinitionPolicy.kt
@@ -17,7 +17,6 @@ class PresentationDefinitionPolicy : CredentialWrapperValidatorPolicy(
 
         val requestedTypes = presentationDefinition.primitiveVerificationGetTypeList()
 
-        println(data)
         val presentedTypes =
             data.jsonObject["vp"]!!.jsonObject["verifiableCredential"]?.jsonArray?.mapNotNull {
                 it.jsonPrimitive.contentOrNull?.decodeJws()?.payload


mikeplotean:
wouldn't it make sense to have the web apps also under [applications](https://docs.walt.id/#apps)? and add another level for extensions?
so the project hierarchy would look like:
- waltid-applications
  - android
  - cli
  - web-portal
  - web-wallet
  - web3login
- waltid-extensions
  - crypto-android
  - crypto-oci

and the others:
- waltid-services
  - issuer-api
  - verifier-api
  - wallet-api
  - service-commons
  - web3-login-microservice
- waltid-libraries
  - crypto
  - did
  - mdoc
  - sdjwt
  - openid4vc
  - verifiable-credentials
  - reporting (maybe should be pulled out of the identity stack?)

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/config/ConfigManager.kt
```diff
@@ -0,0 +1,133 @@
...
+
+typealias WaltConfig = Any
+
+object ConfigManager {
+
+    val log = LoggerFactory.getLogger(ConfigManager::class.java)!!
+
+    val registeredConfigurations = ConcurrentLinkedQueue<ConfigData>()
+    val loadedConfigurations = HashMap<Pair<String, KClass<out WaltConfig>>, WaltConfig>()
+    val preloadedConfigurations = HashMap<Pair<String, KClass<out WaltConfig>>, WaltConfig>()
```
mikeplotean, line 17: since the e2e tests structure is going to change, is config preloading required?

waltid-services/waltid-wallet-api/start-db.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
```
mikeplotean, line 2: not sure it should be part of the project
I think it might be a leftover, I think it was a convenience script when api was tested against postgres, but it's not referenced anywhere in the project
so maybe this can be removed?

waltid-services/waltid-wallet-api/src/test/kotlin/E2EWalletTestLocal.kt
```diff
@@ -17,19 +14,9 @@ import io.ktor.client.request.*
...
 import kotlin.io.path.Path
 import kotlin.io.path.absolutePathString
 import kotlin.test.Test
-import kotlin.test.assertNotEquals
-import kotlin.time.Duration.Companion.seconds
-import id.walt.issuer.base.config.ConfigManager as IssuerConfigManager
-import id.walt.webwallet.config.ConfigManager as WalletConfigManager
-import id.walt.webwallet.config.WebConfig as WalletWebConfig
 
 class E2EWalletTestLocal : E2EWalletTestBase() {
```
mikeplotean, line 21: since we're moving to a new e2e test structure, maybe this can be removed and have the e2e tests added with another pr?

waltid-services/waltid-wallet-api/src/test/kotlin/E2EWalletTestBase.kt
```diff
@@ -1,7 +1,6 @@
 import id.walt.crypto.keys.KeyGenerationRequest
 import id.walt.crypto.keys.KeyType
-import id.walt.issuer.IssuanceExamples.issuerOnboardingRequestDefaultExample
-import id.walt.issuer.IssuerOnboardingResponse
+import id.walt.issuer.issuance.IssuerOnboardingResponse
 import id.walt.webwallet.db.models.AccountWalletListing
```
mikeplotean, line 4: since we're moving to a new e2e test structure, maybe this class can be removed and have the e2e tests added with another pr?

waltid-services/waltid-wallet-api/src/main/resources/example_klogging.json
```diff
@@ -0,0 +1,20 @@
+{
```
mikeplotean, line 1: this looks different from klogging.conf inside the test folder
is it on purpose?

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/AuthController.kt
```diff
@@ -123,17 +125,19 @@ fun Application.configureSecurity() {
...
         }
 
-        jwt("auth-oauth-jwt") {
-            realm = OidcLoginService.oidcRealm
-            // verifier(jwkProvider, oidcRealm)
-            verifier(OidcLoginService.jwkProvider)
+        if (FeatureManager.isFeatureEnabled(FeatureCatalog.oidcAuthenticationFeature)) {
+            jwt("auth-oauth-jwt") {
+                realm = OidcLoginService.oidcRealm
+                // verifier(jwkProvider, oidcRealm)
```
mikeplotean, line 134: since the file was reformatted, maybe the commented code can also be cleaned up?

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -441,12 +438,18 @@ class SSIKit2WalletService(
...
 
     override suspend fun generateKey(request: KeyGenerationRequest): String = let {
-        if (request.backend == "oci-rest-api" && request.config == null) {
+        if (request.config == null) {
+            ConfigManager.getConfig<KeyGenerationDefaultsConfig>().getConfigForBackend(request.backend)?.let {
+                request.config = it
+            }
+        }
+
+        /*if (request.backend == "oci-rest-api" && request.config == null) {
```
mikeplotean, line 447: maybe this commented code can be removed?

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/OidcLoginService.kt
```diff
@@ -0,0 +1,22 @@
+package id.walt.webwallet.service
+
+import com.auth0.jwk.JwkProvider
+import com.auth0.jwk.JwkProviderBuilder
+import id.walt.webwallet.service.WalletServiceManager.oidcConfig
+import java.net.URL
+import java.util.concurrent.TimeUnit
+
+object OidcLoginService {
```
mikeplotean, line 9: not sure it's a service, doesn't do anything but property initialization
maybe it can be renamed to smth. like OidcLoginService-/-Configuration or -Data

mikeplotean (changes requested):

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/featureflag/FeatureManager.kt
```diff
@@ -0,0 +1,155 @@
...
+import id.walt.commons.config.runconfig.RunConfiguration
+import io.klogging.noCoLogger
+
+object FeatureManager {
+    val baseFeatures = HashSet<String>()
+    val enabledFeatures = HashSet<String>()
+    val disabledFeatures = HashSet<String>()
+    val registeredFeatures = HashMap<String, AbstractFeature>()
+
+    private val log = noCoLogger("FeatureManager")
```
mikeplotean, line 13: maybe this can be renamed to `logger`?

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/config/ConfigManager.kt
```diff
@@ -0,0 +1,133 @@
...
+import io.klogging.java.LoggerFactory
+import java.util.concurrent.ConcurrentLinkedQueue
+import kotlin.reflect.KClass
+import kotlin.reflect.jvm.jvmName
+
+typealias WaltConfig = Any
+
+object ConfigManager {
+
+    val log = LoggerFactory.getLogger(ConfigManager::class.java)!!
```
mikeplotean, line 13: maybe this can be renamed to `logger` and be private?

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/config/ConfigManager.kt
```diff
@@ -0,0 +1,133 @@
...
+                        )
+                }
+        }
+
+    fun registerConfig(
+        id: String,
+        type: KClass<out WaltConfig>,
+    ) = registerConfig(ConfigData(id, type/*, false, multiple, onLoad*/))
+
+    fun registerRequiredConfig(
```
mikeplotean, line 96: is it still relevant, seems it has no references?

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/config/ConfigManager.kt
```diff
@@ -0,0 +1,133 @@
...
+                        ?: throw IllegalArgumentException(
+                            "Invalid config class type: \"${loadedConfig::class.jvmName}\" is not a \"${ConfigClass::class.jvmName}\"!"
+                        )
+                }
+        }
+
+    fun registerConfig(
+        id: String,
+        type: KClass<out WaltConfig>,
+    ) = registerConfig(ConfigData(id, type/*, false, multiple, onLoad*/))
```
mikeplotean, line 94: please remove commented code

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/featureflag/FeatureManager.kt
```diff
@@ -0,0 +1,155 @@
+package id.walt.commons.featureflag
+
+import id.walt.commons.config.ConfigManager
+import id.walt.commons.config.runconfig.RunConfiguration
+import io.klogging.noCoLogger
+
+object FeatureManager {
+    val baseFeatures = HashSet<String>()
```
mikeplotean, line 8: seems to not have any references, is it still required?

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/featureflag/FeatureManager.kt
mikeplotean, line 13: perhaps this also can be renamed to `logger`?

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/featureflag/FeatureManager.kt
```diff
@@ -0,0 +1,155 @@
...
+    fun registerCatalogs(catalogs: List<ServiceFeatureCatalog>) {
+        catalogs.forEach { catalog ->
+            registerBaseFeatures(catalog.baseFeatures)
+        }
+        catalogs.forEach { catalog ->
+            registerOptionalFeatures(catalog.optionalFeatures)
+        }
+    }
+
+    fun registerFeature(feature: AbstractFeature) {
```
mikeplotean, line 97: `registerFeature` can be private

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/featureflag/FeatureManager.kt
```diff
@@ -0,0 +1,155 @@
...
+        runIfEnabled(feature, this)
+    }
+
+    fun registerBaseFeatures(baseFeatures: List<BaseFeature>) {
+        baseFeatures.forEach {
+            log.debug { "Registering base feature \"${it.name}\"..." }
+            registerFeature(it)
+        }
+    }
+    fun registerOptionalFeatures(optionalFeatures: List<OptionalFeature>) {
```
mikeplotean, line 77: maybe also private?

mikeplotean (changes requested):

mikeplotean:
in its current state, wallet-api fails to start because registration-defaults.conf could not be loaded

this is also reproducible for the unrefactored version (issue being [reported](https://github.com/walt-id/waltid-identity/issues/475)). the [proposed fix](https://github.com/walt-id/waltid-identity/pull/476) also works for the refactored version.

mikeplotean:
also, just be aware, current db config is set up for postgres

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/webauthn/WebauthnService.kt
```diff
@@ -9,9 +9,9 @@ import com.webauthn4j.data.client.Origin
 import com.webauthn4j.data.client.challenge.Challenge
 import com.webauthn4j.data.client.challenge.DefaultChallenge
 import com.webauthn4j.server.ServerProperty
-import com.webauthn4j.validator.exception.ValidationException
-
+import com.webauthn4j.validator.exception.ValidationException*/
 
+/* Complete this implementation:
```
mikeplotean, line 14: maybe it is good to have an issue on github, so it doesn't get lost or forgotten?

waltid-services/waltid-wallet-api/src/main/resources/db/postgres/V1__create_tables_account_email_wallet_accountwallet_walletoperationhistories.sql
```diff
@@ -0,0 +1,227 @@
+-- ----------------------------------
+-- Emails table
```
mikeplotean, line 2: since there's no migration, maybe these migration files can also be removed?

waltid-services/waltid-wallet-api/k8s/deployment-prod.yaml
```diff
@@ -33,7 +33,6 @@ metadata:
   name: wallet-oidc-config
 data:
   oidc.conf: |
```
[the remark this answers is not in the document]
mikeplotean: there's a wallet-db-config mounted as `db.conf`
and a wallet-datasource-config mounted as `db.sqlite.conf`
I expect it to have the unified config

BREAKING_CHANGES.md
```diff
@@ -0,0 +1,30 @@
+# Breaking changes since 0.3.1
+
+## Configuration:
+
+#### Feature system
+- Introduced feature system, which is controlled by `_features.conf`.
+
+#### Database (waltid-wallet-api)
+- Unified `db.sqlite.conf`, `db.postgres.conf`, `db.mssql.conf`, `db.conf` (referencing either of the first 3) to just `db.conf`
```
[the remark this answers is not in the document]
mikeplotean: can update the .env hierarchy to select the jdbc-url and driver-class based on the engine value

waltid-services/waltid-wallet-api/k8s/deployment-portal.yaml
```diff
@@ -28,14 +28,6 @@ data:
 ---
```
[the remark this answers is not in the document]
mikeplotean: there's a wallet-db-config mounted as `db.conf`
and a wallet-datasource-config mounted as `db.sqlite.conf`
I expect it to have the unified config

mikeplotean (changes requested):

mikeplotean (approved):

mikeplotean (dismissed):

mikeplotean (dismissed):

mikeplotean (approved):

mikeplotean (approved):

PR #461, opened 2024-06-03
Docker compose tests

ahmedsmiai:


.github/workflows/docker-compose-test.yml
```diff
@@ -0,0 +1,48 @@
...
+          KUBE_CONFIG: ${{ secrets.HETZNER_KUBE_CONFIG }}
+        with:
+          args: wait --for=condition=ready pod -l app=podman-priv --timeout=120s
+
+      - name: Run Tests
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.HETZNER_KUBE_CONFIG }}
+        with:
+          args: exec -it $(kubectl get pods -l app=podman-priv -o jsonpath="{.items[0].metadata.name}") -- bash -c "cd waltid-identity/docker-compose && pip3 install requests && python3 test-script.py"
```
mikeplotean, line 34: I think the test script has been renamed to `api-test.py`

.github/workflows/docker-compose-test.yml
```diff
@@ -0,0 +1,48 @@
...
+
+#      - name: Generate Test Reports
```
mikeplotean, line 36: maybe this commented code can be removed?

docker-compose/api-test.py
```diff
@@ -0,0 +1,79 @@
...
+        cls.wallet_url = "http://localhost:7001"
+        cls.headers = {"Content-Type": "application/json" }
+        cls.name = ''.join(random.choices(string.ascii_letters + string.digits, k=8))
+        cls.email = f"{cls.name}@example.com"
+        cls.password = "password"  # Replace with a proper password policy if needed
+        cls.account_type = "email"  # Adjust this based on your actual account types
+        cls.did_method = "jwk"  # Add your own value
+        cls.key_algorithm = "Ed25519"  # Add your own value
+        cls.key_origin = "jwk"  # Add your own value
+        cls.jwk_payload = "{ 'backend': 'jwk', 'keyType': 'Ed25519' }"  # Add your own value
```
mikeplotean, line 20: wouldn't it be better to use json.dumps?
e.g. 
```
json.dumps({"backend":"jwl","keyType":"Ed25519"})
```

because otherwise it throws:
```
Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 2: Encountered an unknown key ''backend'' at path: $
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: { 'backend': 'jwk', 'keyType': 'Ed25519' }
```

.github/workflows/docker-compose-test.yml
```diff
@@ -0,0 +1,48 @@
...
+          KUBE_CONFIG: ${{ secrets.HETZNER_KUBE_CONFIG }}
+        with:
+          args: apply -f docker-compose/podman.yaml
+
+      - name: Wait for pod to be ready
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.HETZNER_KUBE_CONFIG }}
+        with:
+          args: wait --for=condition=ready pod -l app=podman-priv --timeout=120s
```
mikeplotean, line 27: doesn't this wait only for the host container to get started (not the docker-compose stack)?

mikeplotean (changes requested):

docker-compose/api-test.py
```diff
@@ -0,0 +1,79 @@
...
+import json
+
+class WalletAPITest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.wallet_url = "http://localhost:7001"
+        cls.headers = {"Content-Type": "application/json" }
+        cls.name = ''.join(random.choices(string.ascii_letters + string.digits, k=8))
+        cls.email = f"{cls.name}@example.com"
```
mikeplotean, line 14: I'd rather randomize the email than the name

mikeplotean (changes requested):

docker-compose/api-test.py
ahmedsmiai: The email already uses the randomized name for prefix, therefore i guess the email is randomized aswell, or do you have another idea ?

mikeplotean: oh, my bad

mikeplotean (approved):

PR #479, opened 2024-06-10
Fix issuance flow: Docker compose hostname update

SuperBatata:


mikeplotean (approved):
but I'm afraid this will brake the compose for everyone again
and we'll have to go though the same "have you updated your hosts file"

SuperBatata:
@mikeplotean  I understand your concern . To mitigate this, we could consider merging these changes last minute and ensuring the updates are clearly highlighted in the release notes. This way, users will be informed and can make the necessary adjustments proactively.

PR #484, opened 2024-06-12
chore: Temporarily commented custom credential issuance button due to issuer-metadata

dinkar-jain:


mikeplotean (dismissed):
maybe adding the reason why the code is commented will prevent it from getting deleted with some future PR which will request to "remove the commented code"
this is not nice to have a comment explaining a commented code, but the other solutions I can see would be:
- remove it completely
- or add some logic to make it configurable

not a request, just smth. to mention

dinkar-jain:
> maybe adding the reason why the code is commented will prevent it from getting deleted with some future PR which will request to "remove the commented code" this is not nice to have a comment explaining a commented code, but the other solutions I can see would be:
> 
> * remove it completely
> * or add some logic to make it configurable
> 
> not a request, just smth. to mention

Added a comment for now, to prevent blindly removing it in the future

mikeplotean (approved):

PR #537, opened 2024-06-26
[PRO-784] Error importing elliptic curve public raw keys

mikeplotean:


mikeplotean:
fixes PRO-784

mikeplotean:
fixed by https://github.com/walt-id/waltid-identity/pull/546

PR #538, opened 2024-06-26
[PRO-785] Wrong credentialConfigurationId in issueBatch sample payload

mikeplotean:
fixes PRO-785


mikeplotean:
updated `/openid4vc/jwt/issueBatch` sample payload to openbadge + bankId credentials,
so that it's similar to the `/openid4vc/sdjwt/issueBatch` sample payload

Issue #593, opened 2024-07-09
Missing `Location` header field on resource creation with a POST


pohutukawa:
**Describe the bug**
It is common in RESTful CRUD services to return the resource URI of a newly created resource with a `POST` HTTP operation via the `Location` header field. Many operations (e.g. creation of a DID) return information on the new resource name in the response body as a string (even though the request has specified an accepted type `application/json`). However, when importing a JWK key, only an empty response body is returned, and the resource or `keyId` of the new key is completely unknown (unless a `kid` element was part of the JWK imported already).

It is customary to use the HTTP protocol on RESTful CRUD operations by specifying the `Location` header field with the full URI to the created resource. That also eliminates the need of a HTTP body in the response. (If I'm picky, we'd expect a `201 Create` response here instead of a `200 OK`, but that's tolerable if the information is available).

**To Reproduce**
Steps to reproduce the behaviour:
1. Import a new key via a JWK with a `POST /wallet-api/wallet/{wallet}/keys/import`
2. Receive a `200 OK` response
3. Inspect all the header fields and the body of the response, without any indication of the newly created key ID or resource URI given

**Expected behaviour**
Preferably the new resource URI in the `Location` header field of the response. Alternatively (or both) the resource URI or minimally the new key ID in the response body.

**Actual behaviour**
Unidentifiable key ID or new key's resource URI.

**Environment**
`curl` command line, scripted (e.g. Python) or via the Swagger UI.

cpatsonakis:
Hi!

Thank you for bringing this issue to our attention.

As this affects multiple CRUD-based endpoints of the Wallet API, further investigation is required to
organize, probably multiple, PRs towards a resolution.

We will update this issue accordingly following our internal discussion.

Thanks again!

SuperBatata:
should be fixed with this pr : #597

Issue #592, opened 2024-07-09
a few 500 Internal Server Errors (SQLITE_CONSTRAINT_PRIMARYKEY)


pohutukawa:
**Describe the bug**
Exploring the wallet-api, I have stumbled upon a few unhandled exceptions leading to 500 Internal Server Errors. In this case they were all tied to primary key constraint errors in the DB, so I would guess that they're likely applicable to other persistence back-ends as well.

Specifically they have occurred in the following cases:

* Create an account if the account ID (email) already exists. (`{"exception":true,"status":"Internal Server Error","code":"500","message":"Could not register user: Account already exists!"}`)
* Create a DID (Web) if the DID already exists. (`{"exception": true, "status": "Internal Server Error", "code": "500", "message": "org.sqlite.SQLiteException: [SQLITE_CONSTRAINT_PRIMARYKEY] A PRIMARY KEY constraint failed (UNIQUE constraint failed: wallet_dids.wallet, wallet_dids.did)"}`)
* Import a key (JWK) if the key ID (auto-generated) already exists. (`{"exception":true,"status":"Internal Server Error","code":"500","message":"org.sqlite.SQLiteException: [SQLITE_CONSTRAINT_PRIMARYKEY] A PRIMARY KEY constraint failed (UNIQUE constraint failed: wallet_keys.wallet, wallet_keys.kid)"}`)

**To Reproduce**
Steps to reproduce the behaviour:
1. Import a key in JWK format: `POST /wallet-api/wallet/{wallet}/keys/import`
2. Import that same key in JWK format again.
3. Observe the API returning a `500` error due to an unhandled exception.

**Expected behaviour**
A `409 Conflict` or `400 Bad Request` error response after successfully handling the exception.

**Actual behaviour**
An unhandled exception/not gracefully handled exception leading to a `500 Internal Server Error` response without significant meaningful context assisting the user to resolve the situation.
eenshots to help explain your problem.

**Environment**
**Desktop (please complete the following information):**
 - Browser: `curl` command line or other scripted (e.g. via Python `requests`)
**Additional context**

PR #597, opened 2024-07-09
Fix issues in  wallet api

SuperBatata:
## Description

Fixes #[issue number]

## Type of Change

## Checklist

### DIFF
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
@@ -454,6 +454,10 @@ class SSIKit2WalletService(
 
         val key = keyResult.getOrThrow()
         val keyId = key.getKeyId()
+        if (KeysService.get(walletId, keyId) !== null) {
+            throw IllegalArgumentException("Key with ID $keyId already exists in the database")
+        }
+
         eventUseCase.log(
             action = EventType.Key.Import,
             originator = "wallet",
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/AuthController.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/AuthController.kt
@@ -295,6 +295,7 @@ fun Application.auth() {
                     response {
                         HttpStatusCode.Created to { description = "Registration succeeded " }
                         HttpStatusCode.BadRequest to { description = "Registration failed" }
+                        HttpStatusCode.Conflict to { description = "Account already exists!" }
                     }
                 }) {
                 val req = loginRequestJson.decodeFromString<AccountRequest>(call.receive())
@@ -303,7 +304,13 @@ fun Application.auth() {
                         call.response.status(HttpStatusCode.Created)
                         call.respond("Registration succeeded ")
                     }
-                    .onFailure { call.respond(HttpStatusCode.BadRequest, it.localizedMessage) }
+                    .onFailure {
+                        if (it.message?.contains("Account already exists!") == true) {
+                            call.respond(HttpStatusCode.Conflict, it.localizedMessage)
+                        } else {
+                            call.respond(HttpStatusCode.BadRequest, it.localizedMessage)
+                        }
+                    }
             }
 
             authenticate("auth-session", "auth-bearer", "auth-bearer-alternative") {
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/account/AccountsService.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/account/AccountsService.kt
@@ -24,47 +24,95 @@ object AccountsService {
 
     private val defaultGenerationConfig by lazy { ConfigManager.getConfig<RegistrationDefaultsConfig>() }
 
-    suspend fun register(tenant: String = "", request: AccountRequest): Result<RegistrationResult> = when (request) {
-        is EmailAccountRequest -> EmailAccountStrategy.register(tenant, request)
-        is AddressAccountRequest -> Web3WalletAccountStrategy.register(tenant, request)
-        is OidcAccountRequest -> OidcAccountStrategy.register(tenant, request)
-        is KeycloakAccountRequest -> KeycloakAccountStrategy.register(tenant, request)
-        is OidcUniqueSubjectRequest -> OidcUniqueSubjectStrategy.register(tenant, request)
+//    suspend fun register(tenant: String = "", request: AccountRequest): Result<RegistrationResult> = when (request) {
+//        is EmailAccountRequest -> EmailAccountStrategy.register(tenant, request)
+//        is AddressAccountRequest -> Web3WalletAccountStrategy.register(tenant, request)
+//        is OidcAccountRequest -> OidcAccountStrategy.register(tenant, request)
+//        is KeycloakAccountRequest -> KeycloakAccountStrategy.register(tenant, request)
+//        is OidcUniqueSubjectRequest -> OidcUniqueSubjectStrategy.register(tenant, request)
+//
+//    }.onSuccess { registrationResult ->
+//        val registeredUserId = registrationResult.id
+//
+//        val createdInitialWalletId = transaction {
+//            WalletServiceManager.createWallet(tenant, registeredUserId)
+//        }.also { walletId ->
+//            //TODO: inject
+//            WalletServiceManager.issuerUseCase.add(IssuerDataTransferObject.default(walletId))
+//        }
+//
+//        val walletService = WalletServiceManager.getWalletService(tenant, registeredUserId, createdInitialWalletId)
+//        WalletServiceManager.eventUseCase.log(
+//            action = EventType.Account.Create,
+//            originator = "wallet",
+//            tenant = tenant,
+//            accountId = registeredUserId,
+//            walletId = createdInitialWalletId,
+//            data = AccountEventData(accountId = request.name)
+//        )
+//
+//        // Add default data:
+//
+//        val createdKey = walletService.generateKey(defaultGenerationConfig.defaultKeyConfig)
+//
+//        val createdDid = walletService.createDid(
+//            method = defaultGenerationConfig.didMethod,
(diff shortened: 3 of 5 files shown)


mikeplotean (dismissed):

alegomes:
I was testing the #592. 

First run is ok.

image

The second run doesn't return the SQLITE_CONSTRAINT_PRIMARYKEY anymore. Great. 

image

However, it still returns 500 and, according to the [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-500-internal-server-error), this return code is for "unexpected conditions" which I'm sure it's the case.

> The 500 (Internal Server Error) status code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request.

It's totally predictable that the user, intentionally or not, try to create duplicated DIDs.

As such, I think that, maybe, and error [409](https://www.rfc-editor.org/rfc/rfc9110.html#name-409-conflict) or even a generic 400 might be more suitable for this situation.

Let's make the web more semantic :-)

alegomes:
And I want to take advantage of ticket #593 to register another possible improvement that we can make in the future in our APIs: HATEOAS

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/AuthController.kt
```diff
@@ -295,15 +295,30 @@ fun Application.auth() {
...
+                }
+                val accountRequest = loginRequestJson.decodeFromString<AccountRequest>(call.receive())
+                AccountsService.register("", accountRequest)
                     .onSuccess {
                         call.response.status(HttpStatusCode.Created)
                         call.respond("Registration succeeded ")
                     }
-                    .onFailure { call.respond(HttpStatusCode.BadRequest, it.localizedMessage) }
+                    .onFailure {
+                        if (it.message?.contains("Account already exists!") == true) {
```
mikeplotean, line 316: comparing against the error message to compute the error status doesn't feel right
maybe we can have a dedicated type for application logic errors
e.g. have an entity representing an `Error`:
```kotlin
abstract class Error(
    val code: String,
    override val message: String,
) : Throwable()
```
and then define for each module (Account, Did, Key, etc.) such errors, e.g.:
```kotlin
object Errors {
    object General {
        class Conflict(type: KClass<*>, data: String) = Error(
            code = "conflict.already.exists",
            message = "${type.getFullName()} for $data already exists."
        )
    }
}
```
then throw an instance of this error (e.g. `throw Errors.General.Conflict(Account::class, request.email)`)

mikeplotean: computing the status could be done potentially using ktor StatusPages:
```kotlin
install(StatusPages) {
    exception<Throwable> { call, cause ->
        if(cause is Errors.General.Conflict) {
            call.respondText(text = "409: $cause" , status = HttpStatusCode.Conflict)
        } else {
            call.respondText(text = "500: $cause" , status = HttpStatusCode.InternalServerError)
        }
    }
}
```

mikeplotean: or extending the Error class with a method that inheritors would have to implement
e.g.
```kotlin
abstract class Error(
    val code: String,
    override val message: String,
) : Throwable() {
	abstract fun toHttpStatus(): HttpStatusCode
}

object Errors {
    object General {
        class Conflict(type: KClass<*>, data: String) = Error(
            code = "conflict.already.exists",
            message = "${type.getFullName()} for $data already exists."
        ){
        	override fun toHttpStatus(): HttpStatusCode = HttpStatusCode.Conflict
        }
    }
}
```

mikeplotean (approved):

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/AuthController.kt
SuperBatata: will implement this feature in a separate branch

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/DidCreation.kt
```diff
@@ -49,9 +49,18 @@ object DidCreation {
...
-                DidWebMethodName, extractDidCreateParameters(DidWebMethodName, context.request.queryParameters)
-            ).let { context.respond(it) }
+            val parameters = extractDidCreateParameters(DidWebMethodName, context.request.queryParameters)
+            runCatching {
+                getWalletService().createDid(DidWebMethodName, parameters)
+            }.onFailure {
+                if (it.message?.contains("DID already exists") == true) {
+                    context.respond(HttpStatusCode.Conflict, it.message!!)
+                } else {
+                    context.respond(HttpStatusCode.InternalServerError, it.message ?: "Internal Server Error")
```
mikeplotean, line 59: maybe also display the error (not just internal server error)
if so, the whole failure block could be simplified, e.g.
```kotlin
val status = if (it.message.contains("DID already exists") == true) {
    HttpStatusCode.Conflict
} else {
    HttpStatusCode.InternalServerError
}
context.respond(status, it.message?:"some-fallback-message")
```

mikeplotean (dismissed):

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/DidCreation.kt
```diff
@@ -49,9 +50,26 @@ object DidCreation {
                 }
             }
         }) {
-            getWalletService().createDid(
-                DidWebMethodName, extractDidCreateParameters(DidWebMethodName, context.request.queryParameters)
-            ).let { context.respond(it) }
+            val parameters = extractDidCreateParameters(DidWebMethodName, context.request.queryParameters)
+            runCatching {
```
mikeplotean, line 54: I think this is not required anymore (catching the exception in controller), since there is a type - ConflictException - for it in service-commons WebExceptions that is thrown by the did-service, so status-pages should be able to display it

mikeplotean (approved):

mikeplotean (approved):

mikeplotean (dismissed):

alegomes (changes requested):
For the sake of consistency, I think the status error 409 should be returned when a  Key is duplicated. Then, it would be the same as for duplicated accounts and DIDs.

image

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -124,11 +127,11 @@ class SSIKit2WalletService(
     }
 
     override suspend fun restoreCredential(id: String): WalletCredential =
-        credentialService.restore(walletId, id) ?: error("Credential not found: $id")
+        credentialService.restore(walletId, id) ?: throw NotFoundException("Credential not found: $id")
 
     override suspend fun getCredential(credentialId: String): WalletCredential =
         credentialService.get(walletId, credentialId)
-            ?: throw IllegalArgumentException("WalletCredential not found for credentialId: $credentialId")
+            ?: throw NotFoundException("WalletCredential not found for credentialId: $credentialId")
```
alegomes, line 134: This error message exposes a class and a parameter name that doesn't make sense to the final user. Why not maintain the consistency with other messages just with "Credential not found: $id"?

alegomes: However, I would still change that last message to "Credential with ID $id not found" because, just by looking at the final error message "Credential not found: 98723645453738", it's not clear what this number is about.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -436,34 +439,52 @@ class SSIKit2WalletService(
...
                 "pem" -> JWKKey.importPEM(jwkOrPem)
                 "jwk" -> JWKKey.importJWK(jwkOrPem)
-                else -> throw IllegalArgumentException("Unknown key type: $type")
+                else -> throw UnsupportedMediaTypeException("Unknown key type: $keyType")
+            }.getOrThrow()
+
+            val keyId = key.getKeyId()
+
+            if (KeysService.exists(walletId, keyId)) {
+                throw ConflictException("Key with ID $keyId already exists in the database")
```
alegomes, line 463: That's a very clear message. 👏

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -343,7 +346,7 @@ class SSIKit2WalletService(
 
     override suspend fun loadDid(did: String): JsonObject =
         DidsService.get(walletId, did)?.let { Json.parseToJsonElement(it.document).jsonObject }
-            ?: throw IllegalArgumentException("Did not found: $did for account: $walletId")
+            ?: throw NotFoundException("Did not found: $did for account: $walletId")
```
alegomes, line 349: image

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/DidController.kt
```diff
@@ -106,9 +106,10 @@ fun Application.dids() = walletRoute {
                 }
             }
             response {
-                HttpStatusCode.OK to {
-                    description = "DID created"
-                }
+                HttpStatusCode.OK to { description = "DID created" }
+                HttpStatusCode.Conflict to { description = "DID already exists" }
+                HttpStatusCode.BadRequest to { description = "DID could not be created" }
```
alegomes, line 111: Can we provide the reason?

alegomes (approved):
Just some minor suggestions to enhance Dev Experience.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/DidController.kt
SuperBatata: What do you mean by reason ?

PR #605, opened 2024-07-11
chore: Updated k8s config for wallet to auto-deploy new wallet

dinkar-jain:


dinkar-jain:
The contents of this PR is incorporated properly into the main branch with PR [698](https://github.com/walt-id/waltid-identity/pull/698)

PR #630, opened 2024-07-17
Fix raw import of secp256k1 and secp256r1 keys

### LINKED ISSUE #629: Raw import of secp256r1 and secp256k1 keys does not work
**Describe the bug**
Attempting to import raw encoded secp256r1 and secp256k1 keys throws an invalid point encoding error.

**To Reproduce**
Invoke the `importRawPublicKey` function of the `Key` type with the raw encoded bytes of a secp256r1/secp256k1 key.

**Expected behavior**
Should correctly parse an instance of `Key` type.

**Actual behavior**
A clear and concise description of what actually happens.

**Screenshots**

**Environment**
**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

cpatsonakis:
## Description

Fixes #629.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have performed a self-review of my code
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/waltid-libraries/waltid-crypto/src/jvmMain/kotlin/id/walt/crypto/keys/JWKKeyGenerator.jvm.kt
+++ b/waltid-libraries/waltid-crypto/src/jvmMain/kotlin/id/walt/crypto/keys/JWKKeyGenerator.jvm.kt
@@ -13,9 +13,11 @@ import com.nimbusds.jose.jwk.gen.RSAKeyGenerator
 import com.nimbusds.jose.util.Base64URL
 import id.walt.crypto.keys.jwk.JWKKey
 import id.walt.crypto.keys.jwk.JWKKeyCreator
+import org.bouncycastle.asn1.ASN1BitString
 import org.bouncycastle.asn1.ASN1Integer
 import org.bouncycastle.asn1.ASN1Sequence
 import org.bouncycastle.jce.ECNamedCurveTable
+import org.bouncycastle.math.ec.ECPoint
 import java.security.KeyFactory
 import java.security.interfaces.RSAPublicKey
 import java.security.spec.RSAPublicKeySpec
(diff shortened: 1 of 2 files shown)


waltid-libraries/waltid-crypto/src/jvmMain/kotlin/id/walt/crypto/keys/JWKKeyGenerator.jvm.kt
```diff
@@ -39,26 +42,41 @@ object JvmJWKKeyCreator : JWKKeyCreator {
...
+        val bcEC = ECNamedCurveTable.getParameterSpec(curve.name).curve
+        val pubKeyEcPoint = decodeRawPubKeyEcPoint(bcEC, rawPublicKey)
+        return ECKey.Builder(
+            curve,
+            Base64URL.encode(pubKeyEcPoint.rawXCoord.encoded),
+            Base64URL.encode(pubKeyEcPoint.rawYCoord.encoded),
+        ).build()
+    }
+
+    private fun decodeRawPubKeyEcPoint(bcEC: ECCurve, rawEncodedPoint: ByteArray): ECPoint =
```
cpatsonakis, line 71: I want to add a note here just to make it super clear for review purposes. The reason why this try-catch block was added is so that Uniresolver test cases in the did library don't start randomly breaking. Note that this ASN1 based decoding process is the correct one, but for some reason that doesn't work for the uniresolver stuff. I don't know why this is the case, perhaps they made a mistake when encoding the did strings, but this is nevertheless a fact.

I didn't want to delete these test cases (although I could have, since did:key resolution does no longer depend on the uniresolver), so this point is up for discussion which can be addressed by a small update in the context of this PR.

PR #721, opened 2024-08-29
Sd jwt vc iss should support dids

severinstampler:
SD-JWT-VC issuance and presentation/verification should work with DID as issuer key reference. 
Currently x5c certificates have to be provided to derive the issuer key for verification.


chsavvaidis:
- [ ]  Make issuer did nullable in the issuance request when x.509 are used
- [ ]  CryptographicBindingSupportMethod in issuer metadata is always did. should we add x509?
- [ ]  Add examples in swagger
- [ ]  Rename, split, add required code to LspPotentialTest to make:  haip ietf sd-jwt  tests and oid4vc ietf sd-jwt  tests since
- [ ]  Why disable metadata conf in issuer k8s?

PR #754, opened 2024-09-17
implement X25519 as Key Type.

### LINKED ISSUE #580: Add support for X25519 key type
**Is your feature request related to a problem? Please describe.**
Support for X25519 keys would allow the `keyAgreement` property to be properly defined in the DID docs generated by the `waltid-did` library. A manifestation of this problem is described in #579.

**Describe the solution you'd like**
Add support for X25519 keys in the `waltid-crypto` library.

**Additional context**
Thankfully, both bouncy castle and nimbus support this key type.

mr-sebbo:
## Description

KeyType now allows X25519 to be stored.

[Fixes #580](https://github.com/walt-id/waltid-identity/issues/580)

## Type of Change

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have performed a self-review of my code
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


cpatsonakis:
Hello @mr-sebbo! First of all, thank you for the time and effort that you put to contribute to our stack.
We find it very cool and greatly appreciate that open source users have ideas and valuable feedback that
can enhance the set of features supported by our components.

We have reviewed your PR, discussed it internally and have come to the conclusion that merging it will
lead to some issues and incompatibilities that we would like to inform you about. The first issue is that
there are failing test cases related to the multiplatform (MP) nature of our core libraries, which is an important
aspect of our stack that we want to improve over time. The second (interrelated) issue is that this PR
*breaks the contract* of the `Key` interface, i.e., the functionality that it promises to deliver, which essentially
relates to signing and verification keys. Adding on top the issue of MP, we need to be conscious of the fact
that there are provisions in place to account for, e.g., JS builds. Furthermore, there exists also the domain of
KMS integrations that we support, which complicates matters even further.

That being said, we are evaluating our options on how we can add support for this particular key type. In
the context of these efforts, it would be very useful for us and we would greatly appreciate if you could
provide additional information/context on the use case(s), or requirements, that you have in mind in a
separate issue thead. Example questions that come to mind:

- Is there a particular reason that you need `X25519` keys? In the context of the JOSE family of standards and
in conjunction to the DID Core specification, one can use RSA, secp256r1 and secp256k1 JWKs to, for instance
["... to wrap a decryption key for the recipient."](https://www.w3.org/TR/did-core/#key-agreement:~:text=to%20wrap%20a%20decryption%20key%20for%20the%20recipient.)
- We would be very much interested to know in which DID method(s) your interests lie, in order to take this
into consideration for future updates on our DID Lib. 

Thank you once again for your efforts and the valuable feedback that you have provided!

pohutukawa:
Hi @cpatsonakis,

thanks for that thorough reply. As this is an open source project under your
(Walt.id's) domain/ownership, I can fully understand if this pull request as it
currently stands may not fit well with your requirements. Though, there are some
aspects raised that I would like to put down. Excuse me, If I may put my
'university lecturer' hat on to explain some things. If there are other concerns
besides the 'cryptography' one, please, let us know. We may put in some
additional work to make it a better fit, possibly with some input from Walt.id
staff as we're not very familiar with the software architecture and design
principles put forward in this work.

Also, before we get to the cryptography bit, the note that DID Docments are
**not** exclusively used for publication/management of signature and
verification keys **only.** Particularly the `keyAgreement` element speaks
firmly towards encryption and confidentiality, which would then either need to
be left out or satisfied with a suitable public key (which is not a pure signing
key, such as EdDSA). Also, accepted good practice strongly encourages to
separate the concerns of signing and encryption into different key pairs.

But not the cryptography bit.

Public key encryption can be achieved in two different ways, but in the end it
comes down to provide a secure mechanism to obtain a symmetric key for
encryption of the bulk of the content (as public key encryption is about 4-5
orders of magnitude computationally more expensive than symmetric cryptography
of the same strength). This can be achieved using two different ways: key
wrapping/encapsulation (e.g. via RSA or other KEM algorithms) or key
agreement/derivation (e.g. via a Diffie-Hellman flavour, DH or ECDH). In the
former case only a single key pair is involved and the symmetric key is 'chosen'
by the encrypting party. In the latter case, a key is derived through the
involvement of two key pairs, thus giving the controllers of both key pairs
influence on the key strength. At present, the second method (key
agreement/derivation) is commonly more favoured.

Most commonly used public key cryptography (including signing) at present is
based on the discrete logarithm problem (DLP) in its traditional form (using
exponentiation of very large integers) or in its elliptic curve form (using
scalar multiplication of points on a curve). The former category is covered by
algorithms like RSA, DSA and DH. These require significant sizes in keys (a few
1000 bits large) and are generating proportionally large 'payloads' (ciphers,
signatures) and require more processing demands (memory, CPU cycles, entropy for
key generation). For that reason most modern systems have pivoted or built on
top of elliptic curve cryptography (based on the elliptic curve DLP with scalar
multiplication) for its more favourable properties in compact size and compute
requirements.

Ciphers are originating from different 'origins'. Some are defined/promoted
through standards organisations (e.g. NIST in the USA), whereas others have
gained de facto adoption through thorough research, publications and public
trust. The former category include RSA, curves like NIST P-256 (secp256r1), SHA,
AES, etc. This is great for cases where compliance with (government) standards
is required. Unfortunately there have been many reported or suspected cases
where trust in these official bodies has been broken (e.g. double EC random
number generators, the parameters of NIST-P curves, the recent kerfuffle around
the post-quantum standard competition Kyber algorithm with faulty snake oil
calculations on the claimed security of parameter sets). For these types of
reasons many have opted to put more trust into alternative ciphers/curves, such
as X25519/Ed25519, ChaCha20/Salsa20 with Poly1305, secp256k1 ('Bitcoin'/Koblitz
curve), etc. These have all received their 'street creds' through thorough
research and documented reasoning. And particularly the combo X25519/Ed25519 +
ChaCha20/Salsa20 + Poly1305 has gained vast adoption across browsers, web
servers, OpenSSH, OpenSSL and many other implementations, and are nowadays one
of the most popular underpinning of public networks and security.

Lastly, in cryptography one must always keep a perspective of at least 10 years
forward into the future. That time frame likely covers a point in time where
quantum computers will gain the ability to feasibly break public key ciphers
based on the DLP problem [0] (traditional and EC) through Shor's algorithm, and
weaken symmetric ciphers by their square root in size (halving the bit strength)
through Grover's algorithm. Mitigation for the latter is easy (doubling
symmetric key strength requirements to 256 bits), but the former requires
completely new algorithms. The risk position for today is the 'Harvest now,
decrypt later' paradigm that will pose problems. Unfortunately, even though some
new ciphers have been standardised recently already, they have not yet undergone
the same cryptanalytical scrutiny as 'traditional' ciphers, yet. So they are
bearing the risk of being subject to yet unknown weaknesses discovered. Thus,
the current expectation is to meet the problem using hybrid approaches, fusing
traditional **and** post-quantum ciphers together into protocols [1]. A lot of
development in this direction is already underway with various specs being
drafted and refined (e.g. [2] for TLS). Many of these are building particularly
on X25519 as a core ECDH curve.

Anyway, to make this not too much longer: The above are our motivators to use
X25519 for public key encryption purposes, just like you (Walt.id) have made
Ed25519 their default cipher for signing. Additionally we are starting to look
for where we can align ourselves as best as we can on new things with
requirements for the post-quantum era, where these are starting to play into our
preferences for decision making as well.

For us the DID Document serves as a means to make public key sets for the
different purposes publicly known and discoverable. If we would have to pivot
after a time (e.g. if X25519 is not yet supported, but will become in the
future), we also will then need the ability to publicise multiple public keys
for the different purposes, either to allow for a later migration to alternative
keys (e.g. secp256r1 -> X25519) or for the case where keys need to be rotated,
but older content must not be rendered unverifiable.

As for the DID methods, currently we're aiming (for simplicity and general
acceptance) at `did:web`, but that may be superseded in the future by more
capable ones. Or (ideally) our customers would 'bring' their own from other
ecosystems (e.g. from the EU, from governments, etc.).

Sorry for the lengthy post ...


[0] https://www.linkedin.com/posts/jaime-gomez-garcia_pqc-rsa-cryptography-activity-7233821796649582592-nLpS/
[1] https://www.latacora.com/blog/2024/07/29/crypto-right-answers-pq/
[2] https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design-10

PR #761, opened 2024-09-20
WAL-242 refactor and add sdjwt vc support in external signature handlers

cpatsonakis:
## Description

Fixes #[issue number]

## Type of Change

## Checklist


waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeUtils.kt
```diff
@@ -1,16 +1,115 @@
...
+        keyId: String,
+        audience: String,
+        nonce: String?,
+        credentials: List<WalletCredential>,
+        disclosures: Map<String, List<String>>?,
+    ) = credentials.filter { it.format == CredentialFormat.sd_jwt_vc }.map { credential ->
+        val serializedVcWithDisclosures = listOf(
+            credential.document
+        ).plus(
+            if (disclosures?.containsKey(credential.id) == true) {
```
mikeplotean, line 96: maybe this could be a separate method, e.g. getDisclosures:
```kotlin
private fun getDisclosures(
    disclosures: Map<String, List<String>>?,
    credentialId: String
) = if (disclosures?.containsKey(credentialId) == true) {
    "~${disclosures[credentialId]!!.joinToString("~")}~"
} else "~"
```

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/openapi/examples/ExchangeExternalSignaturesExamples.kt
```diff
@@ -0,0 +1,480 @@
...
+
+    fun submitOid4vpRequestIETFSDJWTVCExample(): ValueExampleDescriptorDsl.() -> Unit = {
+        value = submitOid4vpRequestIETFSDJWTVCExample
+    }
+
+    //OID4VCI Examples
+
+    private val prepareOid4vciRequestDefaultExample = PrepareOID4VCIRequest(
+        did = "did:jwk:eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2Iiwia2lkIjoiQ3ZlUzFZdWdUNWV2VnoxOGswWHBjVGQtUVVsS1AxRzdCdTJJMkpoSjF2YyIsIngiOiJCeGN2d0lBVDhudzRtdjVtUDhSTzlrYmc3N25sbVZoOG5RcWZVUEtuRDVBIiwieSI6IkthU01seTBRQUhSQzZKMXFDRDloRkZnV0dXMXp1cExwU3ZtVGtHMDJmNHcifQ",
+        offerURL = "openid-credential-offer://localhost:22222/?credential_offer_uri=http%3A%2F%2Flocalhost%3A22222%2Fopenid4vc%2FcredentialOffer%3Fid%3D8030eb87-fa89-4820-ad36-8e89cc9ccdfe",
```
mikeplotean, line 262: I think localhost:22222 won't work out-of-the-box when running docker or from ide. It probably needs to be set up to `baseUrl` from _issuer-service.conf_ (or verifier, accordingly).
Also, there are several instances of them, maybe worth extracting them into a constant.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeExternalSignaturesController.kt
```diff
@@ -79,111 +84,148 @@ fun Application.exchangeExternalSignatures() = walletRoute {
...
+                        example(
+                            "IETF SD-JWT Verifiable Credential",
+                            ExchangeExternalSignaturesExamples.prepareOid4vpResponseIETFSDJWTVCExample(),
+                        )
                     }
                 }
             }
         }) {
             val walletService = getWalletService()
+            runCatching {
```
mikeplotean, line 104: I don't think `runCatching` is required here, since any exception would be caught by [StatusPages](https://github.com/walt-id/waltid-identity/blob/de6ebe904f0c6516b35fdc2988e1f7d33f3e90f7/waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/web/plugins/StatusPages.kt#L26) from service-commons and handled accordingly

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeExternalSignaturesController.kt
```diff
@@ -192,163 +234,182 @@ fun Application.exchangeExternalSignatures() = walletRoute {
...
-                        alg = authReq.clientMetadata!!.authorizationEncryptedResponseAlg!!,
-                        enc = authReq.clientMetadata!!.authorizationEncryptedResponseEnc!!,
-                        mapOf(
-                            "epk" to runBlocking { ephemeralWalletKey.getPublicKey().exportJWKObject() },
-                            "apu" to JsonPrimitive(Base64URL.encode(authReq.nonce).toString()),
-                            "apv" to JsonPrimitive(
-                                Base64URL.encode(authReq.nonce!!).toString()
-                            )
-                        )
+            runCatching {
```
mikeplotean, line 257: same here

however, this being the final step of the presentation, there probably has to be a distinction in the cause of the exception
and since there is an instance of incorrect logic mentioned, maybe have a separate ticket for solving it

mikeplotean (approved):

waltid-services/waltid-e2e-tests/src/test/kotlin/ExchangeExternalSignaturesApi.kt
```diff
@@ -378,40 +517,266 @@ class ExchangeExternalSignatures {
...
+            contentType(ContentType.Application.Json)
+            setBody(
+                buildJsonObject {
+                    put(
+                        "request_credentials",
+                        JsonArray(
+                            listOf(
+                                RequestedCredential(
+                                    format = VCFormat.sd_jwt_vc,
+                                    vct = "http://localhost:22222/identity_credential",
```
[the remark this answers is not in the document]
cpatsonakis: did something better I believe, fixed in c391cdab.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/exchange/IssuanceServiceBase.kt
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    @OptIn(ExperimentalSerializationApi::class)
+    protected suspend fun getCredentialData(
+        processedOffer: ProcessedCredentialOffer, manifest: JsonObject?,
+    ) = let {
+        val credential = processedOffer.credentialResponse.credential!!.jsonPrimitive.content
+
+        when (val credentialFormat = processedOffer.credentialResponse.format) {
```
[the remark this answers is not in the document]
cpatsonakis: in addition to the above, I also improved error handling a little bit and logging in the context of 68c4db19.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/openapi/examples/ExchangeExternalSignaturesExamples.kt
cpatsonakis: I don't understand, even if we do change the base URL in the example, it won't work because it requires interaction with the issuer to generate the offer in the first place who, in turn, generates the session identifier that is encoded in the offer's uri.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeUtils.kt
cpatsonakis: fixed in f489dda6.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeExternalSignaturesController.kt
cpatsonakis: I think it's much cleaner in the context of the handler to have it like this, as the response can be computed according to what is thrown. Also, in general, I am against introducing assumptions about what other components do, or handle, when not necessary. StatusPages are a relatively "new thing" and we'll see how they will turn out...

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeExternalSignaturesController.kt
cpatsonakis: Refer to my above comment.

On the incorrect logic part, this does not only affect this handler, but also the regular `usePresentationRequest` handler. A separate work item has been already created to address the issue that responses obtained following the submission of Authorization Responses need to be reworked.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/openapi/examples/ExchangeExternalSignaturesExamples.kt
mikeplotean: ok, this is just an example as documentation (unlike the ones in issuer- or verifier-api which could be used as actual request payload)

although it doesn't matter, I still think it would be nice to have it display the ones that we configured by default, http://localhost:7002 - for issuer and http://localhost:7003 - for verifier.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeExternalSignaturesController.kt
mikeplotean: agree with the assumptions aspect
but here, the actual error is hidden behind the "Unknown" - bad-request. The recipient won't receive AuthorizationError, "ephemeral key", etc. But maybe this is also smth. that could be handled within the presentation error logic ticket.

mikeplotean (approved):

mikeplotean (approved):

PR #762, opened 2024-09-20
WAL-199 - Enhance error handling in waltid library

SuperBatata:
## Description

Fixes #[issue number]

## Type of Change

## Checklist


mikeplotean (dismissed):

waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/keys/oci/OCIKeyRestApi.kt
cpatsonakis, line 1: Perhaps for cases that a cryptographic method is not supported, it would be useful to add a new exception type for not supported, or not implemented methods, e.g., something like CryptographicMethodNotSupported, or we could even make it more specific types depending on the context, e.g., KeyExportNotSupported, or even something like KeyOperationNotSupported to make it even broader. See below for instances that this can be applied.

cpatsonakis: Since I can't annotate code that has not been modified in this PR, some pointers are as follows:
`getThumbprint()`
`exportJWK()`: Btw, we implement `exportJWKObject()` but not this one???
`exportPEM()`
`getPublicKeyRepresentation()`

cpatsonakis (dismissed):

waltid-libraries/waltid-library-commons/build.gradle.kts
```diff
@@ -0,0 +1,207 @@
...
+
+plugins {
+    kotlin("multiplatform")
+    kotlin("plugin.serialization")
+    id("maven-publish")
+    id("com.github.ben-manes.versions")
+    id("love.forte.plugin.suspend-transform") version "2.0.20-0.9.2"
+}
+
+group = "id.walt"
```
mikeplotean, line 19: this being multiplatform now, I think it should be `id.walt.library-commons` or similar, so all build artifacts will go under the same root group, regardless of platform

for example for the crypto lib, it's `id.walt.crypto`

waltid-libraries/waltid-library-commons/src/commonMain/kotlin/id/walt/commons/exceptions/CryptoExceptions.kt
```diff
@@ -0,0 +1,56 @@
...
+object TSEError {
+    class LoginException(errors: List<String>) : CryptoException(
+        HttpStatusCode.InternalServerError,
+        "Errors occurred at TSE login: ${errors.joinToString { it }}"
+    )
+
+    class MissingAuthTokenException :
+        CryptoException(HttpStatusCode.InternalServerError, "Did not receive token after login!")
+
+    class MissingKeyNameException : CryptoException(HttpStatusCode.BadRequest, "Key name is missing in the request!")
```
mikeplotean, line 35: nitpicking alert!
I think the correct form in this case would be "missing from"

mikeplotean (dismissed):

mikeplotean (approved):

PR #777, opened 2024-09-27
WAL 202 did exceptions

SuperBatata:
## Description

Enhance did exceptions with custom exceptions with meaningful error messages

## Type of Change

## Checklist

## Breaking

- \-


waltid-libraries/waltid-did/src/commonMain/kotlin/id/walt/did/dids/document/models/service/Service.kt
```diff
@@ -65,17 +69,18 @@ data class ServiceMap(
 ) {
 
     init {
-        require(id.isNotBlank()) { "Service property id cannot be blank" }
+        require(id.isNotBlank()) { throw InvalidServiceIdException("Service property id cannot be blank") }
```
mikeplotean, line 72: `require` already throws an `IllegalArgumentException` and within that lambda function you provide the message which the exception should be constructed with

image

so, what happens when you throw within the lambda `lazyMessage` function itself is:
1. `require` calls `lazyMessage()` to retrieve the message / object to get passed along with the exception
2. during that call it fails
3. and throws the exception that was constructed within the `lazyMessage` function (hence never finishing the assignment to `message` and not proceeding with the rest of the flow)

an option would be to create a custom require method, e.g.:
```kotlin
private fun require(condition: Boolean, type: KType, callback: () -> String) {
    if (!condition) {
        throw createException(type, callback())
    }
}

private fun createException(type: KType, message: String) = when (type) {
    typeOf<InvalidServiceIdException>() -> InvalidServiceIdException(message)
    typeOf<InvalidServiceTypeException>() -> InvalidServiceTypeException(message)
    typeOf<EmptyServiceEndpointException>() -> EmptyServiceEndpointException(message)
    else -> IllegalArgumentException(message)
}
```
then call it, e.g.:
```kotlin
init {
    require(id.isNotBlank(), typeOf<InvalidServiceIdException>()) { "Service property id cannot be blank" }
    type.forEach {
        require(it.isNotBlank(), typeOf<InvalidServiceTypeException>()) { "Service type strings cannot be blank" }
    }
    require(
        serviceEndpoint.isNotEmpty(), typeOf<EmptyServiceEndpointException>()
    ) { "Service endpoint set cannot be empty" }
    customProperties?.forEach {
        require(!reservedKeys.contains(it.key), typeOf<ReservedKeyOverrideException>()) {
            "Invalid attempt to override reserved Service property with key ${it.key} via customProperties map"
        }
    }
}
```
in pure jvm, `require` could use more reflection, e.g.:
```kotlin
private fun require(condition: Boolean, clazz: KClass<Throwable>, callback: () -> String) {
    if (!condition) {
        val constructor = clazz::class.primaryConstructor
        throw constructor!!.call(callback()).objectInstance!!
    }
}
```
but I'm not sure smth. similar is available in multiplatform
however, neither one is an elegant solution

mikeplotean (changes requested):

PR #920, opened 2025-02-04
ci/cd : added opa server to docker-compose.yaml

SuperBatata:
## Description

completes WAL-931 : added opa server to docker compose

## Type of Change

## Checklist

## Breaking

- \-


docker-compose/docker-compose.yaml
```diff
@@ -180,6 +183,18 @@ services:
     volumes:
       - ./vault/init.sh:/vault/scripts/init.sh
 
+  opa-server:
+    image: openpolicyagent/opa:latest
+    container_name: opa-server
+    profiles:
+      - identity
```
mikeplotean, line 190: I would create a separate profile name for it, e.g. opa and replace identity with it
so that, as a user, I don't put additional burden on my machine if I don't use dynamic policies

docker-compose/docker-compose.yaml
```diff
@@ -52,10 +52,13 @@ services:
       dockerfile: waltid-services/waltid-verifier-api/Dockerfile
     depends_on:
       - caddy
+      - opa-server
```
mikeplotean, line 55: I understand the dependency, but perhaps there's a way to make it conditionally, based on profile or anything else
because currently, as a user, I am forced to use opa server, although I might not need dynamic policies for my use case

docker-compose/docker-compose.yaml
```diff
@@ -180,6 +183,18 @@ services:
...
       - ./vault/init.sh:/vault/scripts/init.sh
 
+  opa-server:
+    image: openpolicyagent/opa:latest
+    container_name: opa-server
+    profiles:
+      - identity
+      - all
+    ports:
+      - "8181:8181"
```
mikeplotean, line 193: perhaps the exposed port could be a variable in the .env file

mikeplotean (changes requested):
please check the comments, maybe there are some ways available to handle

mikeplotean (approved):

Issue #1058, opened 2025-05-13
Externalize image repository and api endpoints


tdiesler:
The image repository and API endpoint URLs are currently hard coded in various config files.
These should be configurable through

1. .env
2. envars 

The later is important for k8s

PR #1059, opened 2025-05-13
[#1058] Externalize image repository and api endpoints

tdiesler:
See #1058


tdiesler:
This PR leaves the original

ISSUER_CALLBACK_URL=http://localhost:7100

in place. Is that correct or should it perhaps be the url of the demo/dev wallet respectively?

dinkar-jain (changes requested):

dinkar-jain (dismissed):

tdiesler:
Rebased

docker-compose/.env
```diff
@@ -1,3 +1,11 @@
+VERSION_TAG=0.14.0-dev
```
dinkar-jain: Let's use the stable version "0.14.0"

docker-compose/docker-compose.yaml
```diff
@@ -102,33 +102,33 @@ services:
...
+      NEXT_PUBLIC_ISSUER: "http://localhost:$ISSUER_API_PORT"
+      NEXT_PUBLIC_VERIFIER: "http://localhost:$VERIFIER_API_PORT"
+      NEXT_PUBLIC_WALLET: "http://localhost:$DEMO_WALLET_FRONTEND_PORT"
       PORT: $WEB_PORTAL_PORT
 
   vc-repo:
-    platform: linux/x86_64
-    image: docker.io/waltid/vc-repository:latest
+    platform: linux/amd64
+    image: waltid/vc-repository:latest
```
dinkar-jain: Let's add "${IMAGE_PREFIX}" and "${VERSION_TAG:-latest}" here too

docker-compose/docker-compose.yaml
```diff
@@ -174,7 +174,7 @@ services:
     restart: always
 
   vault-init:
-    image: hashicorp/vault:latest
+    image: hashicorp/vault:${VERSION_VAULT:-latest}
```
dinkar-jain, line 176: Let's specify "platform" here too...

docker-compose/docker-compose.yaml
```diff
@@ -152,7 +152,7 @@ services:
       - $POSTGRES_DB_PORT:5432 #TODO: tcp/udp proxy with caddy
 
   vault:
-    image: hashicorp/vault:latest
+    image: hashicorp/vault:${VERSION_VAULT:-latest}
```
dinkar-jain, line 154: Let's specify "platform" here too...

docker-compose/docker-compose.yaml
```diff
@@ -102,33 +102,33 @@ services:
...
       context: ../
       dockerfile: waltid-applications/waltid-web-portal/Dockerfile
     depends_on:
       - caddy
     environment:
-      NEXT_PUBLIC_VC_REPO: "http://host.docker.internal:$VC_REPO_PORT"
-      NEXT_PUBLIC_ISSUER: "http://host.docker.internal:$ISSUER_API_PORT"
-      NEXT_PUBLIC_VERIFIER: "http://host.docker.internal:$VERIFIER_API_PORT"
-      NEXT_PUBLIC_WALLET: "http://host.docker.internal:$DEMO_WALLET_FRONTEND_PORT"
+      NEXT_PUBLIC_VC_REPO: "http://localhost:$VC_REPO_PORT"
```
dinkar-jain: The docker containers ports are not binded to localhost, so this won't work.

docker-compose/docker-compose.yaml
tdiesler: hashicorp/vault is multiarch - imho, we should leave it to the engine to resolve the platform
https://hub.docker.com/r/hashicorp/vault/tags

docker-compose/docker-compose.yaml
tdiesler: I made this change because 'host.docker.internal' does not work on DockerDesktop MacOS, whereas 'localhost' does. AFAIK these services bind to 0.0.0.0 which would cover localhost.

Perhaps we need to talk about the scope of supported environments. If this is narrow (e.g. just linux + docker) it only needs to work there and I would have to adjust that compose definition locally.

We could also externalize these host definition to .env and document there what it should be for the set of supported environments (imho preferred).

docker-compose/.env
tdiesler: Are you sure? Using a stable version as an image tag suggests just that i.e. the image is stable and running the codebase of the corresponding git tag.

While we're working with snapshots, the image tag should somehow document that - similar to 'latest'.
Ultimately, your call of course - just say the word.

docker-compose/docker-compose.yaml
tdiesler: The vc-repo is not built with the VERSION_TAG from the env i.e. it is not part of `docker compose build`
Nevertheless, I externalized like this: VERSION_VC_REPO=latest

tdiesler:
@dinkar-jain is there anything more to do here?

docker-compose/.env
dinkar-jain: Thanks for the input—totally see where you're coming from with tagging the image to reflect its snapshot nature. That said, we'd prefer to go with the stable version tag in this case, since some folks are using Docker Compose mainly to run the application rather than build it.

tdiesler: ok, changed to 0.15.0 (i.e. the next stable release we are working on)

docker-compose/.env
```diff
@@ -1,3 +1,15 @@
+VERSION_TAG=0.15.0
```
dinkar-jain: Can you please update this to 0.14.0, since we have not released 0.15.0

docker-compose/.env
```diff
@@ -1,3 +1,15 @@
...
+
+VERSION_OPA=1.4.2
+VERSION_VAULT=1.19.3
```
dinkar-jain, line 4: Let's update the versions to "latest". Is there any particular reason for these?

docker-compose/.env
tdiesler: Done

docker-compose/.env
tdiesler: For reproducibility (i.e. go back in time to an arbitrary commit), these versions would need to be stable

dinkar-jain (approved):

PR #1092, opened 2025-05-29
Wal 704

mikeplotean:
## Description

Implements status-policy as a generalized credential status policy supporting W3C (BitstringStatusList, StatusList2021, RevocationList2020) and IETF (TokenStatusList<sup>1</sup>) status verification.

<sup>1</sup> - only JWT format supported, CWT is not supported

## Type of Change

## Checklist

## Breaking

- \-

### DIFF
--- a/waltid-libraries/credentials/waltid-verification-policies/src/jvmMain/kotlin/id/walt/policies/policies/StatusPolicy.jvm.kt
+++ b/waltid-libraries/credentials/waltid-verification-policies/src/jvmMain/kotlin/id/walt/policies/policies/StatusPolicy.jvm.kt
@@ -1,63 +1,16 @@
 package id.walt.policies.policies
 
-import id.walt.policies.policies.status.CredentialFetcher
-import id.walt.policies.policies.status.bit.BitValueReaderFactory
-import id.walt.policies.policies.status.content.JsonElementParser
-import id.walt.policies.policies.status.content.JwtParser
-import id.walt.policies.policies.status.entry.IETFEntryExtractor
-import id.walt.policies.policies.status.entry.W3CEntryExtractor
-import id.walt.policies.policies.status.model.*
-import id.walt.policies.policies.status.reader.IETFJwtStatusValueReader
-import id.walt.policies.policies.status.reader.W3CStatusValueReader
-import id.walt.policies.policies.status.validator.IETFStatusValidator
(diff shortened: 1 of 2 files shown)


waltid-libraries/credentials/waltid-verification-policies/src/jvmMain/kotlin/id/walt/policies/policies/StatusPolicy.jvm.kt
```diff
@@ -0,0 +1,127 @@
...
+    private val w3cStatusReader = W3CStatusValueReader(jwtParser)
+    @Transient
+    private val ietfJwtStatusReader = IETFJwtStatusValueReader(jwtParser)
+    @Transient
+    private val bitValueReaderFactory = BitValueReaderFactory()
+    @Transient
+    private val w3cStatusValidator = W3CStatusValidator(credentialFetcher, w3cStatusReader, bitValueReaderFactory)
+    @Transient
+    private val ietfStatusValidator = IETFStatusValidator(credentialFetcher, ietfJwtStatusReader, bitValueReaderFactory)
+    //endtodo
```
hokorn: I would suggest moving all transient variables and implementation to an object (object StatusPolicyImplementation)
The implementation then only delegates the verify to the implementation object. This would prevent creating all the helper object when deserializing a StatusPolicy. Also the HttpClient would be created once.

```
@Serializable
actual class StatusPolicy : StatusPolicyMp() {

    @JvmBlocking
    @JvmAsync
    actual override suspend fun verify(data: JsonObject, args: Any?, context: Map<String, Any>): Result<Any> {
        return StatusPolicyImplementation.verify(data, args)
    }
}

object StatusPolicyImplementation {

    //todo: inject through constructor (not allowed by interface atm - needs serializable)
    private val httpClient = HttpClient {
        install(ContentNegotiation) {
            json(Json { ignoreUnknownKeys = true })
        }
    }

    private val credentialFetcher = CredentialFetcher(httpClient)
    .....
}
```

mikeplotean: do you mean to have all dependencies in this singleton and reference it in both status-policy and revocation-policy?

PR #1101, opened 2025-06-04
Fix #1100 (Throwable with cause=null) in RevocationPolicy.jvm.kt

### LINKED ISSUE #1100: Throwable with cause=null when validating invalid RevocationList
## Description
When validating using RevocationPolicy of an invalid Revocation List, the resulting `Result` can contain a `Throwable` without any information or cause.

E.g.

Image

## To Reproduce

1. Use `RevocationPolicy` with `Verifier` on a revocation list that is not encoded corrently (or at all).
2. Make sure resulting error within the policy verification results contains an empty `Throwable`.

## Expected behavior
The Throwable should contain a valid `cause` containing the cause of the error. 

## Actual behavior
Null is used as the cause of the throwable.

## Environment
- OS version: Linux (Fedora 42)
- Java version: 21
- Waltid policies version: `id.walt.policies:waltid-verification-policies-jvm:1.0.2505271309-SNAPSHOT`

## Fix

Line 59 in RevocationPolicy.jvm.kt:
Image

`it.cause` is null:

Image

Image

This is because it refers to itself, Kotlin returns null when this happens. (See [Why does Throwable.getCause check to see if 'cause' is 'this' before returning `null` instead of just returning the cause directly?](https://stackoverflow.com/questions/64850746/why-does-throwable-getcause-check-to-see-if-cause-is-this-before-returning))

**The fix**

Change Line 59 in RevocationPolicy.jvm.kt to use `it` instead of `it.cause`.

daniel-dewa:
## Description

Read the issue #1100 for more info.

## Type of Change

- [X] bug fix - change which fixes an issue

One line changed.

### DIFF
--- a/.run/Issuer.run.xml
+++ b/.run/Issuer.run.xml
@@ -1,11 +1,11 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Issuer" type="JetRunConfigurationType">
-    <option name="ALTERNATIVE_JRE_PATH" value="corretto-21" />
+    <option name="ALTERNATIVE_JRE_PATH" value="17" />
     <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="id.walt.issuer.MainKt" />
     <module name="waltid-identity.waltid-services.waltid-issuer-api.main" />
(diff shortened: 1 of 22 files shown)


daniel-dewa:
Closing PR because of merge conflicts. It will be up to walt.id to implement a fix.

PR #1121, opened 2025-06-20
fix/api.json: fixed openApi metadata so a kotlin client can be genera…

hokorn:
## Description

[Bug description](https://linear.app/walt-id/issue/WAL-1410/bug-can-not-generate-client-from-apijson-wallt-issuer-and-verifier)

Fixes for Swagger api.json so a Kotlin client can be generated from open API json

## Type of Change

- [X] bug fix - change which fixes an issue

## Checklist

- [X] code cleanup and self-review

## Breaking

- \-

### DIFF
--- a/waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/web/modules/OpenApiModule.kt
+++ b/waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/web/modules/OpenApiModule.kt
@@ -4,18 +4,8 @@ import com.sksamuel.hoplite.simpleName
 import id.walt.commons.config.statics.BuildConfig
 import id.walt.commons.config.statics.ServiceConfig
 import io.github.smiley4.ktoropenapi.OpenApi
-import io.github.smiley4.ktoropenapi.config.ExampleEncoder
-import io.github.smiley4.ktoropenapi.config.InfoConfig
-import io.github.smiley4.ktoropenapi.config.OpenApiPluginConfig
-import io.github.smiley4.ktoropenapi.config.SchemaGenerator
-import io.github.smiley4.ktoropenapi.config.descriptors.AnyOfTypeDescriptor
-import io.github.smiley4.ktoropenapi.config.descriptors.ArrayTypeDescriptor
-import io.github.smiley4.ktoropenapi.config.descriptors.EmptyTypeDescriptor
-import io.github.smiley4.ktoropenapi.config.descriptors.KTypeDescriptor
-import io.github.smiley4.ktoropenapi.config.descriptors.RefTypeDescriptor
-import io.github.smiley4.ktoropenapi.config.descriptors.SerialTypeDescriptor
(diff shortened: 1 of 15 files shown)


waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeController.kt
```diff
@@ -232,9 +231,8 @@ fun Application.exchange() = walletRoute {
             }
             response {
                 HttpStatusCode.OK to {
-                    body<CredentialOffer> {
-                        description = "Resolved credential offer"
-                    }
+                    description = "Resolved credential offer"
+                    body<CredentialOffer> {}
```
mikeplotean: just wondering..
why body here is called with an empty lambda, while above (body<String>) it's a regular function call
noticed these for Map as well

mikeplotean (approved):
I think this is a good opportunity to define a standardized way of creating the api routes, smth. like, every api input / output:
-  should contain examples
- should have description at the root level, same as the type
  - description should not be nested inside the type level
- and so on..

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/web/controllers/exchange/ExchangeController.kt
hokorn: No special reason, I think it makes no difference

waltid-services/waltid-e2e-tests/src/test/kotlin/Draft11.kt
```diff
@@ -146,7 +146,7 @@ class Draft11(private val client: HttpClient)  {
 
         val signingKey = JWKKey.importJWK(matchingKey.toString()).getOrThrow()
 
-        assertTrue {signingKey.verifyJws(authJarTokenRequest.request!!).isSuccess}
+        assertTrue (signingKey.verifyJws(authJarTokenRequest.request!!).isSuccess)
```
hokorn, line 149: Fixed compiler warning

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/web/modules/OpenApiModule.kt
```diff
@@ -208,4 +262,124 @@ object OpenApiModule {
...
+    }
+}
+
+
+/**
+ * This analyzer is needed for attributes like:
+ * @Contextual val id: Uuid
+ */
+private object ContextualSerializationTypeAnalyzerModule : SerializationTypeAnalyzerModule {
+
```
hokorn, line 342: This Analyzer is needed, because we have some attributes like `@Contextual val id: Uuid` for instance in the [AccountsService.kt](https://github.com/walt-id/waltid-identity/blob/main/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/account/AccountsService.kt)

PR #1103, opened 2025-06-04
Fix bad decoding of RevocationPolicy bitstring #1075

daniel-dewa:
## Description

Changes decoding to use base64url while skipping the first character (that is used as marker by multibase).
See #1075 for more info.

## Type of Change

- [X] bug fix - change which fixes an issue
- [ ] new feature - change which adds functionality

## Checklist

- [ ] code cleanup and self-review
- [ ] unit + e2e test coverage **TODO**
- [ ] documentation updated accordingly

PR #1123, opened 2025-06-23
BitstringStatusList fix: drop first character of decoding

daniel-dewa:
## Description

Current implementation/decoder will fail on _some_ multibase-base64url strings, e.g. `udGVzdDIy`

Change decoder line to drop first character.

More info https://github.com/walt-id/waltid-identity/pull/1103#issuecomment-2996405359

relates to WAL-1532


mikeplotean:
This is handled by https://github.com/walt-id/waltid-identity/pull/1132
thanks for your contribution!

PR #1153, opened 2025-07-15
fix: VCT URL resolution

dinkar-jain:


chsavvaidis (changes requested):
does the vctUrl.pathname includes a leading "/" slash, and the string literal "/.well-known/vct/" also ends with a "/". This results in a double slash (//) in the final URL path e.g. https://issuer.portal.test.waltid.cloud/.well-known/vct//draft13/BankId

chsavvaidis (approved):

PR #1194, opened 2025-08-13
Fix/api image working dir

dinkar-jain:
Added "workingDirectory" to the Gradle Jib configuration to maintain the same volume mounting behavior when building the Docker image with Ktor instead of a Dockerfile.

### DIFF
--- a/waltid-services/waltid-issuer-api/build.gradle.kts
+++ b/waltid-services/waltid-issuer-api/build.gradle.kts
@@ -236,6 +236,9 @@ ktor {
         )
     }
     jib {
+        container {
+            workingDirectory = "/waltid-issuer-api"
+        }
         from {
             platforms {
                 platform {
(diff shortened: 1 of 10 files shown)


waltid-services/waltid-issuer-api/build.gradle.kts
```diff
@@ -236,6 +236,9 @@ ktor {
         )
     }
     jib {
+        container {
+            workingDirectory = "/waltid-issuer-api"
```
mikeplotean, line 240: could you mention why this is required
if I recall correctly, yesterday there was a fix where these paths were removed
is it a different thing? don't the other workflows need it too, e.g. compose-testing?

mikeplotean: perhaps in the pr description

mikeplotean (approved):

waltid-services/waltid-issuer-api/build.gradle.kts
dinkar-jain: Done

Issue #1241, opened 2025-09-18
Signer may incorrectly add  JWT header 'kid'


tdiesler:
Proof JWT headers must contain either jwk or kid

We currently always add the kid header

JakeFernandes98:
Thanks for raising this issue! We have started tracking this on our internal task board and an engineer will pick this up soon. We have taken this on lower priority, but if you have a project that relies on this please reach out to discuss options for support contracts for dedicated support via https://walt.id/contact

philpotisk:
@tdiesler Your PR https://github.com/walt-id/waltid-identity/pull/1242 got successfully merged, thanks

PR #1242, opened 2025-09-18
Signer may incorrectly add JWT header 'kid'

tdiesler:
See #1241


tdiesler:
now handled like this

```
                headers.containsKey("jwk") -> {
                    val jwk = headers.getValue("jwk").jsonObject
                    val kid = jwk["kid"]?.jsonPrimitive?.content
                    if (kid != null && kid != alias)
                        throw IllegalArgumentException("JWT headers.jwk.kid must match alias: $alias")
                }

```

PR #1253, opened 2025-09-29
Alias when generating or importing a key to the wallet

SuperBatata:
## Description

## Type of Change

## Checklist

- [ ] code cleanup and self-review

## Breaking

- \-

### DIFF
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
@@ -35,11 +35,7 @@ import id.walt.oid4vc.responses.TokenResponse
 import id.walt.webwallet.FeatureCatalog
 import id.walt.webwallet.config.KeyGenerationDefaultsConfig
 import id.walt.webwallet.config.RegistrationDefaultsConfig
-import id.walt.webwallet.db.models.WalletCategoryData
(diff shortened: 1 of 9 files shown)


waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -649,6 +650,24 @@ class SSIKit2WalletService(
                 throw ConflictException("Key with ID $keyId already exists in the database")
             }
 
+            val alias: String? = runCatching {
+                val trimmed = jwkOrPem.trim()
+                if (trimmed.startsWith("{")) {
```
mikeplotean: is there a particular reason to check this?
imho it's enough to just run catching, no need for the conditional
also, I think it can be optimized, e.g. avoid reading content twice:
```kotlin
val alias: String? = runCatching {
    val trimmed = jwkOrPem.trim()
    val json = Json.parseToJsonElement(trimmed).jsonObject
    json["alias"]?.jsonPrimitive?.contentOrNull?.takeIf { it.isNotBlank() }
        ?: json["name"]?.jsonPrimitive?.contentOrNull?.takeIf { it.isNotBlank() }
        ?: json["jwk"]?.jsonObject?.let { jwk ->
            jwk["alias"]?.jsonPrimitive?.contentOrNull
                ?: jwk["name"]?.jsonPrimitive?.contentOrNull
        }
}.getOrNull()
```
or if the switch option is preferred, then maybe have the content stored in `val`s

PR #1260, opened 2025-10-02
Session Management and Token Security Improvements (fixes wal-68)

philpotisk:
# Session Management and Token Security Improvements (Fixes wal-68)

This PR addresses critical security issues discovered during penetration testing related to session management and bearer token handling in the Walt.Id Wallet service.

## 🔒 Issues Addressed

1. **Logout doesn't invalidate tokens** - Tokens remained valid even after logout
2. **Token expiration only in days** - No support for minutes/seconds granularity  
3. **No refresh token mechanism** - No way to refresh tokens without re-authentication
4. **No idle timeout** - No automatic logout for inactive users

### 1. Token Blacklisting Service
- **File**: `TokenBlacklistService.kt`
- Implements server-side token blacklisting
- Tokens are invalidated immediately upon logout
- Automatic cleanup of expired tokens
- Thread-safe implementation using coroutines

### 2. Flexible Token Expiration
- **File**: `TimeUtils.kt`
- Support for various time formats:
  - `"15m"` or `"15 minutes"` → 15 minutes
  - `"2h"` or `"2 hours"` → 2 hours  
  - `"1d"` or `"1 days"` → 1 day
  - `"30s"` or `"30 seconds"` → 30 seconds
  - `"3600"` → 3600 seconds (numeric fallback)

### 3. Refresh Token Mechanism
- **Files**: `RefreshTokenController.kt`, updated `AuthController.kt`
- Access tokens with short expiration (default: 15 minutes)
- Refresh tokens with longer expiration (default: 7 days)
- Automatic token rotation for security
- Seamless token refresh without user intervention

### 4. Idle Timeout Implementation
- **File**: `tokenManager.ts`
- Automatic logout after configurable idle time (default: 30 minutes)
- Activity detection (mouse, keyboard, touch events)
- Graceful token refresh before timeout

### 5. Enhanced Frontend Token Management
- **Files**: `tokenManager.ts`, `apiClient.ts`, updated `authentication.ts`
- Automatic token refresh
- Idle timeout handling
- Secure token storage
- API client with automatic authentication

### Backend Configuration
Update `config/auth.conf`:

```hocon
# Access token lifetime (15 minutes)
tokenLifetime = "15m"

# Refresh token lifetime (7 days)  
refreshTokenLifetime = "7d"

# Idle timeout in minutes (30 minutes)
idleTimeoutMinutes = 30
```

### Frontend Configuration
The frontend automatically uses the backend configuration. The idle timeout can be customized in `tokenManager.ts`:

[...]

PR #1262, opened 2025-10-03
Webwallet logout

### LINKED ISSUE #1246: Logout not working: bearer token still valid after logout
## Describe the bug
The bearer token is still valid and can be used after calling /wallet-api/auth/logout endpoint.

## To Reproduce

1. Login using /wallet-api/auth/login endpoint. Method does not matter, could be email, oidc...
2. Check that the user is logged in and can fetch account data by calling /wallet-api/auth/user-info
3. Logout using /wallet-api/auth/logout endpoint
4. Call /wallet-api/auth/user-info endpoint using earlier acquired token. 

Examplary call that was used:
`curl -X 'GET'   'https://wallet.demo.walt.id/wallet-api/auth/user-info'   -H 'accept: application/json' -H "Authorization: Bearer eyJhbGciOi..."`

## Expected behavior

When calling /wallet-api/auth/user-info in step 4 I am expecting to receive "User nod logged in" o "Invalid bearer token" response

## Actual behavior

Bearer token is still valid and all user data is returned.

## Environment

Tested both on publicly deployed API and locally on latest cloned git repo.

waltkb:
# Preamble
The Web Wallet Backend contains two authentication backends:
1. legacy (internally sometimes referred to as just "auth"): The initial custom authentication system.
2. ktor-authnz: The new authentication system built for the enterprise stack.

Initially, the first authentication backend ("legacy") was written in a simple and safe way issuing & validating opaque tokens (random string that is in the backend internally mapped to a user), set with a httpOnly cookie.  When a HTTP request is received, the internal mapping is queried for the login token, and if valid, for the mapped user.
Initially, an in-memory backend storage was used (for having essentially no latency), however, for a (at this point in time unsupported) setup of a customer, a distributed login was required. Instead of going with the recommended route (moving the token handling to a Redis database), the login was at some point rewritten and based on JWT login tokens, which is an oxymoron combination of words that should not exist. No matter what you try, JWTs are never a good fit for login tokens, even with short expiration time and token rotation (one would still need token blacklisting, and at this point the whole reason for JWTs went out of the window).

When this was rewritten with the introduction of JWTs, the login system obviously immediately inherits all the downsides of JWTs. This is fine, provided that everyone is aware of them (the biggest downside being that JWTs officially do not have any kind of logout support), however this does not appear to be the case. The JWT system was not chosen as part of a trade-off assessment or consideration, but to fit into a customer checkbox requirement.

Additionally, from all I can see the implementation of the JWT based login is broken in various places, for example:
- while the expiration token can be configured in a configuration file and is even set within the token, I can find no check on the claim within the token verification, not in the current state nor any historic states of the code.
- while various people touched the file, and even added additional claims to the JWT (adding the claims to the JWT then issued during login), no-one has ever added a check on the opposite side when verifying the token, which means that all the additional claims (besides `exp`, e.g. `iat`, `aud`) (added by multiple different people) are never verified.

The new login system based on waltid-ktor-authnz was introduced for web3 authentication in the past and can already be utilized for that.

See the references regarding JWT for login session management below:
- https://blog.muhib.me/why-you-should-not-use-jwt-for-authentication#heading-disadvantages-of-jwt <- exact problems displayed in current Web Wallet
- https://apibakery.com/blog/tech/no-jwt/ <- showing how the previously suggested fix is strongly recommend against
- https://gist.github.com/samsch/0d1f3d3b4745d778f78b230cf6061452 <- further explanations
- Diagram that shows all the creative but invalid ways people try to work around the limitations of JWT: http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/ & http://cryto.net/%7Ejoepie91/blog/2016/06/19/stop-using-jwt-for-sessions-part-2-why-your-solution-doesnt-work/

# Changes
This PR extends the waltid-ktor-authnz integration (called "ktor-authnz" in the Web Wallet) to also be usable for the password login flow (besides the already existing web3 login flow), so that the legacy authentication system can be circumvented entirely. This does not require changes in the flow of waltid-ktor-authnz, as it already implements this (and others) flow, in fact being the primary login flow of the Enterprise stack. Instead, the Web Wallet is extended to be able to also use the flow now. However, an additional token and session store was added to waltid-ktor-authnz to fit with the deployment requirements of the Web Wallet.

This new behavior has to be enabled:

`_features.conf`:
```hocon
enabledFeatures = [
    # external-signature-endpoints,
    # trusted-ca,
    # entra,
    ktor-authnz, # <-- Enable authentication backend ktor-authnz based on waltid-ktor-authnz
    # dev-mode
    # ...
]
disabledFeatures = [
    auth   # (legacy auth)  # <-- Disable old legacy authentication backend
]
```

`ktor-authnz.conf`:
```hocon
######## UNCHANGED - START ########

# Will secure login cookies with `Secure` context, enable HTTS and HTTP->HTTPS redirect
requireHttps = false


# Provide pepper to use for additional password salting (unique string for your deployment,
# has to be shared between instances).
pepper = "waltid"

# Hash algorithm to use for passwords for signing.
# You can choose from algorithms like: ARGON2, PBKDF2, PBKDF2_COMPRESSED, BCRYPT, SCRYPT, BALLON_HASHING, MESSAGE_DIGEST, NONE
hashAlgorithm = ARGON2

# Configure the Auth Flow (refer to: waltid-ktor-authnz)
authFlow = {
    method: email
    expiration: "7d" # optional: Set expiration time for login tokens, e.g. a week
    ok: true # Auth flow ends successfully with this step
}

cookieDomain = null

# If you previously used other (older) password hash algorithms, you
# can use this function to migrate old hashes to new hash algorithms. This
# works at login-time: When a user logs in with a password that uses a hash algorithm
# on this list, the password will be re-hashed in the specified replacement algorithm.
# If null is used as hash algorithm selector, all algorithms expect for the target
# algorithm will be converted automatically.
hashMigrations = {
    MESSAGE_DIGEST: ARGON2 # E.g.: Convert all the MD5 hashes to Argon2 hashes
}

######## UNCHANGED - END ########

# Setup how to issue and verify tokens
# Supported:
# - STORE_IN_MEMORY: In memory token store (single-instance, no configuration necessary)
# - STORE_VALKEY: Store in Redis/Valkey/KeyDB (multi-instance, distributed logout supported) - configure Redis/Valkey/KeyDB instance below
# - JWT: Tokens as stateless JWT (multi-instance, no logout support as it is stateless!) - configure keys below
tokenType = STORE_VALKEY

######## UNCHANGED - START ########
# --- Required for JWK tokens:

# Key (all waltid-crypto supported) to sign login token - has to be key allowing signing (private key)
# signingKey = {"type": "jwk", "jwk": {"kty": "OKP", "d": "z8Lk85rAtfv2RJN_cD_-9nqHHwKTlTQ5_I53LcsHjC4", "use": "sig", "crv": "Ed25519", "x": "Ew76rQJ9gPHCOBOwJlf__Il5IjgSAc3bQ_a8psd-F3E", "alg": "EdDSA"}}

# Key (all waltid-crypto supported) to verify incoming login tokens - public key is ok.
# verificationKey = {"type": "jwk", "jwk": {"kty": "OKP", "d": "z8Lk85rAtfv2RJN_cD_-9nqHHwKTlTQ5_I53LcsHjC4", "use": "sig", "crv": "Ed25519", "x": "Ew76rQJ9gPHCOBOwJlf__Il5IjgSAc3bQ_a8psd-F3E", "alg": "EdDSA"}}
######## UNCHANGED - END ########

# --- Required for Token Store tokens (opaque tokens):

# - Host:
# valkeyUnixSocket = "..."
valkeyHost = "127.0.0.1"
valkeyPort = 6379

# - Authentication: (if server auth is enabled)
# valkeyAuthUsername = "" # optional
# valkeyAuthPassword = ""

# - Expiration of tokens:
valkeyRetention = "7d"
```

As you can see, the most important point is the new option to configure the token store:
```hocon
# Setup how to issue and verify tokens
# Supported:
# - STORE_IN_MEMORY: In memory token store (single-instance, no configuration necessary)
# - STORE_VALKEY: Store in Redis/Valkey/KeyDB (multi-instance, distributed logout supported) - configure Redis/Valkey/KeyDB instance below
# - JWT: Tokens as stateless JWT (multi-instance, no logout support as it is stateless!) - configure keys below
tokenType = STORE_VALKEY
```
In the example you can see that with this current state, three options for storing tokens (and to some degree also sessions) are available:
- In memory: The token & session is stored in memory. This is a valid approach (and all features of an authentication system, e.g. logout, can be provided with this architecture) and also the simplest (requiring no configuration). However, obviously this limits the authentication to work on a single instance of the application.
- Valkey: The token & session is stored in a Redis/Valkey/KeyDB compatible data store. This is (/ can be, depending on db latency) still highly performant, and still a valid full-featured approach (distributed logout etc.), and will now also work over multiple distributed instances (of course as long as they access the same data store).
- JWT + session store: This is the non-recommended approach where the login tokens are issued as stateless JSON Web Tokens. This should only be used in circumstances when you have legacy infrastructure that depends on it (albeit even with this they it at least be moved to PASETO v4 instead of JWT, however as we all know, legacy infrastructure is hard to move forward). This has the typical downsides of JWTs (no control over them once issued as not stored), so this of course will apply various downsides of JWTs: no client-controlled logout, no blacklisting of user sessions, massively reduced storage security (easily exploited through Javascript XSS attacks) when token rotation is added, manually or through extra infrastructure handle requests to manually add tokens (when using token rotation). Additionally, the session still has to be stored (as this is a core building block and as such requirement of waltid-ktor-authnz), so an additional session store (which is then different from the token store) has to be utilized. For example, the token type can be JWT, but then a different session store (e.g. in-memory) has to be used. However, even when using the JWT token type, at least the `exp` claim is indeed actually checked with the waltid-ktor-authnz backed authentication system.

### Frontend changes
No frontend logic changes of any kind were required in the Frontend, only the URL needs to be replaced (the frontend shall send a signIn request to the legacy authentication backend to `POST /wallet-api/auth/login`, and a request to the ktor-authnz authentication backend to `POST /wallet-api/auth/account/emailpass`.

To allow easy doing of such, a config switch was introduced: When env.NUXT_PUBLIC_AUTH_USE_KTORAUTHNZ is set to true, /account/emailpass is used, otherwise the old /login is used in the authentication configuration.

To set the new variable, one can use the usual ways:
- for local deployment: Append the line `NUXT_PUBLIC_AUTH_USE_KTORAUTHNZ=true` to the `.env` file
- docker run: `docker run -e NUXT_PUBLIC_AUTH_USE_KTORAUTHNZ=true <...>`
- manual prefixing of commands with `NUXT_PUBLIC_AUTH_USE_KTORAUTHNZ=true `

# Result

When using a token storage system that supports maintaining state (currently configuring either `tokenType = STORE_VALKEY` or `tokenType = STORE_IN_MEMORY`), all logout-related issues are immediately resolved.

# References

- Fixes #1246
- Fixes wal-68
- Closes #1259
  - This PR does not change the backend handling for the tokens and as such cannot address the referenced issues.
- Closes #1260 
  - This PR was AI generated and contains several issues; contains review

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@
 
 Available for Kotlin/Java and JavaScript environments.
 
-- **Crypto** ([Docs](https://docs.oss.walt.id/issuer/sdks/manage-keys/overview) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-libraries/crypto/waltid-crypto)) -
+- **Crypto** ([GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-libraries/crypto/waltid-crypto)) -
   create and use keys based on different algorithms and KMS backends (in-memory, AWS, Hashicorp TSE, OCI)
-- **DID** ([Docs](https://docs.oss.walt.id/issuer/sdks/manage-dids/overview) | [GitHub](https://github.com/walt-id/waltid-identity/blob/main/waltid-libraries/waltid-did/README.md)) -
+- **DID** ([GitHub](https://github.com/walt-id/waltid-identity/blob/main/waltid-libraries/waltid-did/README.md)) -
   create, register, and resolve DIDs on different ecosystems.
-- **W3C Credentials** ([Docs](https://docs.oss.walt.id/issuer/sdks/manage-credentials/overview) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-libraries/credentials/waltid-w3c-credentials)) -
+- **W3C Credentials** ([GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-libraries/credentials/waltid-w3c-credentials)) -
   issue and verify W3C credentials as JWTs and SD-JWTs.
 - **mdoc Credentials** ([GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-libraries/credentials/waltid-mdoc-credentials)) -
   issue and verify mdoc credentials (mDL ISO/IEC 18013-5).
@@ -35,11 +35,11 @@ Available for Kotlin/Java and JavaScript environments.
 
 A set of APIs to build issuer, verifier, and wallet capabilities into any app.
 
-- **Issuer API** ([Docs](https://docs.walt.id/issuer/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-issuer-api)) -
+- **Issuer API** ([Docs](https://docs.walt.id/community-stack/issuer/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-issuer-api)) -
   enable apps to issue credentials (W3C JWTs and SD-JWTs) via OID4VC.
-- **Verifier API**  ([Docs](https://docs.walt.id/verifier/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-verifier-api)) -
+- **Verifier API**  ([Docs](https://docs.walt.id/community-stack/verifier/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-verifier-api)) -
   enable apps to verify credentials (W3C JWTs and SD-JWTs) via OID4VP/SIOPv2.
-- **Wallet API** ([Docs](https://docs.oss.walt.id/wallet/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-wallet-api)) -
+- **Wallet API** ([Docs](https://docs.walt.id/community-stack/wallet/api/getting-started) | [GitHub](https://github.com/walt-id/waltid-identity/tree/main/waltid-services/waltid-wallet-api)) -
   extend apps with wallet capabilities to collect, store, manage and share identity credentials and tokens.
 
 ### Apps
@@ -91,11 +91,11 @@ cd waltid-applications/waltid-cli && ./waltid-cli.sh
 The walt.id identity repo is part of The Community Stack, walt.id's collection of open-source products to build identity and wallet
 solutions. Learn more [here](https://walt.id/blog/p/community-stack).
 
-![waltid-identity-architecture](https://github.com/user-attachments/assets/98c020fe-dc37-46fd-9886-613ee8fc8760)
+![waltid-identity-architecture](https://github.com/user-attachments/assets/56a69598-c9f0-4f4a-a071-05fb98d247ba)
 
 ## Join the community
 
-* Connect and get the latest updates: [Discord](https://discord.gg/AW8AgqJthZ) | [Newsletter](https://walt.id/newsletter) | [YouTube](https://www.youtube.com/channel/UCXfOzrv3PIvmur_CmwwmdLA) | [Twitter](https://mobile.twitter.com/walt_id)
+* Connect and get the latest updates: [Discord](https://discord.gg/AW8AgqJthZ) | [Newsletter](https://walt.id/newsletter) | [YouTube](https://www.youtube.com/channel/UCXfOzrv3PIvmur_CmwwmdLA) | [X](https://x.com/walt_id)
 * Get help, request features and report bugs: [GitHub Issues ](https://github.com/walt-id/waltid-identity/issues)
 
 ## License
--- a/waltid-applications/waltid-cli/README.md
+++ b/waltid-applications/waltid-cli/README.md
@@ -412,7 +412,7 @@ Arguments:
 ```
 
 Before signing a VC, we need
-to [create a VC](https://docs.oss.walt.id/issuer/sdks/manage-credentials/sign/w3c-credential), which can be done
+to [create a VC](https://docs.walt.id/community-stack/issuer/sdks/manage-credentials/sign/w3c-credential), which can be done
 manually or programmatically.
 
 Since our focus here is on the CLI and not the Walt.id API, let's see what it's like to manually create a VC.
@@ -420,7 +420,7 @@ Since our focus here is on the CLI and not the Walt.id API, let's see what it's 
 #### 1. Create a VC
 
 Walt.id
-provides [a repository of VCs](https://docs.oss.walt.id/issuer/sdks/manage-credentials/sign/w3c-credential#manual-create-credential)
+provides [a repository of VCs](https://docs.walt.id/community-stack/issuer/sdks/manage-credentials/sign/w3c-credential#manual-create-credential)
 that can be used as templates for creating your own. Choose the one that best suits your needs.
 
 Let's choose the [OpenBadgeCredential](https://github.com/walt-id/waltid-identity/blob/main/waltid-applications/waltid-cli/src/jvmTest/resources/vc/openbadgecredential_sample.json), adjust it
--- a/waltid-applications/waltid-web-wallet/apps/waltid-demo-wallet/package.json
+++ b/waltid-applications/waltid-web-wallet/apps/waltid-demo-wallet/package.json
@@ -30,7 +30,7 @@
     "@vite-pwa/nuxt": "^1.0.4",
     "@vueuse/core": "^13.6.0",
     "@vueuse/nuxt": "13.6.0",
-    "nuxt": "3.17.7",
+    "nuxt": "3.19.0",
     "nuxt-icon": "^1.0.0-beta.7",
     "typescript": "5.8.3"
   }
--- a/waltid-applications/waltid-web-wallet/apps/waltid-dev-wallet/package.json
+++ b/waltid-applications/waltid-web-wallet/apps/waltid-dev-wallet/package.json
@@ -32,7 +32,7 @@
     "@vite-pwa/nuxt": "^1.0.4",
     "@vueuse/core": "^13.6.0",
     "@vueuse/nuxt": "13.6.0",
-    "nuxt": "3.17.7",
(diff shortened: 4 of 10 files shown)


waltkb:
@dinkar-jain see Matrix PM for testing infos + next tasks

JakeFernandes98:
This is a good opportunity to create a readme that summaries the library and its usage (a.k.a pull in and update/extend the contents of the Quickstart.md and probably amendment markdown). Revise the other docs and see if they need to be updated as well. I would recommend you also reference the docs in the doc folder in the main readme so they developers know they're there and what each are used for (rather than having to guess off of the name or a skim read). 

This will make sure the other developers know how to consume this library properly and we don't run into this sort of issue again in the future @waltkb

dinkar-jain:
@waltkb
1. When I start the wallet API, create an account, and log in, my wallet appears. However, if I log out, restart the wallet API, and log in again with the same account, the wallet is missing. image
2. The backend isn't removing the login cookie during logout.
3. Since we’re creating a second cookie with Nuxt Auth (without the HTTP-only flag), why not make the backend cookie itself lack the HTTP-only flag? This would eliminate the Nuxt Auth dependency and dual JWT management (as done in PR #1259). Alternatively, we could add comments in both FE and BE configurations to sync cookie expiration times.
4. Update the demo wallet Nuxt config for Nuxt Auth, similar to the dev wallet update.

waltkb:
1. Do you have `recreateDatabaseOnStart = true` set in `db.conf`? Then your wallet is missing because it is deleted per your configuration. Nevertheless, the account tables have been included there too, so on a recreate the authnz accounts will also be dropped.
2. The cookie is invalidated on the backend, and the frontend drops it's cookie, so the logout happens with both sides being aware; the backend cookie is also flagged for deletion, but the browser appears to ignore that because the deleted cookie appended has less  extensions set. Nevertheless, the deleted cookie will now have the same extensions, so that also the invalidated unused cookie is deleted, just to be tidy.
3. This would cause loss of functionality from the auth library that is tightly integrated into Nuxt
4. Replicated to demo wallet

waltid-applications/waltid-web-wallet/apps/waltid-demo-wallet/src/app.vue
```diff
@@ -28,6 +28,21 @@ import "uno.css";
...
+
+    if (oldStatus == "loading" && newStatus == "unauthenticated") {
+        const path = route.fullPath
+        const toUrl = "/login?redirect=" + path
+        console.log("Redirecting to: " + toUrl)
+
+        navigateTo(toUrl)
+    }
+})
+
```
dinkar-jain, line 45: We don't need this; Nuxt auth handles it automatically. I’ve tested it without, and it works.

waltid-applications/waltid-web-wallet/apps/waltid-dev-wallet/src/app.vue
```diff
@@ -28,6 +28,21 @@ import "uno.css";
...
+
+    if (oldStatus == "loading" && newStatus == "unauthenticated") {
+        const path = route.fullPath
+        const toUrl = "/login?redirect=" + path
+        console.log("Redirecting to: " + toUrl)
+
+        navigateTo(toUrl)
+    }
+})
+
```
dinkar-jain, line 45: We don't need this; Nuxt auth handles it automatically. I’ve tested it without, and it works.

waltid-services/waltid-wallet-api/config/ktor-authnz.conf
```diff
@@ -33,3 +28,32 @@ cookieDomain = null
...
+# - Host:
+# valkeyUnixSocket = "..."
+valkeyHost = "127.0.0.1"
+valkeyPort = 6379
+
+# - Authentication: (if server auth is enabled)
+# valkeyAuthUsername = "" # optional
+# valkeyAuthPassword = ""
+
+# - Expiration of tokens:
```
dinkar-jain, line 58: Can we add a comment here to mention that if this is changed, it should also be updated in the frontend nuxt.config?

waltid-applications/waltid-web-wallet/apps/waltid-demo-wallet/src/app.vue
waltkb: I tested it too, and it did not work without it? Because I would have actually expected nuxt-auth to do this automatically

waltid-applications/waltid-web-wallet/apps/waltid-dev-wallet/src/app.vue
waltkb: I tested it too, and it did not work without it? Because I would have actually expected nuxt-auth to do this automatically

waltid-services/waltid-wallet-api/config/ktor-authnz.conf
waltkb: If nuxt-auth handles the auto-logout gracefully (as you write above, maybe I am doing something wrong when testing), then the two durations do not have to match, it has configurable auth session status refresh that constantly refreshes when you focus the window, even periodically is possible

See https://auth.sidebase.io/guide/application-side/configuration#sessionrefresh

dinkar-jain: This is related to the refresh token, which we don't support in the BE. If we did, we need to configure it here: [https://auth.sidebase.io/guide/local/quick-start#refresh](https://auth.sidebase.io/guide/local/quick-start#refresh), as the refresh handler calls `auth.refresh()`: [https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/utils/refreshHandler.ts](https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/utils/refreshHandler.ts).

PR #1263, opened 2025-10-07
fixed redirect url after verification

SuperBatata:
## Description

## Type of Change

## Checklist

- [ ] code cleanup and self-review

## Breaking

- \-

### DIFF
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
@@ -76,7 +76,7 @@ import org.jetbrains.exposed.v1.core.eq
 import org.jetbrains.exposed.v1.jdbc.insert
 import org.jetbrains.exposed.v1.jdbc.selectAll
 import org.jetbrains.exposed.v1.jdbc.transactions.transaction
-import java.util.Base64
+import java.util.*
 import kotlin.time.Clock
 import kotlin.time.Duration.Companion.seconds
 import kotlin.time.ExperimentalTime
@@ -267,6 +267,36 @@ class SSIKit2WalletService(
         }
         logger.debug { "HTTP Response: $resp, body: $httpResponseBody" }
 
+        var redirectFromBody: String? = null
+        var errorUriFromBody: String? = null
+        if (httpResponseBody != null && httpResponseBody.couldBeJsonObject()) {
+            val bodyJson = httpResponseBody.parseAsJsonObject().getOrNull()
+            val bodyRedirect = bodyJson?.get("redirect_uri")?.jsonPrimitive?.contentOrNull
+            val bodyError = bodyJson?.get("error_uri")?.jsonPrimitive?.contentOrNull
+            @Suppress("HttpUrlsUsage")
+            if (bodyRedirect != null && (bodyRedirect.startsWith("http://") || bodyRedirect.startsWith("https://"))) {
+                redirectFromBody = bodyRedirect
+            }
+            @Suppress("HttpUrlsUsage")
+            if (bodyError != null && (bodyError.startsWith("http://") || bodyError.startsWith("https://"))) {
+                errorUriFromBody = bodyError
+            }
+        } else if (isResponseRedirectUrl) {
+            redirectFromBody = httpResponseBody
+        }
+
+        if (redirectFromBody != null) {
+            return Result.success(redirectFromBody)
+        }
+        if (errorUriFromBody != null) {
+            return Result.failure(
+                PresentationError(
+                    message = "Presentation failed - redirecting to error page",
+                    redirectUri = errorUriFromBody
+                )
+            )
+        }
+
         parameter.selectedCredentials.forEach {
             credentialService.get(walletId, it)?.run {
                 eventUseCase.log(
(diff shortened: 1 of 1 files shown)


waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -267,6 +267,36 @@ class SSIKit2WalletService(
...
         logger.debug { "HTTP Response: $resp, body: $httpResponseBody" }
 
+        var redirectFromBody: String? = null
+        var errorUriFromBody: String? = null
+        if (httpResponseBody != null && httpResponseBody.couldBeJsonObject()) {
+            val bodyJson = httpResponseBody.parseAsJsonObject().getOrNull()
+            val bodyRedirect = bodyJson?.get("redirect_uri")?.jsonPrimitive?.contentOrNull
+            val bodyError = bodyJson?.get("error_uri")?.jsonPrimitive?.contentOrNull
+            @Suppress("HttpUrlsUsage")
+            if (bodyRedirect != null && (bodyRedirect.startsWith("http://") || bodyRedirect.startsWith("https://"))) {
```
mikeplotean: perhaps startsWIth.. could be extracted into a separate method and used in both cases, body-redirect and body-error
e.g.
```kotlin
fun String.isUrl() = this.startsWith("http://") || this.startsWith("https://")
```
or maybe using regex:
```kotlin
fun String.isUrl(): Boolean {
    val urlPattern = "^https?://.*".toRegex(RegexOption.IGNORE_CASE)
    return urlPattern.matches(this)
}
```
but perhaps better:
```kotlin
fun String.isUrl() = runCatching {
        val url = URI.create(this).toURL()
        url.protocol in listOf("http", "https") && url.host.isNotEmpty()
    }.getOrDefault(false)
```

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -267,6 +267,36 @@ class SSIKit2WalletService(
         }
         logger.debug { "HTTP Response: $resp, body: $httpResponseBody" }
 
+        var redirectFromBody: String? = null
+        var errorUriFromBody: String? = null
+        if (httpResponseBody != null && httpResponseBody.couldBeJsonObject()) {
```
mikeplotean, line 272: this could be refactored for a cleaner view, imho, e.g.
```kotlin
private fun tryGetRedirect(
        httpResponseBody: String?,
        isResponseRedirectUrl: Boolean
    ): Pair<String?, String?> {
        var redirectFromBody: String? = null
        var errorUriFromBody: String? = null
        if (httpResponseBody != null && httpResponseBody.couldBeJsonObject()) {
            val bodyJson = httpResponseBody.parseAsJsonObject().getOrNull()
            val bodyRedirect = bodyJson?.get("redirect_uri")?.jsonPrimitive?.contentOrNull
            val bodyError = bodyJson?.get("error_uri")?.jsonPrimitive?.contentOrNull
            @Suppress("HttpUrlsUsage")
            if (bodyRedirect != null && (bodyRedirect.isUrl())) {
                redirectFromBody = bodyRedirect
            }
            @Suppress("HttpUrlsUsage")
            if (bodyError != null && (bodyError.isUrl())) {
                errorUriFromBody = bodyError
            }
        } else if (isResponseRedirectUrl) {
            redirectFromBody = httpResponseBody
        }
        return Pair(errorUriFromBody, redirectFromBody)
    }
```
and then call tryGetRedirect like:
```kotlin
val (redirectFromBody, errorUriFromBody) = tryGetRedirect(httpResponseBody, isResponseRedirectUrl)
```

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
```diff
@@ -267,6 +267,36 @@ class SSIKit2WalletService(
...
+            }
+            @Suppress("HttpUrlsUsage")
+            if (bodyError != null && (bodyError.startsWith("http://") || bodyError.startsWith("https://"))) {
+                errorUriFromBody = bodyError
+            }
+        } else if (isResponseRedirectUrl) {
+            redirectFromBody = httpResponseBody
+        }
+
+        if (redirectFromBody != null) {
```
mikeplotean, line 288: what if both `redirectFromBody` and `errorUriFromBody` are not null?

SuperBatata: this scenario can't exist afaik , as the verifier will return only one option depending on the verification result , however if this happens the code will return success with redirectFromBody and ignore errorUriFromBody. 
In usePresentationRequest, the checks are performed in this order:

1 - If redirectFromBody != null → Result.success(redirectFromBody) is returned immediately.
2 - Else if errorUriFromBody != null → Result.failure(PresentationError(..., redirectUri = errorUriFromBody)).

what do you think ?

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
SuperBatata: i dont see the need to extract it in a separate function as it's used only here

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/SSIKit2WalletService.kt
mikeplotean: was asking just hypothetically, it could be that the 'tryGetRedirect' implementation changes over time and it happens both end up not null; this might represent an additional state (a superposition xD) and maybe require specific handling, currently the handling is decoupled and easy to miss

but agree, currently, if both are not null, the redirect takes priority

PR #1302, opened 2025-10-28
Feature/wal 263 : Implement a new VP-level policy to enforce required credentials in a presentation

SuperBatata:
Currently, the vp_policies in waltid-verifier-api do not allow enforcing that multiple required credentials are presented together within a single verifiable presentation (VP).

We need to create a VP-level policy that ensures:

The presented VP includes a gx:LegalPerson credential.

The VP also includes a LRN credential, which must be one of:

gx:vatId

gx:EORI

gx:LEI.

This policy ensures that the verifier only accepts a presentation if it fulfills both mandatory and conditional credential requirements.

### DIFF
--- a/waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
+++ b/waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
@@ -26,11 +26,10 @@ class VpRequiredCredentialsPolicy : CredentialWrapperValidatorPolicy() {
     @JsPromise
     @JsExport.Ignore
     override suspend fun verify(data: JsonObject, args: Any?, context: Map<String, Any>): Result<Any> {
-        val argObj = (args as? JsonElement)?.jsonObject
-            ?: return Result.failure(IllegalArgumentException("vp_required_credentials: missing or invalid args (expected JSON object)"))
+        require(args is JsonObject) { "vp_required_credentials: missing or invalid args (expected JSON object)" }
 
-        val requirements = argObj["required"]?.jsonArray
-            ?: return Result.failure(IllegalArgumentException("vp_required_credentials: 'required' array missing in args"))
+        val requirements = args["required"]?.jsonArray
+        require(requirements is JsonArray) { "vp_required_credentials: 'required' array missing in args" }
 
         val presentedTypes: List<String> = collectPresentedTypes(data)
 
@@ -38,8 +37,6 @@ class VpRequiredCredentialsPolicy : CredentialWrapperValidatorPolicy() {
 
         requirements.forEach { reqEl ->
             val req = reqEl.jsonObject
-            val isRequired = req["required"]?.jsonPrimitive?.booleanOrNull ?: true
-            if (!isRequired) return@forEach
 
             when {
                 req.containsKey("credential_type") -> {
@@ -56,7 +53,7 @@ class VpRequiredCredentialsPolicy : CredentialWrapperValidatorPolicy() {
                         ?.filter { it.isNotBlank() }
                         ?: emptyList()
                     if (anyOf.isEmpty()) {
-                        return Result.failure(IllegalArgumentException("vp_required_credentials: 'any_of' must be a non-empty array of strings"))
+                        return@forEach
                     }
                     val foundAny = anyOf.any { t -> presentedTypes.any { it == t } }
                     if (!foundAny) missingMessages.add("Missing required credential: ${anyOf.joinToString(" OR ")}")
--- a/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/openapi/VerifierApiExamples.kt
+++ b/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/openapi/VerifierApiExamples.kt
@@ -47,9 +47,9 @@ object VerifierApiExamples {
                 "policy": "vp_required_credentials",
                 "args": {
                     "required": [
-                        { "credential_type": "gx:Issuer", "required": true },
-                        { "credential_type": "gx:LegalPerson", "required": true },
-                        { "any_of": [ "gx:EORI", "gx:LeiCode", "gx:VatID" ], "required": true }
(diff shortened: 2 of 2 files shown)


waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
```diff
@@ -0,0 +1,105 @@
...
+    override val description: String =
+        "Verifies that a Verifiable Presentation includes a set of required credential types (e.g., bankId and any of a group of diplomas)"
+    override val supportedVCFormats = setOf(VCFormat.jwt_vp, VCFormat.jwt_vp_json)
+
+    @JvmBlocking
+    @JvmAsync
+    @JsPromise
+    @JsExport.Ignore
+    override suspend fun verify(data: JsonObject, args: Any?, context: Map<String, Any>): Result<Any> {
+        val argObj = (args as? JsonElement)?.jsonObject
```
mikeplotean: could use:
```kotlin
require(args is JsonObject) { "vp_required_credentials: missing or invalid args (expected JSON object)" }
```
and use args below

waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
```diff
@@ -0,0 +1,105 @@
...
+            ?: return Result.failure(IllegalArgumentException("vp_required_credentials: missing or invalid args (expected JSON object)"))
+
+        val requirements = argObj["required"]?.jsonArray
```
mikeplotean: could use require here as well:
```kotlin
val requirements = args["required"]?.jsonArray
require(requirements is JsonArray) { "vp_required_credentials: 'required' array missing in args" }
```

waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
```diff
@@ -0,0 +1,105 @@
...
+            ?: return Result.failure(IllegalArgumentException("vp_required_credentials: 'required' array missing in args"))
+
+        val presentedTypes: List<String> = collectPresentedTypes(data)
+
+        val missingMessages = mutableListOf<String>()
+
+        requirements.forEach { reqEl ->
+            val req = reqEl.jsonObject
+            val isRequired = req["required"]?.jsonPrimitive?.booleanOrNull ?: true
```
mikeplotean: does it mean if 'required' is not present it defaults to `true`, meaning it is required?
especially considering the conditional from below

waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
```diff
@@ -0,0 +1,105 @@
...
+                    val found = presentedTypes.any { it.equals(credentialType, ignoreCase = false) }
+                    if (!found) missingMessages.add("Missing required credential: $credentialType")
+                }
+
+                req.containsKey("any_of") -> {
+                    val anyOf = req["any_of"]?.jsonArray?.mapNotNull { it.jsonPrimitive.contentOrNull }
+                        ?.filter { it.isNotBlank() }
+                        ?: emptyList()
+                    if (anyOf.isEmpty()) {
+                        return Result.failure(IllegalArgumentException("vp_required_credentials: 'any_of' must be a non-empty array of strings"))
```
mikeplotean: imho, an empty anyOf could also be valid

waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/openapi/VerifierApiExamples.kt
```diff
@@ -40,6 +40,22 @@ object VerifierApiExamples {
         ]
     """.trimIndent()
 
+    // language=json
+    private val vpRequiredCredentialsData = """
+        [
+            {
+                "policy": "vp_required_credentials",
+                "args": {
+                    "required": [
```
mikeplotean, line 49: wouldn't a different model fit better? e.g.:
```json
{
    "required":
    [
        "gx:Issuer",
        "gx:LegalPerson"
    ],
    "optional":
    [
        "gx:EORI",
        "gx:LeiCode",
        "gx:VatID"
    ]
}
```

SuperBatata: they are not optional , we are trying to verify this condition ( A and B and (X or Y or Z)) so if we use optional it becomes misleading

waltid-libraries/credentials/waltid-verification-policies/src/commonMain/kotlin/id/walt/policies/policies/vp/VpRequiredCredentialsPolicy.kt
SuperBatata: Exactly , as we are trying to enforce the presentation of the credential (which is making it required)

mikeplotean: then maybe remove the 'required' field, since it's already within the 'required' array?
my understanding, as a user, if I don't want to make a credential mandatory, then I just don't mention it in this policy, is this an option at all?

waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/openapi/VerifierApiExamples.kt
mikeplotean: my main concern here is about having 'required' mentioned 3 times:
1. within the policy name
2. the container array
3. the individual field
could the field be removed?
not sure I understand its purpose

Issue #1291, opened 2025-10-22
did:web resolution fails when identifier includes port and path (follow-up to #1142)


rodmenm:
## Describe the bug
When resolving a did:web identifier that includes both a port and a path, the resolver fails to correctly interpret the identifier.
This causes credential issuance to fail.

This is a follow-up to issue #1142 , which was closed but remains unresolved.

## To Reproduce
1. In your Docker environment, create a new did:web that includes a port and a path (for example:
did:web:domain%3A1100/api/v1)
2. Attempt to issue a credential from the VC repo and assign it to this DID.
3. The API service will fail to resolve the did:web.

## Expected behavior
The did:web identifier should be correctly resolved, and the credential should be successfully issued and assigned to the wallet.

## Actual behavior
The resolver misinterprets the colon (:) in the domain as a slash (/), preventing proper DID resolution and credential issuance.
For example, a the issuer did resolver parses it as did:web:domain:1100:api:v1 instead of did:web:domain%3A1100:api:v1.

## Environment

- OS version: Windows 11
- Docker version: 24.0.7

JakeFernandes98:
Fixed through  #1316 . Let me know if you have any further questions!

rodmenm:
Hi, I’ve tested this by changing the image versions in the docker-compose file and modifying the variable that allows resolving did:web using HTTP.
I’m attaching the wallet-api logs from when it’s launched with docker-compose.

```bash
2025-11-09 19:17:00 -- walt.id wallet [latest] --
2025-11-09 19:17:00 18:17:00.546879935  INFO [                main] :        Configuration : Configuration set using the DSL with append=true
2025-11-09 19:17:00 18:17:00.674560601  INFO [                main] :       walt.id wallet : Starting walt.id wallet [latest]...
2025-11-09 19:17:00 18:17:00.712944773  INFO [                main] :         Feature-Init : Initializing features...
2025-11-09 19:17:00 18:17:00.713878516  INFO [                main] :         Feature-Init : Registering common feature catalog...
2025-11-09 19:17:00 18:17:00.719555082  INFO [                main] :         Feature-Init : Registering service feature catalog...
2025-11-09 19:17:00 SLF4J(I): Connected with provider of type [io.klogging.slf4j.KloggingServiceProvider]
2025-11-09 19:17:00 18:17:00.722051799  INFO [                main] :         Feature-Init : Loading features...
[...]

PR #1316, opened 2025-11-04
HTTPs toggle for DID Web resolution

philpotisk:
## Description

This PR allows to toggle HTTP(s) for testing DID Web resolution locally in order to solve this issue https://github.com/walt-id/waltid-identity/issues/1291

### DIFF
--- a/waltid-libraries/credentials/waltid-digital-credentials/src/commonMain/kotlin/id/walt/credentials/CredentialParser.kt
+++ b/waltid-libraries/credentials/waltid-digital-credentials/src/commonMain/kotlin/id/walt/credentials/CredentialParser.kt
@@ -17,6 +17,7 @@ import id.walt.credentials.signatures.SdJwtCredentialSignature
 import id.walt.credentials.signatures.sdjwt.SdJwtSelectiveDisclosure
 import id.walt.credentials.utils.JwtUtils
 import id.walt.credentials.utils.JwtUtils.isJwt
+import id.walt.credentials.utils.SdJwtUtils.dropDollarPrefix
 import id.walt.credentials.utils.SdJwtUtils.getSdArrays
 import id.walt.credentials.utils.SdJwtUtils.parseDisclosureString
 import id.walt.crypto.keys.DirectSerializedKey
@@ -181,6 +182,8 @@ object CredentialParser {
         signature: String,
     ): Pair<CredentialDetectionResult, DigitalCredential> {
         val containedDisclosables = payload.getSdArrays()
+        val containedDisclosablesSaveable = containedDisclosables.dropDollarPrefix()
+
         val containsDisclosures = containedDisclosables.count() >= 1
 
         fun detectedSdjwtSigned(
@@ -242,7 +245,7 @@ object CredentialParser {
                 detectedSdjwtSigned(CredentialPrimaryDataType.SDJWTVC, SDJWTVCSubType.sdjwtvcdm) to
                         SdJwtCredential(
                             dmtype = SDJWTVCSubType.sdjwtvcdm,
-                            disclosables = containedDisclosables,
+                            disclosables = containedDisclosablesSaveable,
                             disclosures = availableDisclosures,
                             signature = SdJwtCredentialSignature(plainSignature, header, availableDisclosures),
                             signed = signedCredentialWithoutDisclosures,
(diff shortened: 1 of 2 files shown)


waltid-libraries/waltid-did/src/commonMain/kotlin/id/walt/did/dids/resolver/local/DidWebResolver.kt
```diff
@@ -127,7 +126,16 @@ class DidWebResolver(private val client: HttpClient) : LocalResolverMethod("web"
     }
 
     companion object {
-        const val URL_PROTOCOL = "https"
+
+        var urlProtocol = "https"
```
mikeplotean: since `urlProtocol` is a static public variable, I could update it from anywhere (just by class name), overwriting the effects of `enableHttps`
it probably needs to be private, or perhaps reconsider the logic, e.g. use a computed property:
```kotlin
var httpsEnabled: Boolean = true

val urlProtocol: String
    get() = if (httpsEnabled) "https" else "http"
```

waltid-services/waltid-integration-tests/src/main/kotlin/id/walt/test/integration/tests/DidsWalletIntegrationTest.kt
```diff
@@ -66,16 +68,20 @@ class DidsWalletIntegrationTest : AbstractIntegrationTest() {
...
+        DidWebResolver.enableHttps(false)
+        val did = DidService.resolve(createdDidString).getOrNull()
+        assertNotNull(did)
         assertEquals(createdDidString, loadedDid["id"]?.jsonPrimitive?.content)
         val verificationMethod = loadedDid["verificationMethod"]?.jsonArray?.get(0)?.jsonObject
         assertNotNull(verificationMethod?.jsonObject).also {
             assertNotNull(it["publicKeyJwk"]?.jsonObject)
             assertEquals("JsonWebKey2020", it["type"]?.jsonPrimitive?.content)
-            assertEquals("did:web:domain:path", it["controller"]?.jsonPrimitive?.content)
+            assertEquals("did:web:localhost%3A22323:wallet-api:registry:mydid", it["controller"]?.jsonPrimitive?.content)
```
mikeplotean: wouldn't it be better to extract the `"${environment.e2e.host}:${environment.e2e.port}"` and `"/wallet-api/registry/mydid"` into separate variables and use them in the assertion as well with a different port separator for domain, or for example:
```kotlin
private enum class ColonEncoding(val value: String) {
    PLAIN(":"),
    URL("%3A"),
    PATH("/")
}

private fun createDomainString(encoding: ColonEncoding = ColonEncoding.PLAIN): String =
    "${environment.e2e.host}${encoding.value}${environment.e2e.port}"

private fun createPathString(encoding: ColonEncoding = ColonEncoding.PLAIN): String =
    ":wallet-api:registry:mydid".replace(":", encoding.value)
```
in this case, the 'arrange'  ('given') part of the test would be:
```kotlin
options = mapOf("domain" to createDomainString(), "path" to createPathString(ColonEncoding.PATH))
```
and the assertion:
```kotlin
expected = "did:web:${createDomainString(ColonEncoding.URL)}${createPathString()}",
```

waltid-services/waltid-integration-tests/src/main/kotlin/id/walt/test/integration/tests/DidsWalletIntegrationTest.kt
```diff
@@ -2,6 +2,8 @@
 
 package id.walt.test.integration.tests
 
+import id.walt.did.dids.DidService
+import id.walt.did.dids.resolver.local.DidWebResolver
```
mikeplotean, line 6: perhaps, the [DidWebResolverTest](https://github.com/walt-id/waltid-identity/blob/main/waltid-libraries/waltid-did/src/jvmTest/kotlin/resolvers/DidWebResolverTest.kt) could also be updated with cases for http and fixed for the https cases

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/Main.kt
```diff
@@ -54,9 +54,12 @@ suspend fun main(args: Array<String>) {
 }
 
 fun webWalletSetup() {
-    log.info { "Setting up..." }
-    val runtimeConfig = ConfigManager.getConfig<RuntimeConfig>()
-    DidWebResolver.enableHttps(runtimeConfig.enableDidWebResolverHttps)
+    log.info { "Setting up wallet ..." }
+    try {
```
mikeplotean: this could also be done using runCatching, e.g.:
```kotlin
runCatching { DidWebResolver.enableHttps(ConfigManager.getConfig<RuntimeConfig>().enableDidWebResolverHttps) }.onFailure {
    log.error(it) { "Could not load `enableDidWebResolverHttps` from runtime config. " + it.message }
}
```

mikeplotean (dismissed):

mikeplotean (approved):

PR #1325, opened 2025-11-06
Updated gradle-setup action to accept location as input

dinkar-jain:


.github/workflows/api-docker.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         with:
           token: ${{ secrets.GH_ACCESS_TOKEN }}
       - name: Configure gradle
-        uses: walt-id/waltid-identity/.github/actions/gradle-setup-action@a4a92e12ee9b7396ee9a22083ff21b3277c627d6
+        uses: walt-id/waltid-identity/.github/actions/gradle-setup-action@0d1d3a9838e7266c3ac353d3bb627c5161bb43eb
```
mikeplotean, line 38: doesn't it require an update in the build workflow as well (for the api-docker and gradle references)?

dinkar-jain: Done, thanks for reminding

PR #1335, opened 2025-11-10
Dev mode

philpotisk:
## Description

- Added a typed dev-mode feature with DevModeConfig and ConfigManager.getDevModeConfig() so every service can centrally toggle HTTP/HTTPS DID Web resolution
- docker-compose and integration configs now include _features.conf and dev-mode.conf 
- Issuer, verifier (v1 & v2), and wallet mains all initialize ServiceCommons.initDevMode()
- Deprecated WaltidServices in favor of using DidService directly 
- Simplified wallet manifest endpoints by deleting the Entra mock extractor

This video shows the feature-toogle "dev-mode" in action:
[Watch the video](https://www.youtube.com/watch?v=2T6wT6SH7Kk)

## Type of Change

## Checklist

- [x] unit + e2e test coverage -> implicitly tested by: https://github.com/walt-id/waltid-identity/pull/1316
- [ ] documentation updated accordingly -> feature-toggle documented in config. The public docs might need an update.

## Breaking

- \-

### DIFF
--- a/waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/Main.kt
+++ b/waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/Main.kt
@@ -1,18 +1,13 @@
 package id.walt.issuer
 
-import id.walt.commons.ServiceCommons
 import id.walt.commons.ServiceConfiguration
 import id.walt.commons.ServiceInitialization
 import id.walt.commons.ServiceMain
-import id.walt.commons.config.ConfigManager
-import id.walt.commons.config.list.DevModeConfig
 import id.walt.commons.featureflag.CommonsFeatureCatalog
-import id.walt.commons.featureflag.FeatureManager
 import id.walt.commons.featureflag.FeatureManager.whenFeature
 import id.walt.commons.web.WebService
 import id.walt.crypto.keys.aws.WaltCryptoAws
 import id.walt.did.dids.DidService
-import id.walt.did.dids.resolver.local.DidWebResolver
 import id.walt.issuer.entra.entraIssuance
 import id.walt.issuer.issuance.OidcApi.oidcApi
 import id.walt.issuer.issuance.issuerApi
--- a/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/Main.kt
+++ b/waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/Main.kt
@@ -1,12 +1,8 @@
 package id.walt.verifier
 
-import id.walt.commons.ServiceCommons
 import id.walt.commons.ServiceConfiguration
 import id.walt.commons.ServiceInitialization
 import id.walt.commons.ServiceMain
-import id.walt.commons.config.ConfigManager
-import id.walt.commons.config.list.DevModeConfig
-import id.walt.commons.featureflag.FeatureManager
 import id.walt.commons.featureflag.FeatureManager.whenFeature
 import id.walt.commons.web.WebService
 import id.walt.did.dids.DidService
--- a/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/Main.kt
+++ b/waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/Main.kt
@@ -1,6 +1,5 @@
 package id.walt.webwallet
 
-import id.walt.commons.ServiceCommons
 import id.walt.commons.ServiceConfiguration
 import id.walt.commons.ServiceInitialization
 import id.walt.commons.ServiceMain


docker-compose/issuer-api/config/_features.conf
waltkb, line 1: Should this be the default?

docker-compose/verifier-api/config/_features.conf
waltkb, line 1: Should this be the default?

waltkb (dismissed):

docker-compose/verifier-api/config/dev-mode.conf
```diff
@@ -0,0 +1 @@
+enableDidWebResolverHttps=false
```
mikeplotean, line 1: maybe this could be renamed to smth. more generic, since it's not only about did-web resolver, e.g. `enableHttps` or `useHttps` or just `https` or `secure`?

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/Main.kt
```diff
@@ -28,6 +29,7 @@ suspend fun main(args: Array<String>) {
             init = {
                 WaltidServices.minimalInit()
```
mikeplotean: deprecated feature, use `DidService` instead

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/Main.kt
```diff
@@ -28,6 +29,7 @@ suspend fun main(args: Array<String>) {
...
                 WaltCryptoAws.init()
+                ServiceCommons.initDevMode()
```
mikeplotean: imho, this can be confusing
my thoughts would be: 
- why do I have to call this in production
- what if I don't call it at all

waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/config/ConfigManager.kt
```diff
@@ -108,6 +110,13 @@ object ConfigManager {
                 }
         }
 
+    fun getDevModeConfig(): DevModeConfig {
```
mikeplotean: imho, this feels a bit 'hacky', for several reasons
reason#1
I think the decision whether to use the config or not belongs to the client, i.e. issuer-api, wallet-api, etc. (see my comment in issuer-api Main.kt)

reason#2
calling feature-manager in config-manager kind of breaks the 'encapsulation'
besides, my understanding of this hierarchy is that config-manager is a dependency of feature-manager, if this is the case, then it could potentially lead to circular dependency

reason#3 (some sort of clarification for #2)
having a separate way to retrieve the dev-mode config makes it a special config, but this is not reflected in any way within the type, i.e. `DevModeConfig` is nothing different from `WebConfig` or other config I can create within the clients, I could do the same with any config, e.g. have a separate method for `AuthConfig` from wallet-api within service-commons

waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/Main.kt
```diff
@@ -22,6 +23,7 @@ suspend fun main(args: Array<String>) {
                     registerResolver(LocalResolver())
                     updateResolversForMethods()
                 }
+                ServiceCommons.initDevMode()
```
mikeplotean: same concern as with issuer-api

waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/openid4vp/verifier/Main.kt
```diff
@@ -27,6 +28,7 @@ suspend fun main(args: Array<String>) {
                     registerResolver(LocalResolver())
                     updateResolversForMethods()
                 }
+                ServiceCommons.initDevMode()
```
mikeplotean: same concern as with verifier-api

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/Main.kt
```diff
@@ -43,10 +41,11 @@ suspend fun main(args: Array<String>) {
             ),
             init = {
                 webWalletSetup()
-                WaltidServices.minimalInit()
+                DidService.minimalInit()
                 WaltCryptoOci.init()
                 WaltCryptoAws.init()
                 Db.start()
+                ServiceCommons.initDevMode()
```
mikeplotean: same concern as with verifier-api

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/dids/DidsService.kt
```diff
@@ -39,15 +39,15 @@ object DidsService {
     fun add(wallet: Uuid, did: String, document: String, keyId: String, alias: String? = null) = transaction {
         val now = Clock.System.now()
         val didExists = WalletDids.selectAll()
-            .where { (WalletDids.wallet eq wallet.toJavaUuid()) and (WalletDids.did eq did.replace("%3A", ":")) }
+            .where { (WalletDids.wallet eq wallet.toJavaUuid()) and (WalletDids.did eq did) }
```
mikeplotean, line 42: just to make sure, this has been added as a [fix](https://github.com/walt-id/waltid-identity/commit/8899e3c7f3580186d35dcb553aec9ec1f45c4d41) a while ago

docker-compose/verifier-api/config/dev-mode.conf
philpotisk: this is only for did:web resolvers. it is not a general setting for all http-clients.

waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/dids/DidsService.kt
philpotisk: I don't fully understand why this was introduced. But according the spec "%3A" https://w3c-ccg.github.io/did-method-web/#example-creating-the-did-with-optional-path-and-port, this is the way ports are encoded.

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/Main.kt
philpotisk: nice catch

waltid-services/waltid-verifier-api/src/main/kotlin/id/walt/verifier/Main.kt
mikeplotean: init-dev-mode is executed now at the service-commons level, on `initFeatures`, being called right after `ServiceInitialization` is invoked, only when the feature is enabled, failing silently with a waning in cases of I/O issues with the config file

mikeplotean (approved):
the 10 issues sonar reports are all 'unused import' related, looks like an easy fix

PR #1345, opened 2025-11-19
feat: add mdocs-credential custom serializer to include the mso

mikeplotean:
## Description

This PR introduces a custom serializer for `MdocsCredential` to ensure the Mobile Security Object (mso) is correctly included during serialization. This change addresses issues related to mdoc status mapping (linked to WAL-309).

- implements a custom `KSerializer` for `MdocsCredential` that explicitly serializes the `mso` field
- refactors:
  - extracts the credential serial name and format into reusable constants
  - adds an `mso` decoding hook to `MdocsCredential` to facilitate better testing
- adds a new test case to verify the correct serialization of `MdocsCredential`
- adds [`mokkery`](https://mokkery.dev) as a dependency to `waltid-digital-credentials` to support the new testing requirements
- offtopic: configures the jvm-toolchain for libraries: `openid-verifier`, `openid-verifier-api`, `waltid-mdoc-credentials2`

## Type of Change

## Checklist

## Breaking

- \-

### DIFF
--- a/waltid-libraries/credentials/waltid-mdoc-credentials2/build.gradle.kts
+++ b/waltid-libraries/credentials/waltid-mdoc-credentials2/build.gradle.kts
@@ -22,6 +22,7 @@ val enableAndroidBuild = getSetting("enableAndroidBuild")
 val enableIosBuild = getSetting("enableIosBuild")
 
 kotlin {
+    jvmToolchain(21)
     jvm()
 
     js(IR) {
--- a/waltid-libraries/protocols/waltid-openid4vp-verifier-openapi/build.gradle.kts
+++ b/waltid-libraries/protocols/waltid-openid4vp-verifier-openapi/build.gradle.kts
@@ -39,6 +39,10 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.9.0")
 }
 
+kotlin {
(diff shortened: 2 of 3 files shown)


waltkb (changes requested):
tbh I don't like introducing a separate Serializer + wrapper with custom serialization + deserialization. Is this just for accessing the MSO? This shouldn't really be necessary, as almost everything of the MSO is already propagated upwards generically. I assume you want to access a potential status list entity - in this case I would add such as generic attribute to the interface above (as the other credentials SD-JWT + W3C could also contain such) instead of exposing anything hyper-specific, after all the credential abstraction layer should allow for generic access wherever possible, otherwise we wouldn't need it and could just pass the specific credential classes around

mikeplotean:
> tbh I don't like introducing a separate Serializer + wrapper with custom serialization + deserialization. Is this just for accessing the MSO? This shouldn't really be necessary, as almost everything of the MSO is already propagated upwards generically. I assume you want to access a potential status list entity - in this case I would add such as generic attribute to the interface above (as the other credentials SD-JWT + W3C could also contain such) instead of exposing anything hyper-specific, after all the credential abstraction layer should allow for generic access wherever possible, otherwise we wouldn't need it and could just pass the specific credential classes around

not sure about what type the status property would have in this case

PR #1365, opened 2025-11-30
fix: conversion function in crypto lib

cpatsonakis:
## Description

Fixes WAL-411.


waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/keys/jwk/JWKKeyCreator.kt
```diff
@@ -65,17 +64,17 @@ ${Base64.Pem.encode(data)}
         val der = if (str.lines()[0].contains("BEGIN CERTIFICATE")) {
             str.replace("BEGIN CERTIFICATE", "")
                 .replace("END CERTIFICATE", "")
+                .replace("\r", "")
```
mikeplotean, line 67: ❤️

Issue #1355, opened 2025-11-25
JWK export is not available for remote keys.


malindaashan:
We are trying to integrate the [Walit.id](http://Walit.id) issuer and have implemented AWS KMS for key management. However, when we attempt to retrieve the offer URL and add an mDoc to the wallet, we encounter the error:

**"JWK export is not available for remote keys."**

Does this occur because the AWS KMS private key never leaves the HSM, and therefore cannot be exported as a JWK?

What is the recommended approach to integrate AWS KMS as our document signer in this scenario? Or do we missing anything in our /issue request

[waltid-issuer.log](https://uploads.linear.app/9cf52057-4834-47fb-8bf8-517f3bf17fc6/1284a83d-316a-4b4d-b034-120dfaba429c/6c549941-97f7-44b3-b49a-182946ec8603?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzljZjUyMDU3LTQ4MzQtNDdmYi04YmY4LTUxN2YzYmYxN2ZjNi8xMjg0YTgzZC0zMTZhLTRiNGQtYjAzNC0xMjBkZmFiYTQyOWMvNmM1NDk5NDEtOTdmNy00NGIzLWI0OWEtMTgyOTQ2ZWM4NjAzIiwiaWF0IjoxNzY0MzIzOTczLCJleHAiOjE3OTU4OTQ1MzN9.7f-7l4oknszw21rVrAfGad-8y8pItrOlKDZA4q48_7A)

JakeFernandes98:
Thanks for raising this issue! We have started tracking this on our internal task board and an engineer will pick this up soon. We have taken this on lower priority, but if you have a project that relies on this please reach out to discuss options for support contracts for dedicated support via https://walt.id/contact

JakeFernandes98:
While we work on reproducing the issue, please check you are issuing mDocs according to [our docs](https://docs.walt.id/community-stack/issuer/api/credential-issuance/mdl-oid4vc), we only mention this as the flow for issuing mdocs differs a bit from other credential standards :)

philpotisk:
Hello @malindaashan , thanks for bringing this topic up! 

Your assumption is correct: the JWK key can not be exported, since the key is managed externally (at AWS in this case).

[...]

PR #1402, opened 2025-12-10
migrated issuer + wallet to use mdocs2

JakeFernandes98:
Updates issuer and related wallet component to use the mdocs2 library to support AWS keys for mdoc issuance.

This is not a high quality migration, but is sufficient given both of the services are soon to be deprecated. Hopefully the PR and accompanying review are useful for implementing the mdocs-2 and cose functionality in the new issuer.

Solves #1355

Fixes WAL-367

### DIFF
--- a/waltid-applications/waltid-credentials/k8s/deployment-prod.yaml
+++ b/waltid-applications/waltid-credentials/k8s/deployment-prod.yaml
@@ -49,13 +49,13 @@ kind: Ingress
 metadata:
     name: credentials-prod-ingress
     annotations:
-        kubernetes.io/ingress.class: "nginx"
         cert-manager.io/cluster-issuer: letsencrypt-prod
         ingress.kubernetes.io/configuration-snippet: |
             if ($host ~ ^(.+)\.waltid\.org$) {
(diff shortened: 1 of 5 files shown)


waltid-services/waltid-wallet-api/src/main/kotlin/id/walt/webwallet/service/oidc4vc/TestCredentialWallet.kt
```diff
@@ -537,9 +537,19 @@ class TestCredentialWallet(
...
         } ?: keyId
-        KeysService.get(kid)?.let {
-            KeyManager.resolveSerializedKey(it.document)
+        
+        val walletKey = KeysService.get(kid)
+        
+        if (walletKey == null) {
+            // Try to resolve directly from DID if it's a DID URL
+            if (DidUtils.isDidUrl(keyId)) {
+                return@runCatching DidService.resolveToKey(keyId).getOrNull()
```
waltkb: Won't this be a public key?

waltkb (approved):

PR #1438, opened 2026-01-08
Updated docker instructions + fixed alignment-warning

philpotisk:
## Description

Updated service READMEs to document the recommended Gradle Ktor/Jib Docker build flow for development (local daemon) and production (multi‑arch push) across verifier‑api2, verifier‑api, issuer‑api, wallet‑api, and enterprise‑api. Added brief notes at the top of each service Dockerfile to clarify they’re for manual/custom Docker workflows and that Gradle Ktor/Jib tasks are preferred.

### DIFF
--- a/waltid-applications/waltid-credentials/k8s/deployment-prod.yaml
+++ b/waltid-applications/waltid-credentials/k8s/deployment-prod.yaml
@@ -49,13 +49,13 @@ kind: Ingress
 metadata:
     name: credentials-prod-ingress
     annotations:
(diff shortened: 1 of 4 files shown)


waltkb:
I am pretty sure that GitHub Markdown is sanitized aggressively enough to break all the changed style tags (resulting in left-aligned text). While the change is understandable (I assume automated inspection by IDE) as the align attribute is technically deprecated in HTML5, it is whitelisted by GitHub. As such, the legacy align attribute has to remain instead of CSS.

PR #1458, opened 2026-01-20
Feature/wal 502

cpatsonakis:
## Summary by CodeRabbit

* **New Features**
  * Added synchronous blocking variants for certificate build/parse/validate on supported platforms.

* **Improvements**
  * Switched onboarding flows to builder-based certificate generation.
  * Expanded supported EC signing keys to secp256r1, secp384r1 and secp521r1.
  * Updated README with linked ISO references and configurable validation guidance; defaults and validity limits clarified (20-year cap).

* **Tests**
  * Reworked onboarding tests and added blocking-vs-suspend equivalence tests; removed legacy ISO mDL onboarding E2E tests.

* **Breaking Changes**
  * Onboarding now expects explicit notBefore/notAfter in certificate request payloads and altered request/response shapes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Resolves WAL-502, WAL-459.

### DIFF
--- a/docker-compose/.env
+++ b/docker-compose/.env
@@ -1,4 +1,4 @@
-VERSION_TAG=0.16.2
+VERSION_TAG=stable
 # Images for ARM (eg. Apple Silicon) are only available with the `latest` tag
 
 VERSION_OPA=1.4.2
--- a/waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/keys/KeySerialization.kt
+++ b/waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/keys/KeySerialization.kt
@@ -7,6 +7,7 @@ import id.walt.crypto.keys.oci.OCIKeyRestApi
 import id.walt.crypto.keys.tse.TSEKey
 import id.walt.crypto.utils.JsonUtils.toJsonElement
 import kotlinx.serialization.InternalSerializationApi
+import kotlinx.serialization.KSerializer
 import kotlinx.serialization.json.*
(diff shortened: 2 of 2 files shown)


cpatsonakis:
@coderabbitai review

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/services/onboarding/models/IssuerAlternativeNameConfiguration.kt
```diff
@@ -6,11 +6,4 @@ import kotlinx.serialization.Serializable
...
-) {
-
-    init {
-        require(!email.isNullOrBlank() || !uri.isNullOrBlank()) {
-            "At least one of 'email' or 'uri' must be specified"
-        }
-    }
-}
\ No newline at end of file
+)
```
chsavvaidis, line 9: Could this change lead to runtime errors or invalid certificates when both email and uri are blank, since the X.509 builder may try to create an empty IssuerAlternativeName/  GeneralNames?

chsavvaidis (approved):

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/services/onboarding/models/IssuerAlternativeNameConfiguration.kt
cpatsonakis: Yes, both the builder and the validator will throw accordingly. Also there are config options in the validation system that specifically treat this case. The spec says that one, or the other, or both must be specified which requires that both are set to nullable.

Issue #1497, opened 2026-02-11
OpenID4VP presentation fails in wallet: `JsonArray is not a JsonPrimitive` when presenting jwt_vc with vc.type as array


dmorperd:
When trying to present a `jwt_vc` credential using OpenID4VP, the wallet fails with:
`Element class kotlinx.serialization.json.JsonArray is not a JsonPrimitive`

It happens at the moment the wallet tries to present the VC after opening the deeplink.

### Expected behavior
Wallet should accept the JWT VC and proceed with VP creation / submission.

### Actual behavior
Wallet crashes/fails with Kotlin serialization error:
`Element class kotlinx.serialization.json.JsonArray is not a JsonPrimitive`

### Deeplink used
openid4vp://authorize?response_type=vp_token&client_id=http%3A%2F%2Flocalhost%3A7003%2Fopenid4vc%2Fverify&response_mode=direct_post&state=E55t9sbQQxXQ&presentation_definition_uri=http%3A%2F%2Flocalhost%3A7003%2Fopenid4vc%2Fpd%2FE55t9sbQQxXQ&client_id_scheme=redirect_uri&client_metadata=%7B%22authorization_encrypted_response_alg%22%3A%22ECDH-ES%22%2C%22authorization_encrypted_response_enc%22%3A%22A256GCM%22%7D&nonce=eb7af869-bd3c-4e00-a838-fce66d7a707f&response_uri=http%3A%2F%2Flocalhost%3A7003%2Fopenid4vc%2Fverify%2FE55t9sbQQxXQ
### Version used
https://github.com/walt-id/waltid-identity

### Reproduction steps
1. Call verifier endpoint to generate deeplink:

```bash
curl -X POST 'http://217.154.108.26:7003/openid4vc/verify' \
  -H 'accept: text/plain' \
  -H 'authorizeBaseUrl: openid4vp://authorize' \
  -H 'responseMode: direct_post' \
  -H 'Content-Type: application/json' \
  -d '{
    "vc_policies": ["signature","expired","not-before","revoked-status-list"],
    "request_credentials": [{
      "format": "jwt_vc",
      "input_descriptor": {
        "id": "e3d700aa-0988-4eb6-b9c9-e00f4b27f1d8",
        "constraints": {
          "fields": [{
            "path": ["$.vc.type"],
            "filter": {
              "contains": { "const": "DataspaceParticipantCredential" },
              "type": "array"
            }
          }]
        }
      }
    }]
  }

PR #1513, opened 2026-02-17
extend PD support with integration test

JakeFernandes98:
Resolves WAL-597 and solves https://github.com/walt-id/waltid-identity/issues/1497

The PresentationDefinitionFilterParser didn't handle the "contains" filter structure used when vc.type is an array.

Added support in both core wallet lib and wallet api. Added an integration test in the OSS and ensured all tests pass in OSS and enterprise

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of credential filter parsing to safely handle non-primitive JSON values and alternative filter shapes, preventing runtime errors and improving presentation request validation.
  * Added support for "contains" style filters when matching credential type arrays.

* **Tests**
  * Added test coverage and a helper to validate array-based "contains" matching for credential types, asserting accepted matches.

### DIFF
--- a/waltid-services/waltid-integration-tests/src/test/kotlin/InputDescriptorMatchingTest.kt
+++ b/waltid-services/waltid-integration-tests/src/test/kotlin/InputDescriptorMatchingTest.kt
@@ -76,12 +76,10 @@ class InputDescriptorMatchingTest(
         // Request: $.credentialSubject.degree.type: "UniversityDegree",
         // --> match should return 0, presentation should be rejected
         verifyCredential(getPresentationRequestByDegreeType("UniversityDegree"), wallet, did, newCredential1, false)
-        
-        // Issue a credential with type array containing "DataspaceParticipantCredential"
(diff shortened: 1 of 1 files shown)


philpotisk (approved):

arnauinfante:
good afternoon, Not sure if you will read this comment, but I am a member from the team of the user which opened the issue. After debugging I found out I was getting the json primitive error because of credential format. I was expecting a jwt_vc_json but it was a credential with a json-ld configuration. When service requests for valid credentials in jwt_vc_json format, it doesn't find any, so it gives an empty array as a response. After that it tries to process it as a json primitve, which leads to the commented error. Modifying the credential config to use a json instead of json-ld fixed the error. Not sure if this pull request handle this scenario but the error showed didn't match with the real problem. If there's no credential matching expected format in presentation defintion, it should give a 400. Also, wallet web interface is not checking format, only credential type filter, so it lists credentials only by type and I could present it even not matching the format.

PR #1568, opened 2026-02-26
set property for CRLDP

JakeFernandes98:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate revocation checking by enabling CRL Distribution Point (CRLDP) support at startup, increasing verification reliability when revocation validation is enabled.
  * Added a runtime startup option to allow environments to enable CRLDP-based revocation checks as needed.

### DIFF
--- a/waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/Main.kt
+++ b/waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/Main.kt
@@ -42,7 +42,7 @@ fun Application.configurePlugins() {
     configureHTTP()
     configureMonitoring()
 }
-K
+
 fun Application.verifierModule(withPlugins: Boolean = true) {
 
     // Enable CRL Distribution Point support for certificate revocation checking


JakeFernandes98:
Put it in the application module, not in the Main function

waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/Main.kt
```diff
@@ -47,8 +42,14 @@ fun Application.configurePlugins() {
...
     configureMonitoring()
 }
-
+K
 fun Application.verifierModule(withPlugins: Boolean = true) {
+
+    // Enable CRL Distribution Point support for certificate revocation checking
+    // This allows the PKIX validator to fetch CRLs from the CRLDistributionPoint
+    // extension in X.509 certificates when enableRevocation is true in vICAL policies
+    System.setProperty("com.sun.security.enableCRLDP", "true")
```
philpotisk, line 51: @JakeFernandes98 @cpatsonakis does this change apply that the documentation is incorrect? https://github.com/walt-id/waltid-identity/blob/main/waltid-libraries/crypto/waltid-x509/README.md#platform-notes , as the the suggestion is to set this via the environment. We should either not hardcode this setting or update the docs.

PR #1679, opened 2026-04-29
add support for status passthrough in issuer

### LINKED ISSUE #1676: mdoc revocation status (`status_list` / `identifier_list`) in issuer API
Hi team,

Is there way issuer api supports  mdoc revocation status in the issuance request payload (for example `status.identifier_list` or `status.status_list`) when calling `/openid4vc/mdoc/issue`?

Thanks,
Ashan

JakeFernandes98:
Adds support for passing status to w3c, ietf sdjwt and mdocs (1) credentials in the issuer

Resolves #1676

### DIFF
--- a/waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
+++ b/waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
@@ -5,74 +5,169 @@ import id.walt.mdoc.mso.IdentifierListInfo
 import id.walt.mdoc.mso.Status
 import id.walt.mdoc.mso.StatusListInfo
 import io.ktor.server.plugins.BadRequestException
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonNull
 import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.intOrNull
 import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.longOrNull
+import java.net.URI
+import java.net.URISyntaxException
 import java.util.Base64
 
 internal fun JsonObject.toMdocIssuerStatusOrNull(): Status? {
     if (isEmpty()) return null
     val statusListEl = this["status_list"]
     val identifierListEl = this["identifier_list"]
-    when {
-        statusListEl != null && identifierListEl != null ->
-            throw BadRequestException("mdoc status_list and identifier_list are mutually exclusive")
+    if (statusListEl.isPresentJson() && identifierListEl.isPresentJson()) {
+        throw BadRequestException(
+            "Only one of status_list or identifier_list may be set. ISO/IEC 18013-5 clause 9.1.2.6 " +
+                "(MSO revocation) requires the identifier list mechanism to use identifier_list and " +
+                "the status list mechanism to use status_list — not both.",
+        )
+    }
+    return when {
+        statusListEl.isPresentJson() ->
+            MdocIssuerStatusJsonParser.parseStatusList(statusListEl.requireJsonObject("status_list"))
 
-        statusListEl != null -> {
-            val sl = statusListEl.jsonObject
-            val idxEl = sl["idx"] ?: throw BadRequestException("status_list.idx is required")
-            val uriEl = sl["uri"] ?: throw BadRequestException("status_list.uri is required")
-            val idx = idxEl.jsonPrimitive.longOrNull?.toUInt()
-                ?: idxEl.jsonPrimitive.intOrNull?.toUInt()
-                ?: throw BadRequestException("status_list.idx must be a non-negative integer")
-            val uri = uriEl.jsonPrimitive.content
-            val certificate = sl["certificate"]?.let { cert ->
-                decodeBinaryJsonPrimitive(cert.jsonPrimitive)
-            }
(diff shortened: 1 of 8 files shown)


waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.serialization.json.longOrNull
+import java.util.Base64
+
+internal fun JsonObject.toMdocIssuerStatusOrNull(): Status? {
+    if (isEmpty()) return null
+    val statusListEl = this["status_list"]
+    val identifierListEl = this["identifier_list"]
+    when {
+        statusListEl != null && identifierListEl != null ->
+            throw BadRequestException("mdoc status_list and identifier_list are mutually exclusive")
```
chsavvaidis: "mdoc status_list and identifier_list are mutually exclusive". Could you point the reference?

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        statusListEl != null -> {
+            val sl = statusListEl.jsonObject
+            val idxEl = sl["idx"] ?: throw BadRequestException("status_list.idx is required")
+            val uriEl = sl["uri"] ?: throw BadRequestException("status_list.uri is required")
+            val idx = idxEl.jsonPrimitive.longOrNull?.toUInt()
+                ?: idxEl.jsonPrimitive.intOrNull?.toUInt()
```
chsavvaidis: refactor this. it seems unsafe

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
```diff
@@ -0,0 +1,78 @@
...
+                ?: throw BadRequestException("status_list.idx must be a non-negative integer")
```
chsavvaidis: Please validate that status_list.uri is a non-blank URI string. Currently an empty or whitespace-only value would be accepted

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
```diff
@@ -0,0 +1,78 @@
...
+            val uri = uriEl.jsonPrimitive.content
+            val certificate = sl["certificate"]?.let { cert ->
+                decodeBinaryJsonPrimitive(cert.jsonPrimitive)
+            }
+            return Status(statusList = StatusListInfo(index = idx, uri = uri, certificate = certificate))
+        }
+
+        identifierListEl != null -> {
+            val il = identifierListEl.jsonObject
+            val uriEl = il["uri"] ?: throw BadRequestException("identifier_list.uri is required")
```
chsavvaidis: same as status_list.uri

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
```diff
@@ -0,0 +1,78 @@
...
+    in '0'..'9' -> c - '0'
+    in 'a'..'f' -> c - 'a' + 10
+    in 'A'..'F' -> c - 'A' + 10
+    else -> throw BadRequestException("Invalid hex character in binary field: $c")
+}
+
+private fun decodeBase64Lenient(s: String): ByteArray =
+    runCatching { Base64.getUrlDecoder().decode(s) }
+        .recoverCatching { Base64.getDecoder().decode(s) }
+        .getOrElse { throw BadRequestException("Value must be hex or valid Base64 (standard or URL-safe)") }
```
chsavvaidis: These private decoding functions should be refactored into a stricter, centralized parser. The current implementation mixes hex detection, prefix handling, Base64 fallback, and validation across several helpers, which makes edge cases harder to reason about.

Consider preserving whether a 0x/0X prefix was provided and treating prefixed values strictly as hex instead of falling back to Base64. Also consider rejecting blank input, normalizing Base64 padding if unpadded Base64URL is supported, and passing the field path into the decoder so error messages clearly identify the invalid field, such as identifier_list.id or status_list.certificate.

chsavvaidis (changes requested):

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
JakeFernandes98: from page 62 of the spec:
"Two mechanisms can be used to encode the revocation information: as an identifier list or as a status list."

However, we can loosen the requirement if you feel this should not be interpreted as mutually exclusive. Personally I don't see why both would be used in conjunction, but please let me know your POV

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/IssuanceCredentialStatus.kt
JakeFernandes98: Improved. It may make sense to centralise this, but I would prefer to it to be done by hand. This is just for the issuer1 (which will be deprecated). We can consider an alternative approach for the issuer2

chsavvaidis (approved):

PR #1682, opened 2026-04-30
resolve hex and binary mixed usage in cwt credentials

JakeFernandes98:
ES: https://github.com/walt-id/waltid-identity-enterprise/pull/432


philpotisk (approved):
nice one

PR #1699, opened 2026-05-07
fix: Pinned vc-repo docker base image

dinkar-jain:


dinkar-jain:
PR #1698 incorporates this fix already.

PR #1728, opened 2026-05-15
remove fractional seconds from mdoc implementation

JakeFernandes98:


waltkb (approved):
Also fixed in conformance suite branch

PR #1779, opened 2026-06-15
improve session invalidation logic

JakeFernandes98:
ES: https://github.com/walt-id/waltid-identity-enterprise/pull/493
Docs: https://github.com/walt-id/waltid-docs/pull/297


waltid-libraries/auth/waltid-ktor-authnz/src/main/kotlin/id/walt/ktorauthnz/auth/DefaultKtorAuthnzAuthentication.kt
```diff
@@ -43,9 +43,19 @@ class DefaultKtorAuthnzAuthentication internal constructor(
...
 
-        val principal = if (KtorAuthnzManager.tokenHandler.validateToken(effectiveToken))
+        if (!KtorAuthnzManager.tokenHandler.validateToken(effectiveToken)) {
+            log.debug { "Invalid token (signature/expiration) for request" }
+            fail(context, AuthenticationFailedCause.InvalidCredentials)
+            return
+        }
+
+        val principal = try {
+            KtorAuthnzManager.tokenHandler.resolveTokenToSession(effectiveToken)
```
waltkb, line 53: Mandatory session lookup breaks Enterprise JWTs across replicas and restarts when using default session store, this means:
- a JWT issued by replica 1 is rejected by replica 2
- every user JWT becomes invalid after an Enterprise restart
- externally supplied JWTs no longer work
- sharing the signing key between replicas is no longer sufficient

waltkb (changes requested):

PR #1812, opened 2026-06-23
Ci/issuer2

dinkar-jain:


chsavvaidis:
@dinkar-jain  Can we use the existing issuer2 config directly instead of copying the full config into the k8s ConfigMap?

JakeFernandes98 (approved):

JakeFernandes98:
> @dinkar-jain Can we use the existing issuer2 config directly instead of copying the full config into the k8s ConfigMap?

As discussed with Dinkar, we've added a ticket to the backlog to implement this improvement for all services in the future

PR #1822, opened 2026-06-29
feat: Added portal2

dinkar-jain:


JakeFernandes98:
Please extend by:

- Adding links to the docs sections for the issuer and verifier on each page
- Updating the issuer page so that we have an additional dropdown for the profile id. When this is changed it should override the profile ID being loaded from the swagger. Right now the swagger has a hardcoded profile id. (using /issuer2/profiles)
- Have a collapsable JSON viewer for the profile as well nearby the profileId dropdown  (using /issuer2/profiles/{profileId})
- For now, filter out the verifier swagger examples to not display anything containing "dc_api" in the swagger title as these will not work with your current QR logic. In the future, we will port Phil's logic for properly handling the dc_api examples and re-enable them
- Lets also do the devops work in this PR to get it deployed with our demo systems on release

JakeFernandes98 (approved):

PR #1846, opened 2026-07-01
fix: restore issuer2 credential catalog

szijpeter:
## Summary

Restores the issuer2 credential metadata and profile configs removed in #1822 for both the default service config and the matching docker-compose copy.

## Why

`main` CI is failing in `:waltid-services:waltid-issuer-api2:test` because the default issuer2 config no longer includes expected profiles such as `universityDegree`.


JakeFernandes98:
Resolved by #1847

PR #1855, opened 2026-07-03
Updates TSE support and improves general error messaging

JakeFernandes98:
Adds alias mappings for external kms where the key type name does not exactly match our internal names.
- This in turn adds support for other keys types in TSE and Azure which we previously did not support

Improves error messaging for TSE, Azure and AWS key generation

Docs: https://github.com/walt-id/waltid-docs/pull/309


waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/keys/ExternalKmsErrorHandling.kt
SuperBatata, line 1: I'd move this inside the `id/walt/crypto/exceptions` folder

JakeFernandes98: moved

SuperBatata (approved):

PR #1883, opened 2026-07-09
Feat/portal2

JakeFernandes98:
Makes some improvements to the Portal 2:

- Adds a simplified mode
- Adds customisable options for signing and certificates


dinkar-jain:
**QR Overflow** - Also happening during issuance with tx-code
image

dinkar-jain:
Simple mode doesn't preserve state. If you change the selection from its default, switch to "Verify from "Issue," and switch back, the selection resets. Advanced mode preserves it. Not a priority, just something to note.

Similarly for "Issuance flow" states in "Issue" and "Claims to request" in "Verify" too

dinkar-jain:
**Button State:** The button should probably also change color, and the cursor should switch from pointer to not-allowed to clearly communicate that the action is currently unavailable.

image

JakeFernandes98:
Resolved by:
- Removing raw openid link from the QR code generation, opting instead for a "Copy openid link" button for more predictable UI sizing
- Added state preservation across navigation
- Added invalid button state

dinkar-jain (approved):

PR #1891, opened 2026-07-10
[WAL-965] follow-up: add OpenID4VP verifier rejection path

szijpeter:
## Summary

Stub follow-up for the WAL-965 verifier rejection path split out from WAL-751. This PR is intentionally stacked on #1882 and currently contains only an empty marker commit so the follow-up scope is visible without adding placeholder implementation code.

## Intended scope

- Add protocol-level `rejectPresentation(...)` to the mobile SDK and Swift facade.
- Verify supported OpenID4VP error codes against the spec and existing verifier2/backend behavior before exposing enum cases.
- Add Compose and native iOS `Decline` actions that send the rejection response to the verifier.
- Test verifier session behavior for user-denied and unsupported/error cases.

## Non-goals

- No transaction-data rendering.
- No claim-level selective disclosure.
- No changes to the #1882 credential-option submit flow except the minimal integration points required for rejection.

## First implementation steps

1. Inspect OpenID4VP verifier response handling and current verifier2 session status transitions.
2. Define the smallest Swift-friendly rejection reason model that maps to verified standard error codes.
3. Add KMP/mobile SDK API + bridge parity.
4. Add demo app UX and tests after backend/session behavior is confirmed.


szijpeter:
Closing/replacing this stub because the proper follow-up Linear ticket is WAL-1078, not WAL-965.

PR #1896, opened 2026-07-10
Changes for France Identitie

JakeFernandes98:
- Add iss to signed Authorisation Request

### DIFF
--- a/waltid-libraries/protocols/waltid-openid4vp/src/commonMain/kotlin/id/walt/verifier/openid/models/authorization/AuthorizationRequest.kt
+++ b/waltid-libraries/protocols/waltid-openid4vp/src/commonMain/kotlin/id/walt/verifier/openid/models/authorization/AuthorizationRequest.kt
@@ -155,6 +155,13 @@ data class AuthorizationRequest(
      */
     @SerialName("expected_origins")
     val expectedOrigins: List<String>? = null,
+
+    /**
+     * REQUIRED in signed Authorization Request Objects. Per JAR/OID4VP, this
+     * identifies the Verifier issuing the request object and matches client_id.
+     */
+    @SerialName("iss")
+    val issuer: String? = null,
 ) {
 
     fun toHttpUrl(url: URLBuilder = URLBuilder("openid4vp://authorize")): Url {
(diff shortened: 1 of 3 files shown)


waltid-libraries/protocols/waltid-openid4vp/src/commonMain/kotlin/id/walt/verifier/openid/models/authorization/AuthorizationRequest.kt
```diff
@@ -155,6 +155,13 @@ data class AuthorizationRequest(
      */
     @SerialName("expected_origins")
     val expectedOrigins: List<String>? = null,
+
+    /**
```
waltkb, line 159: This is overstated. I checked the specs, and iss is optional under OpenID4VP and recommended under generic JAR, not required:

- OpenID4VP says client_id is required and iss is redundant
- iss MAY be present in a Request Object
- If present, the Wallet MUST ignore it

Neither OID4VP1.0 nor RFC 9191 require iss. And the wallet correctly ignores iss for authentication and uses client_id

waltid-libraries/protocols/waltid-openid4vp/src/commonMain/kotlin/id/walt/verifier/openid/models/authorization/AuthorizationRequest.kt
```diff
@@ -155,6 +155,13 @@ data class AuthorizationRequest(
...
+     * REQUIRED in signed Authorization Request Objects. Per JAR/OID4VP, this
+     * identifies the Verifier issuing the request object and matches client_id.
+     */
+    @SerialName("iss")
+    val issuer: String? = null,
 ) {
 
     fun toHttpUrl(url: URLBuilder = URLBuilder("openid4vp://authorize")): Url {
```
waltkb, line 172: This serializes every non-null property, and because signed sessions now set issuer before URL creation, fullAuthorizationRequestUrl includes iss=<client_id> outside the signed JWT.

However, RFC 9101 explicitly registers iss as an OAuth authorization-request parameter and permits Request Object parameters to be duplicated in query parameters, and OpenID4VP also allows additional request parameters and requires wallets to ignore unsupported ones, so this probably flies fine

waltkb (approved):
Would update the Kdoc to specify this is purely France Identité specific

waltid-libraries/protocols/waltid-openid4vp/src/commonMain/kotlin/id/walt/verifier/openid/models/authorization/AuthorizationRequest.kt
JakeFernandes98: clarified in the kdoc

PR #1901, opened 2026-07-14
WAL-1145 Feat: x509 Certificate Util Library

hokorn:
## Description

New x.509 certificate util library which provides tools for create/sign CSRs and certificates

## Type of Change

## Checklist

### API change in walid-issuer-api service
* POST /onboard/iso-mdl/document-signers
  - Request Body change: iacaSigner requires root CA pem
### API change in libs  
* id.walt.openid4vci.core.OAuth2Provider (waltid-openid4vci)
  - API uses  id.walt.certificate.x509.X509Certificate instead of id.walt.x509.CertificateDer
* id.walt.policies2.vc.policies.VicalPolicy (waltid-verification-policies2)
  - removed parameter  enableTrustedChainRoot

Enterprise Stack requires https://github.com/walt-id/waltid-identity-enterprise/pull/599 because of the API changes


hokorn:
porting of all the code to new x509 api is still missing, and then old x509 api can be removed

hokorn (commented):
Did some cleanup

philpotisk:
Found and fixed a critical trust-store issue while reviewing this - pushed directly to this branch as [6a83d53ca](https://github.com/walt-id/waltid-identity/commit/6a83d53ca) since it's a real security gap, happy to discuss/revert if you see it differently.

**The bug:** `X509CertificateChainValidator.validate()` always unioned any caller-supplied trust store with the validator's own configured `trustStore`:
```kotlin
val trustStoreToUse = additionalTrust?.let { CompositeTrustStore(listOf(it, trustStore)) } ?: trustStore
```
For `X509CertificateUtil.Default` (base `trustStore` = `JavaDefaultTrustStore`, i.e. the JVM's public CA bundle), any caller using the two-arg `validateCertificateChain()`/`validatePemCertificateChain()` overloads got their own anchors silently unioned with the entire public web PKI. A chain terminating in *any* real public CA would validate as trusted even when the caller only intended to trust their own curated anchors.

I traced this affecting consumer call sites across the codebase - `VicalPolicy`, `X509ChainClientAttestationVerifier`, and `ClientIdCrypto2` (OpenID4VP `x509_hash`/`x509_san_dns`/`verifier_attestation` client-id schemes) - all call `X509CertificateUtil.Default` + their own trust store via the two-arg overload, expecting that to be the sole trust boundary. It also affects the library's own docs/examples: `SignCertificateExample.kt` in waltid-examples and the equivalent snippet in this library's README both demonstrate exactly this pattern as "how to validate a chain against your own root" - so it's the path of least resistance, not just an isolated misuse.

Contrast with `X509CertificateUtilBuilder.setTrust()`, which was already correct (fully replaces the trust store) - `X509CertificateChainValidationTest`'s own companion object already configures its test utils via `setTrust()` specifically "without a system trust store to ensure the same behavior in JS and JVM", so the safe pattern was already known/used internally, just not consistently.

**The fix:** the caller-supplied store now fully replaces the base one for that call rather than extending it. Renamed `additionalTrust` → `trustOverride` to make the semantics unambiguous, documented on both call sites and the validator, and added a regression test (`trustOverrideShouldReplaceBaseTrustStoreNotMergeWithIt`) that I confirmed fails against the old code and passes against the fix. All existing tests still pass. No consumer-side changes needed - `VicalPolicy` etc. get the fix for free since it's entirely in the shared validator.

One thing I did *not* touch: the vulnerable pattern in `waltid-examples/src/main/kotlin/x509/SignCertificateExample.kt` lives in a separate repo, so I left it as-is - worth a follow-up there so the example teaches the corrected behavior (which it now demonstrates correctly by accident, but the `setTrust`-based pattern in `ConfigureTrustStoreExample.kt` is the one worth pointing people to explicitly).

philpotisk:
Second fix, pushed as [c0ee35069](https://github.com/walt-id/waltid-identity/commit/c0ee35069) - the IACA onboarding endpoint blindly trusted the caller-supplied root.

**The bug:** `OnboardingService.onboardDocumentSigner()` parses `request.iacaSigner.iacaPem` and uses it directly as the trust anchor for the newly-issued Document Signer certificate (`documentSignerCertUtil.validateCertificateChain(listOf(documentSingerCert), iacaCert)`), but `iacaCert` itself is never run through `IsoIaCaRootX509CertificateProfile` - unlike `onboardIACA()`, which does validate its self-issued root against that profile. A caller who controls the private key referenced by `iacaSigner.iacaKey` could get a Document Signer certificate issued "under" a root that isn't `CA=true`, lacks `pathLenConstraint=0`, has non-compliant `keyUsage`, or is missing the mandatory issuer-alt-name extension. The key/cert cryptographic consistency check downstream still holds, so it's not a full auth bypass, but it defeats the point of requiring a root CA pem for this endpoint at all.

**The fix:** validate `iacaCert` against `IsoIaCaRootX509CertificateProfile` via the existing `iaCaRootCertUtil` (already configured with that profile for `onboardIACA`) before using it as a trust anchor.

Bundled in the same commit since it's two lines below in the same function: a copy-paste bug in the response mapping, `notBefore = documentSingerCert.data.validity.notAfter` (should read `.validity.notBefore`) - the issued certificate itself was always correct, only the API response's `DocumentSignerCertificateData.notBefore` field was wrong.

Added a regression test (`onboard Document Signer rejects a non-IACA-profile-compliant root certificate`) - a self-signed cert built without the IACA profile helper, given a validity window wide enough that only the missing profile check can reject it (I initially got this test wrong - my first version failed for an unrelated validity-window mismatch rather than the profile check, so I want to flag that I verified it properly: confirmed it fails without the fix for the *right* reason and passes with it).

Next up from the earlier review: the dead-no-op `IsoIaCaRootX509CertificateProfile.validateSubjectDn` (parses the subject DN and discards the result, so the documented mandatory country/commonName checks never actually run).

philpotisk:
Third fix, pushed as [6d848212d](https://github.com/walt-id/waltid-identity/commit/6d848212d) - the dead-no-op subject DN validator.

**The bug:** `IsoIaCaRootX509CertificateProfile.validateSubjectDn()` parsed the subject DN into RDN groups and then discarded the result - the whole function body was dead code (a `map` whose result is never used, plus commented-out lines suggesting an unfinished attempt). It's wired into `validate()` so it always ran on every IACA root, but it never logged a single ERROR/WARNING regardless of input. The docstring right above it documents mandatory `countryName` (ISO 3166-1 alpha-2, uppercase) and mandatory `commonName` - neither was actually enforced. An IACA root with a missing or lowercase country code, or no commonName at all, passed ISO-profile validation with zero findings.

**The fix:** the sibling `IsoDocumentSignerX509CertificateProfile.validateSubjectDn` already implements the equivalent check correctly, so I ported that pattern (group RDNs by lowercased attribute type name, check `c` matches `^[A-Z]{2}$`, check `cn` is present and non-blank) rather than inventing something new.

Ran the full `waltid-x509` test suite and the issuer-api onboarding tests both before and after to make sure nothing relied on this being a no-op - a couple of existing negative tests now get an extra `subjectDn` error alongside their existing `issuerDn` one, which their `any {}` assertions already tolerate, so no changes needed there.

Added two regression tests (missing commonName, missing/invalid countryName) in `IsoIaCaRootX509CertificateProfileTest` and verified both fail against the no-op and pass against the fix.

That closes out the three issues I flagged as most severe from the earlier review. Remaining from that list, roughly in priority order: `X5CKeyResolver` dropping the intra-chain signature check, `DocumentSignerCertificateBuilder.buildDer` skipping a public-key validation present in the deprecated path, and the BouncyCastle provider-init commit's public mutable var + Android race window. Let me know if you'd like me to keep going on those or pause here.

philpotisk:
Updated the companion examples repo per your local mavenLocal publish - [walt-id/waltid-examples#33](https://github.com/walt-id/waltid-examples/pull/33) (same branch name, same author): clarified the trust-store scoping in the two existing x509 examples, and added `IsoMdlOnboardingExample.kt` demonstrating the IACA-root-profile-compliance fix (a non-compliant caller-supplied root now gets rejected with specific violations rather than silently accepted). Details in the [comment there](https://github.com/walt-id/waltid-examples/pull/33#issuecomment-5281816079). Built and ran all three examples against a local `waltid-x509:1.0.0-SNAPSHOT` publish to confirm they work against the fixed library.

philpotisk:
Reviewed and rewrote the [waltid-x509 README](https://github.com/walt-id/waltid-identity/blob/feat/x509-util-library/waltid-libraries/crypto/waltid-x509/README.md) - pushed [1be2f8560](https://github.com/walt-id/waltid-identity/commit/1be2f8560).

It had drifted significantly from the current API:
- Quick start snippets referenced undefined variables (`key`, `issuerKey`, `issuerCert`, `subjectKey`, `trustRoot`, `chain`) - not actually copy-pasteable. Fixed to use consistent, defined variables throughout.
- "Configure X509CertificateUtil" was an empty section header with no content - filled it in with the `setTrust()`/`addValidators()` pattern and documented the trust-scoping fix from earlier ([6a83d53ca](https://github.com/walt-id/waltid-identity/pull/1901#issuecomment-5281306555)) directly in the library's own docs.
- The entire ISO/IEC 18013-5 section documented the deprecated `id.walt.x509.iso` API - every symbol it referenced (`IACACertificateBuilder`, `IACAValidator`, `CertificateDer`, ...) now carries `@Deprecated`. Replaced with the current profile-based API and added the "always validate a caller-supplied root before trusting it" lesson from the onboarding fix ([c0ee35069](https://github.com/walt-id/waltid-identity/pull/1901#issuecomment-5281498316)).
- "Loading trust anchors from a JVM KeyStore" referenced a helper (`loadTrustAnchorsFromKeyStore`) that isn't part of the current API's public surface documented here - replaced with a short correct helper using `parseCertificateDerEncoded`.
- "Platform notes" still described the old API's `enableRevocation` flag, which doesn't exist on the new validator pipeline - corrected.
- Added an Examples section linking to the three examples in waltid-examples.

Verified every code snippet actually compiles by extracting them into a temporary test file, compiling and running it against the local `waltid-x509` build, then deleted that file before committing (it was a doc-verification aid, not a real regression test).

Also updated the [waltid-examples README](https://github.com/walt-id/waltid-examples/pull/33) to reference the x509 examples, which weren't listed there at all.

JakeFernandes98 (approved):

PR #1903, opened 2026-07-14
first draft of android elegible tasks

JakeFernandes98:
Extends the selective CI runs to be more picky about running android jobs

### DIFF
--- a/.github/workflows/android-eligibility.yml
+++ b/.github/workflows/android-eligibility.yml
@@ -1,0 +1,160 @@
+name: Android eligibility
+
+on:
+  workflow_call:
+    inputs:
+      event-name:
+        description: "Caller event name"
+        required: true
+        type: string
+      ref:
+        description: "Caller git ref"
+        required: true
+        type: string
+      pr-number:
+        description: "Pull request number, if the caller is a PR event"
+        required: false
+        type: string
+        default: ""
+      pr-head-repo:
+        description: "Pull request head repository, if the caller is a PR event"
+        required: false
+        type: string
+        default: ""
+      pr-labels:
+        description: "JSON array of pull request label names"
+        required: false
+        type: string
+        default: "[]"
+      pr-draft:
+        description: "Whether the pull request is currently a draft"
+        required: false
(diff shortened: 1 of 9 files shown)


.github/workflows/android-eligibility.yml
```diff
@@ -0,0 +1,160 @@
...
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          set_decision() {
+            echo "should-run=$1" >> "$GITHUB_OUTPUT"
+            echo "reason=$2" >> "$GITHUB_OUTPUT"
+          }
+
+          is_android_relevant_path() {
```
szijpeter: I'd suggest to add
```
waltid-libraries/*/build.gradle.kts
waltid-libraries/*/*/build.gradle.kts
```

to mirror the setup in the macOS eligibility list. these are quite defensive, but changes in some of these files _could_ cause Android issues.

we could also consider
```
waltid-applications/waltid-android/*
```
until we still have it in the source, but I'm also fine with not including it

szijpeter (approved):
Looks pretty good to me already 👍🏼

Couple of non-blocking improvement suggestions to make this policy easier to operate and maintain:

1. Add a combined full-mobile opt-in label.

   Suggested semantics:
   - `ci:android`: run Android CI only
   - `ci:macos`: run iOS/macOS CI only
   - `ci:mobile`: run both Android and iOS/macOS CI

   That would keep the precise platform-specific escape hatches while providing one obvious label for a full mobile validation sweep.

2. Extract the shared eligibility logic.

   The Android eligibility workflow now mirrors most of the macOS/iOS workflow shape from https://github.com/walt-id/waltid-identity/pull/1887: event handling, same-repo/fork policy, draft behavior, label opt-ins, PR changed-file lookup, and `should-run` / `reason` outputs. It may be worth moving that common orchestration into one reusable mobile eligibility workflow, with Android and macOS/iOS passing only their platform-specific labels and path patterns.

   That should reduce drift between the Android and macOS/iOS policies as they evolve.

PR #1907, opened 2026-07-14
[WAL-1077] feat(mobile): fetch demo transaction data profiles

szijpeter:
## Summary

This follow-up lets the mobile demo apps load transaction-data profile support from the wallet API instead of carrying production demo profile definitions in each app. It stacks on [walt-id/waltid-identity#1893](https://github.com/walt-id/waltid-identity/pull/1893), which keeps the mobile SDK profile registry app-owned and empty by default.

The goal is to make profile support an app/runtime concern for the demo apps too: Android Compose, iOS Compose, and native iOS can point at a transaction-data profile endpoint, pass the fetched profiles into the SDK, and keep ordinary wallet flows usable if that endpoint is not configured or fails.

### Compose Demo Apps
- Adds a `transactionDataProfiles.url` Gradle property for the Android Compose demo app.
- Threads `TRANSACTION_DATA_PROFILES_URL` through the iOS Compose demo app.
- Fetches transaction-data profiles in shared mobile demo logic using Ktor JSON decoding.
- Passes fetched profiles into `MobileWalletConfig` on Android and iOS.

### Native iOS Demo App
- Reads `TRANSACTION_DATA_PROFILES_URL` from environment or user defaults.
- Fetches and decodes the same transaction-data profile payload before constructing `WalletConfiguration`.
- Uses an empty profile list when the endpoint is missing or unavailable, preserving the rest of the wallet demo.

### User Feedback
- Shows a persistent warning when transaction-data profiles could not be loaded.
- Logs the underlying fetch/configuration reason for debugging.
- Avoids silently falling back to bundled demo profiles, so stale local profile definitions cannot mask endpoint drift.

## Architecture Notes

- The SDK remains clean of demo profile defaults. The endpoint is consumed only by demo app wiring.
- Failed profile loading does not prevent credential receive, credential listing, or non-transaction-data presentation flows from working.
- Transaction-data presentation requests are rejected when the fetched profile list is empty, which is preferable to accepting requests with stale built-in demo profiles.

## Caveats and Follow-Ups

- The configured profile endpoint must stay aligned with the transaction-data demo requests. If the endpoint omits a field used by a demo request, the demo will surface the profile warning or reject that transaction-data request rather than substituting local definitions.
- This is intentionally a demo-app integration follow-up, not a mobile SDK endpoint client.

## Breaking

- Demo apps that want transaction-data request support now need to provide `transactionDataProfiles.url` / `TRANSACTION_DATA_PROFILES_URL`. Other wallet demo flows continue without it.

### DIFF
--- a/.github/scripts/mobile-ci/run-android-compose-demo-tests.sh
+++ b/.github/scripts/mobile-ci/run-android-compose-demo-tests.sh
@@ -6,4 +6,5 @@ identity_dir="$(cd "$script_dir/../../.." && pwd -P)"
 
 "$identity_dir/gradlew" -p "$identity_dir" \
   :waltid-applications:waltid-wallet-demo-compose:androidApp:connectedDebugAndroidTest \
+  -PtransactionDataProfiles.url=https://wallet.demo.walt.id/wallet-api/transaction-data-profiles \
   --info
--- a/docker-compose/verifier-api2/config/transaction-data-profiles.conf
+++ b/docker-compose/verifier-api2/config/transaction-data-profiles.conf
@@ -2,7 +2,7 @@ transactionDataProfiles = [
   {
     type = "org.waltid.transaction-data.payment-authorization"
     displayName = "Payment Authorization"
-    fields = ["amount", "currency", "payee"]
+    fields = ["amount", "currency", "payee", "reference"]
   },
   {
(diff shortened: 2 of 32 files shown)


szijpeter:
Folded into #1893.

PR #1912, opened 2026-07-15
[WAL-1184] fix(mobile) Compose demo PIN persistence

szijpeter:
## Summary

Fixes the Compose wallet demo relaunch behavior described in [WAL-1184](https://linear.app/walt-new/issue/WAL-1184/persist-compose-demo-pin-and-show-unlock-screen-after-relaunch).

The Android and iOS demos now ask users to create a PIN only when no PIN has been configured. Later launches show the unlock screen and accept the original PIN.

## What changed

- Added an explicit PIN-store dependency to the shared demo controller.
- Persisted a versioned, salted PBKDF2-SHA256 verifier per wallet ID in app-private Android/iOS preferences; the PIN itself is never persisted.
- Kept the wallet locked behind a retryable error state when PIN storage cannot be read, instead of falling back to PIN setup.
- Kept the web demo intentionally ephemeral and made that choice explicit at construction.
- Moved verifier derivation off the UI thread and disabled duplicate authentication submissions.
- Added controller and shared UI coverage plus deterministic Android and iOS relaunch tests that prove setup occurs once and the original PIN unlocks after restart.
- Documented how demo PIN persistence relates to wallet database persistence.

## Scope

This changes only the Compose demo app unlock flow. It does not change the mobile wallet SDK API, wallet database encryption, or credential persistence behavior.

## Caveats

- Clearing app data or uninstalling the demo resets PIN setup.
- The web preview remains in-memory by design.

## Breaking changes

None.

### DIFF
--- a/waltid-applications/waltid-wallet-demo-compose/README.md
+++ b/waltid-applications/waltid-wallet-demo-compose/README.md
@@ -30,6 +30,8 @@ The Compose iOS demo uses Kotlin direct Xcode integration and a local SwiftPM li
 
 Android and iOS demo targets use the default managed encrypted local persistence. Wallet database files are SQLCipher-encrypted, and managed database keys live in platform-protected storage. During local development, reset wallet state through `MobileWallet.deleteWallet()`, by uninstalling the app, or by deleting the app's local data.
 
+The demo unlock PIN is stored separately as a salted PBKDF2-SHA256 verifier in app-private preferences. The PIN itself is never persisted. Clearing app data or uninstalling the app resets the PIN setup flow together with the local wallet data.
+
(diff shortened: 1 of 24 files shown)


JakeFernandes98 (approved):
Looks good for the compose app, do we not need to implement the same on the native ios app?

szijpeter:
> Looks good for the compose app, do we not need to implement the same on the native ios app?

well spotted. but we don't have a pin screen in the native iOS demo app at all yet. tracked as https://linear.app/walt-new/issue/WAL-1189/add-pin-lock-screen-to-native-ios-wallet-demo

PR #1933, opened 2026-07-20
feat: Apk publish

dinkar-jain:


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,95 @@
...
+permissions:
+  contents: write
+
+jobs:
+  release:
+    name: Create release
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout all repos with ref fallback
+        uses: walt-id/waltid-identity/.github/actions/checkout-repos@864f84343fdeea17f6f63434bc8036c39f178dc7
```
szijpeter, line 29: `checkout-repos` fallback is the established pattern for coordinated branch builds here. For a release asset, should we explicitly pass `inputs.tag` as the ref — or reuse the APK from the earlier Gradle job — to guarantee the APK corresponds to the release tag? If releases are intentionally main-only and branch-based, could we document that contract?

note: the currently pinned `checkout-repos@864f84343fdeea17f6f63434bc8036c39f178dc7` does not expose a ref input. Could this use the established ref-aware pin `4296a7f80cacf7e60cb1a20321956768892eb7e3`, already used by `mobile-sdk-docs.yml` and pass `ref: ${{ inputs.tag }}`?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Enable Android build in unified-build root
+        if: ${{ steps.detect.outputs.exists == 'true' }}
+        uses: walt-id/waltid-identity/.github/actions/set-properties@00d4e55e9865cf17c6703d926dce4a4f6e1d404a
+        with:
+          file: 'gradle.properties'
+          properties: 'enableAndroidBuild=true'
+
+      - name: Build debug APK
+        if: ${{ steps.detect.outputs.exists == 'true' }}
+        run: ./gradlew :waltid-applications:waltid-wallet-demo-compose:androidApp:assembleDebug --no-daemon --console=plain --warning-mode=summary
```
szijpeter: Is publishing the **debug** variant intentional? `assembleDebug` is appropriate for tester sideloading, but it produces a debuggable APK signed with a debug certificate. If this is intended as an official distributable release, it should use a release variant and managed signing. Otherwise, please label the asset clearly as a debug/test build.

Stable signing is also required for in-place upgrades; APKs signed with different debug certificates cannot update one another. See: [Android signing considerations](https://developer.android.com/studio/publish/app-signing#considerations).

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,95 @@
...
+        id: apk
+        if: ${{ steps.detect.outputs.exists == 'true' }}
+        run: |
+          set -euo pipefail
+          apk=$(find waltid-identity/waltid-applications/waltid-wallet-demo-compose/androidApp/build/outputs/apk/debug -name '*.apk' | head -n 1)
+          if [ -z "$apk" ]; then
+            echo "::error::No APK found"
+            exit 1
+          fi
+          dest="waltid-wallet-demo-compose-${{ inputs.version }}.apk"
```
szijpeter, line 82: The filename uses `${{ inputs.version }}`, but the workflow does not propagate that version into the APK’s Android metadata; the app currently builds with `versionName = "0.1.0"` and `versionCode = 1`.

Could the build derive both values from the release version? Android recommends a greater `versionCode` for every successive release so upgrade/downgrade ordering works correctly, while `versionName` is the user-visible release version.

See [Android version settings](https://developer.android.com/studio/publish/versioning#versioningsettings).

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,95 @@
...
+          echo "path=$dest" >> "$GITHUB_OUTPUT"
+
+      - name: Create release and attach APK if present
+        uses: softprops/action-gh-release@v3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ inputs.tag }}
+          files: ${{ steps.apk.outputs.path }}
+          fail_on_unmatched_files: false
```
szijpeter, line 93: The module-missing path skips the APK build, and `fail_on_unmatched_files: false` then allows the release to succeed without an APK. Since this workflow promises an APK asset, please fail closed or consider exposing an explicit `apk_optional` compatibility mode.

szijpeter (commented):
Couple of questions & suggestions, changes LGTM otherwise.

Also: who is the intended audience for this artifact? If it is primarily for invited testers, should we publish through Firebase App Distribution or a Google Play testing track instead of exposing a debug APK through GitHub Releases?

JakeFernandes98:
> Couple of questions & suggestions, changes LGTM otherwise.
> 
> Also: who is the intended audience for this artifact? If it is primarily for invited testers, should we publish through Firebase App Distribution or a Google Play testing track instead of exposing a debug APK through GitHub Releases?

It can be for anyone looking to test our wallet or interoperability with our wallet stack. Also useful for our developers if they want a mobile to test with, they can just have the released versioned rather than building from source each time (as they may not necessarily need the latest features in the mainline)

I think firebase/google distribution is also a good option, we will have to do testflight for iOS anyway. I just saw this as an easier option.

As it's meant to be more of a testing tool, I think releasing the debug variant is ok. It is just a reference implementation at the end of the day and is not meant for end users to use as an actual wallet application. Let me know your thoughts

szijpeter:
> > Couple of questions & suggestions, changes LGTM otherwise.
> > Also: who is the intended audience for this artifact? If it is primarily for invited testers, should we publish through Firebase App Distribution or a Google Play testing track instead of exposing a debug APK through GitHub Releases?
> 
> It can be for anyone looking to test our wallet or interoperability with our wallet stack. Also useful for our developers if they want a mobile to test with, they can just have the released versioned rather than building from source each time (as they may not necessarily need the latest features in the mainline)
> 
> I think firebase/google distribution is also a good option, we will have to do testflight for iOS anyway. I just saw this as an easier option.
> 
> As it's meant to be more of a testing tool, I think releasing the debug variant is ok. It is just a reference implementation at the end of the day and is not meant for end users to use as an actual wallet application. Let me know your thoughts

That sounds reasonable. GitHub Releases is a pragmatic first channel for a public reference/testing tool; Firebase/Play can remain future options. 👍🏼

.github/workflows/create-release.yml
szijpeter: Update: a debug variant sounds appropriate for the stated [intent](https://github.com/walt-id/waltid-identity/pull/1933#issuecomment-5032732299).

One remaining consideration is upgrade behavior. This job runs on a fresh GitHub-hosted VM and does not persist `$HOME/.android/debug.keystore`, so each release APK will use a different debug certificate and cannot update a previous installation. Is uninstall/reinstall acceptable? If so, we should document it; otherwise the test APKs need a persistent non-production signing key.

.github/workflows/create-release.yml
dinkar-jain: The current release flow is branch-based by design. When we create a new major release (for example, `v20.0.0`), we automatically create a corresponding release branch (for example, `v20.x.x`). Subsequent minor/patch releases are made by cherry-picking the required commits onto that release branch and running the workflow from there.

`checkout-repos` checks out the release branch, so the APK is built from the code on that branch. Because of this, we don't need to explicitly pass a release tag as the `ref`, nor do we need to reuse the APK from an earlier Gradle job. The release artifact is built directly from the intended release branch.

.github/workflows/create-release.yml
dinkar-jain: This is the intended behavior. We intentionally allow the workflow to succeed without an APK in the module-missing path so that it can support release branches that don't contain the module. In those cases, skipping the APK build is expected rather than an error, so we don't want the workflow to fail.

.github/workflows/create-release.yml
szijpeter: thanks for the explanation, fine by me then 👍🏼

.github/workflows/create-release.yml
szijpeter: Fair enough. My concern was just that an unexpectedly missing module still produces a green release without an APK. I think we could still explicitly indicate whether an APK is expected and fail when it is missing. not a blocker though

.github/workflows/create-release.yml
dinkar-jain: Updated docs

.github/workflows/create-release.yml
szijpeter: The current `versionCode` derivation can still collide (for example across prereleases) but clean reinstalls make that non-blocking for this scope. Lets revisit it before supporting upgrades or Play distribution.

szijpeter (approved):
LGTM for the documented debug APK / clean-install scope 👍🏼

PR #1935, opened 2026-07-20
[WAL-995] feat(mobile): add mobile OpenID4VCI authorization-code issuance

szijpeter:
## Summary

Adds production OpenID4VCI 1.0 authorization-code credential issuance to the Android and iOS mobile wallet SDK for [WAL-995](https://linear.app/walt-new/issue/WAL-995/accept-credentials-from-issuer2-auth). A durable KMP issuance-session API serves authorization-code and pre-authorized grants, including typed preview, browser handoff, callback continuation, and typed outcomes.

### Shared issuance

- Adds durable, single-use issuance sessions with PKCE, PAR, DPoP, callback binding, deferred results, and replay protection.
- Uses issuer metadata for authorization-server and nonce behavior; proof nonces are requested only when `nonce_endpoint` is advertised, and otherwise omitted.
- Keeps credential proofs bound to the selected holder key and preserves immediate, deferred, cancelled, and failed outcomes.

### Mobile SDKs

- Exposes typed KMP and Swift session APIs and persists active continuations in encrypted mobile storage.
- Retains reviewed-preview receive APIs while adding authorization-code session transitions alongside them.
- Regenerates the public mobile ABI from the resolved source surface.

### Demo applications

- Wires browser authorization, bound callback continuation, cancellation, and typed outcomes into the Compose and native iOS wallet clients.
- Registers the mobile callback scheme on Android and iOS.

## Architecture Notes

- Protocol state, callback validation, token exchange, nonce acquisition, proof construction, and storage stay in the shared wallet layer; applications only drive typed transitions.
- Session transitions are mutex-protected, persisted atomically, and terminal transitions are single-use.

## Caveats and Follow-Ups

- No issuer or verifier server changes are included.
- Recent iOS smoke testing found that the public `issuer2.demo.walt.id` deployment emits duplicate `Location` headers for interactive sign-in, which iOS Safari coalesces into an invalid URL. The source fixes are in [#1852](https://github.com/walt-id/waltid-identity/pull/1852) and [#1934](https://github.com/walt-id/waltid-identity/pull/1934); the deployed revision appears not to have these fixes yet.
- Live authorization-code issuance still requires the issuer interactive browser step; the SDK owns continuation after the callback.

## Breaking Changes

None. Existing pre-authorized and reviewed-preview entry points remain available.

## Recordings

| Compose Android | Compose iOS | Native iOS |
| --- | --- | --- |
[...]

PR #1937, opened 2026-07-20
[WAL-1158] fix(openid4vci): align wallet nonce handling with 1.0

szijpeter:
## Summary

Implements [WAL-1158](https://linear.app/walt-new/issue/WAL-1158/wallet-align-openid4vci-nonce-handling-with-10) as the wallet-side OpenID4VCI 1.0 correction for proof nonce handling and `invalid_nonce` recovery.

When an issuer advertises `nonce_endpoint`, the wallet obtains a fresh `c_nonce` for each proof-bound credential request. When it does not, the wallet still creates the required proof without a `nonce` claim; token-response `c_nonce` values and access tokens are never used as proof nonces.

Enterprise PR: [walt-id/waltid-identity-enterprise#547](https://github.com/walt-id/waltid-identity-enterprise/pull/547)

## What Changed

- adds a reusable typed KMP `NonceRequestBuilder` for nonce-endpoint responses with sanitized failures
- makes proof nonce input nullable and omits the JWT `nonce` claim when no nonce was obtained
- obtains a fresh metadata-advertised nonce for each proof-bound credential request
- removes token-response `c_nonce` and access-token fallback from proof construction and isolated token results
- retries a full credential request exactly once after an issuer returns the standard `invalid_nonce` error, using a newly fetched nonce and freshly signed proof
- keeps isolated `fetch-credential` calls single-shot and propagates the typed credential error so API clients explicitly control any retry
- maps the exact `invalid_nonce` credential error to the Wallet2 HTTP API response
- updates focused wallet, route, and nonce-client coverage plus protocol documentation

## Architecture Notes

`NonceRequestBuilder` is deliberately a small protocol client: it requests and parses the nonce response, preserves cancellation, and redacts malformed response material. The higher-level issuance flow owns the bounded retry because it alone can fetch a nonce and regenerate a proof safely.

This PR is the prerequisite for [walt-id/waltid-identity#1935](https://github.com/walt-id/waltid-identity/pull/1935), which is restacked to consume the shared implementation instead of carrying WAL-1158 inside the broader WAL-995 change.

## Caveats and Follow-Ups

- The wallet intentionally applies no HTTPS-enforcement or redirect-rejection policy to `nonce_endpoint`; transport reachability remains the HTTP client's concern. This is the agreed non-standard-compliance trade-off for these PRs.
- Automatic recovery is limited to one retry and only to the standard `invalid_nonce` error. Other credential errors propagate unchanged.
- Isolated API consumers receive the typed error rather than an implicit replay, because they own the surrounding issuance state.

## Breaking Changes

- `TokenRequestBuilder.TokenResponse.c_nonce` and `c_nonce_expires_in` are removed.
- `RequestTokenResult.cNonce` is removed.
- `RequestNonceResult.nonce` and proof-builder nonce inputs are nullable so issuers can require proofs without advertising a Nonce Endpoint.

### DIFF
--- a/waltid-libraries/protocols/waltid-openid4vc-wallet-server/src/main/kotlin/id/walt/wallet2/server/handlers/Wallet2RouteHandler.kt
+++ b/waltid-libraries/protocols/waltid-openid4vc-wallet-server/src/main/kotlin/id/walt/wallet2/server/handlers/Wallet2RouteHandler.kt
@@ -585,6 +585,17 @@ object Wallet2RouteHandler {
                         )
                     }
 
+                    post("/request-nonce", {
+                        summary = "Isolated: obtain a fresh credential proof nonce"
+                        request { pathParameter<String>("walletId"); body<RequestNonceRequest>() }
+                        response { HttpStatusCode.OK to { body<RequestNonceResult>() } }
+                    }) {
+                        val req = call.receive<RequestNonceRequest>()
+                        call.respond(
+                            WalletIssuanceHandler.requestNonce(request = req)
+                        )
+                    }
+
                     post("/sign-proof", {
                         summary = "Isolated: sign a proof-of-possession JWT"
                         request { pathParameter<String>("walletId"); body<SignProofRequest>() }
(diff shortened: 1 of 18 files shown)


chsavvaidis (changes requested):
see [comment](https://github.com/walt-id/waltid-identity-enterprise/pull/547#issuecomment-5044832877)

chsavvaidis (approved):

PR #1932, opened 2026-07-20
refactor(mobile): enforce result model invariants

szijpeter:
## Summary

Tightens the Kotlin and Swift mobile-wallet result models so callers cannot construct contradictory or incomplete presentation and event states.

This PR now stacks on [walt-id/waltid-identity#1935](https://github.com/walt-id/waltid-identity/pull/1935). It preserves that PR’s session-bound preview handles while keeping ready and invalid presentation previews distinct.

### Closed Event Domain

- Replaces open event value objects with closed Kotlin and Swift enums covering all wallet-core issuance and presentation events.
- Maps the core Kotlin enum exhaustively through the mobile and Swift facades, so a new event requires an explicit consumer mapping.
- Derives each event’s phase and status from its case rather than accepting independently supplied values.

### Presentation Preview Results

- Makes a ready presentation preview’s `client_id` and `nonce` required and non-blank.
- Gives reportable invalid previews a separate partial request context with a required, non-blank `client_id` and optional `nonce`; a missing nonce can itself be the detected validation error.
- Retains the session-bound preview handle on invalid results, so dismissal or rejection remains tied to the reviewed request introduced by #1935.
- Rejects empty credential requirements or alternatives, blank query references, invalid returned credential-option references, and invalid selectable-disclosure combinations.

### Platform Parity

- Keeps Kotlin, SKIE-generated Swift, the Compose demo, and the native iOS demo aligned with the tightened model contracts.
- Updates the published Kotlin ABI baseline for the new public result types.

## Architecture Notes

- Events are enums because their domain is finite and carries no case-specific payload.
- Ready previews expose validated `MobileWalletPresentationRequestInfo` / `PresentationRequestInfo`. Invalid previews expose `MobileWalletPresentationRequestContext` / `PresentationRequestContext` so the API accurately represents a reportable error state without pretending a nonce was validated.
- Public constructors remain available for SDK fixture and preview construction; their checks enforce internally consistent values.

## Caveats and Follow-Ups

- Mobile wallet configuration and protocol-level DCQL grammar or aggregate request validation remain outside this result-model cleanup.

## Breaking Changes

`MobileWalletEvent` and Swift `WalletEvent` are closed enums. Callers constructing event fixtures must select a declared event case; Swift callers that parse a stable name can use `WalletEvent(name:)`.

Ready preview request `clientId` / `clientID` and `nonce` are now required and non-blank. Invalid preview results use the separate partial request-context type, whose client identifier is required while its nonce remains optional.

PR #1938, opened 2026-07-21
[WAL-1200] fix(mobile): bind actions to preview sessions

szijpeter:
## Summary

Resolved issuance and presentation previews were held in process-global caches keyed by wallet and source URL. Re-previewing the same URL could overwrite the resolution a user had reviewed, so a later receive, submit, or reject action could act on different protocol content.

This change binds every reviewed action to an opaque, wallet-scoped preview handle and carries that contract through Kotlin, SKIE, the Swift facade, Compose Android/iOS, and native iOS.

## What changed

- adds distinct issuance and presentation handle types so the two flows cannot be interchanged accidentally
- stores full resolved protocol state only behind a small internal bounded store with secure IDs, expiry, deterministic eviction, wallet binding, and explicit discard
- retains issuance previews after failed or cancelled attempts, consumes them after success, and guards concurrent retry use
- consumes presentation previews for submit and reject, while local dismissal explicitly discards them
- clears preview state during wallet deletion and preserves clear failure reasons for stale, unknown, expired, discarded, evicted, cross-wallet, in-use, and consumed handles
- keeps direct non-previewed issuance as the explicit immediate-receive path
- updates mobile demos and SDK documentation to use the reviewed handle rather than resolving by URL again

## API impact

Reviewed issuance now uses an `IssuancePreviewHandle`; reviewed presentation submit, reject, and discard use a separate `PresentationPreviewHandle`. Cached resolved protocol objects remain internal.

This is an intentional source-level API change for callers of the preview-based mobile wallet flows. The direct issuance API remains available for callers that intentionally skip review.

## Related work

- [WAL-1200](https://linear.app/walt-new/issue/WAL-1200/bind-mobile-wallet-actions-to-reviewed-preview-sessions)
- related model changes: [#1892](https://github.com/walt-id/waltid-identity/pull/1892), [#1932](https://github.com/walt-id/waltid-identity/pull/1932), and [#1926](https://github.com/walt-id/waltid-identity/pull/1926)

[#1892](https://github.com/walt-id/waltid-identity/pull/1892) is merged and included through the current `main` base. [#1932](https://github.com/walt-id/waltid-identity/pull/1932) and [#1926](https://github.com/walt-id/waltid-identity/pull/1926) remain open; their heads are not included.


JakeFernandes98 (approved):
This happened to me while testing a different PR :) Glad to see its resolved already. Approved, but need to check the merge conflicts

PR #1941, opened 2026-07-21
fix(ci:credentials): make npm installs reproducible

szijpeter:
## Summary

This makes `waltid-credentials` installs reproducible so unrelated branches no longer fail when the public npm dependency graph changes.

The application now commits its npm lockfile and uses `npm ci` consistently in Docker, the release workflow, and the local build helper.

## Root Cause

`waltid-credentials` ignored `package-lock.json`, copied only `package.json` into its Docker build, and ran a fresh `npm install`. Consequently, identical commits and the same `node:22-alpine` image could resolve different transitive dependency graphs.

On 2026-07-21, the `@vitejs/devtools` 0.4.2 package family expanded its optional peer graph. npm 10.9.8, bundled with the current Node image, then hit the open npm Arborist peer-resolution bug [npm/cli#5151](https://github.com/npm/cli/issues/5151):

> `Cannot read properties of null (reading 'edgesOut')`

Ordinary reruns did not recover because dependency resolution was deterministic against the new registry state, not a transient network failure.

## What Changed

- Track `package-lock.json` instead of excluding all `*-lock.json` files.
- Use `npm ci` in the multi-platform Docker build, credentials release workflow, and local build helper.
- Align the release workflow with the Docker image on Node.js 22.
- Record `legacy-peer-deps=true` for the locked graph, avoiding the npm 10 Arborist crash while preserving deterministic installs.

## Evidence

- [The original #1938 build failed on three attempts](https://github.com/walt-id/waltid-identity/actions/runs/29813240322) in the same credentials Docker step.
- The identical failure subsequently affected [#1936](https://github.com/walt-id/waltid-identity/actions/runs/29818966317), [#1937](https://github.com/walt-id/waltid-identity/actions/runs/29818349549), and [#1940](https://github.com/walt-id/waltid-identity/actions/runs/29816749126).
- [The multi-platform Docker build passed after applying this change on #1938](https://github.com/walt-id/waltid-identity/actions/runs/29816845667).

## Caveats and Follow-Ups

- `legacy-peer-deps` remains a compatibility workaround for npm 10. It can be reconsidered once the dependency graph or bundled npm resolver no longer triggers the upstream bug.

## Breaking

None. Dependency installation becomes locked and reproducible; application behavior is unchanged.

### DIFF
--- a/.github/workflows/waltid-credentials-release.yml
+++ b/.github/workflows/waltid-credentials-release.yml
@@ -24,9 +24,9 @@ jobs:
       - name: Setup Node.js 🏗
         uses: actions/setup-node@v6
         with:
-          node-version: '20.x'
+          node-version: '22.x'
       - name: Install dependencies 👨🏻‍💻
-        run: npm install
+        run: npm ci
       - name: Build
         run: npm run build
       - name: Set up Docker Buildx
--- a/waltid-applications/waltid-credentials/.gitignore
+++ b/waltid-applications/waltid-credentials/.gitignore
@@ -154,4 +154,5 @@ dist
 *-lock.yml
 *.lock
 *.lockb
+!package-lock.json
 .output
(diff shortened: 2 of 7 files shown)


philpotisk (approved):
🙏

PR #1954, opened 2026-07-22
Crypto2

waltkb:
https://github.com/walt-id/waltid-identity-enterprise/pull/550


waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/WalletPresentFunctionality2.kt
```diff
@@ -602,45 +798,37 @@ object WalletPresentFunctionality2 {
...
         log.trace { "- Start of Wallet Present Handling -" }
         log.trace { "Wallet presentation will use key $holderKey, and did $holderDid" }
 
         // Step 1: Resolve AuthorizationRequest.
-        val resolvedRequest = resolvedAuthorizationRequest ?: try {
-            resolveAuthorizationRequestObject(presentationRequestUrl, unsignedRequestObjectPolicy)
-        } catch (error: AuthorizationRequestResolver.SignedAuthorizationRequestValidationException) {
-            legacyFallbackResult(presentationRequestUrl, legacyFallbackCallback, error)
+        if (resolvedAuthorizationRequest == null) {
+            legacyFallbackResult(presentationRequestUrl, legacyFallbackCallback)
```
szijpeter: Isn't this changing the fallback from a recovery path after a narrowly identified legacy-resolution failure into a path attempted before normal authorization-request resolution?

When a fallback callback is configured, the `mdoc-openid4vp` scheme or the presence of `presentation_definition` / `presentation_definition_uri` is enough to divert the request before `AuthorizationRequestResolver.resolve()` and its client-ID trust validation execute.

Is this the intended shape?

I'd lean towards keeping the strict resolver as the primary path and invoke the fallback only after an explicitly recognized legacy-resolution failure, or require the caller to select an explicit legacy mode.

note: the same pre-resolution fallback also exists in the standalone `resolveAuthorizationRequest() `overload; if adjusted both paths should follow the same resolver-first behavior.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -477,7 +526,7 @@ object WalletPresentationHandler {
         val resolvedAuthorizationRequest = (preview as? PreviewedPresentation.Ready)
             ?.resolvedAuthorizationRequest
             ?: error("Unexpected presentation preview state")
-        val key = wallet.resolveKey(keyId = null)
+        val keyMaterial = wallet.resolveKeyMaterial(null, setOf(KeyUsage.SIGN))
```
szijpeter: Previously, preview validation was based on the same default-key selection strategy used by submission. This change validates against the union of every signing key in the wallet, while submission still independently resolves one unspecified default key.

In a multi-key wallet, preview can therefore accept a request because one key supports the requested algorithm, but submission can select another incompatible key.

I think we should either select and retain a concrete execution key during preview or validate against the exact key that submission will use.

waltid-libraries/protocols/waltid-openid4vci-wallet/src/commonMain/kotlin/id/waltid/openid4vci/wallet/proof/ProofOfPossessionBuilder.kt
```diff
@@ -29,6 +32,15 @@ interface ProofOfPossessionBuilder {
     val proofType: String
 }
 
+interface Crypto2ProofOfPossessionBuilder {
+    suspend fun buildProof(
+        key: Crypto2Key,
+        algorithm: JwsAlgorithm,
+        audience: String,
+        nonce: String,
```
szijpeter: The legacy contract and OpenID4VCI flow allow the `nonce` to be absent (when the issuer has no nonce-endpoint, [see](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#jwt-proof-type)), but this Crypto2 contract narrows it to be required. This prevents the Crypto2 abstraction from representing valid proof requests and is already forcing production code to bypass the interface.

Shall we keep this `String?`, preserve validation of blank-but-present values and route the wallet issuance implementation through the interface rather than maintaining a separate proof-building path?

waltid-libraries/protocols/waltid-openid4vp-verifier/src/jvmMain/kotlin/id/walt/verifier2/handlers/sessioncreation/VerificationSessionCreator.kt
```diff
@@ -152,34 +292,22 @@ object VerificationSessionCreator {
...
-                    )
-                )
+                vpFormatsSupported = baseMetadata.vpFormatsSupported ?: defaultVpFormatsSupported
             )
+        }).let { metadata ->
+            if (isSiop) {
+                metadata.copy(
+                    subjectSyntaxTypesSupported = metadata.subjectSyntaxTypesSupported
+                        ?: listOf("did", "urn:ietf:params:oauth:jwk-thumbprint"),
+                    idTokenSignedResponseAlg = metadata.idTokenSignedResponseAlg ?: "RS256",
```
szijpeter: This defaults SIOPv2 ID Token signing to `RS256`, while the default mobile wallet uses a `P-256 `key and can only sign with `ES256`.

Because the wallet enforces the verifier-requested algorithm, the default `vp_token id_token` flow would fail before signing.

Could we make the algorithm configurable and use `ES256` as the default for the current P-256 mobile setup?

waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-server/src/main/kotlin/id/walt/wallet2/persistence/ExposedKeyStore.kt
szijpeter, line 1: This store still persists the legacy serialized key as the authoritative record and derives/reconciles the Crypto2 descriptor from it. It cannot persist a Crypto2-only key and a stale sidecar is replaced from the legacy representation.

That is reasonable as a temporary migration adapter, but it does not complete the Crypto2 migration for Wallet API2.

Could we make a versioned `StoredKey` descriptor the canonical persisted representation and keep legacy data only for migration?

The persistence contract should support software and managed Crypto2 keys without requiring a serializable legacy Key.

waltid-services/waltid-wallet-api2/src/main/kotlin/id/walt/wallet2/OSSWallet2AuthConfig.kt
szijpeter, line 1: The Crypto2 descriptor is optional while the legacy private key remains mandatory. This means auth cannot be configured with a managed or otherwise Crypto2-only key, and every deployment must continue embedding the legacy private representation.

Could we allow a standalone `StoredKey` configuration and resolve it through a configured Crypto2 runtime/provider?

szijpeter (commented):
I reviewed the mobile wallet, Wallet API2, OpenID4VP wallet, and Verifier2 changes in this PR.

Overall, this is exceptional work - especially given the size and complexity of the migration. 👏🏼

For mobile, I prepared two follow-up PRs:

- https://github.com/walt-id/waltid-identity/pull/1992 - primarily fixes the CI failures and addresses some correctness issues I noticed. I think this could be folded into this PR after review.
- https://github.com/walt-id/waltid-identity/pull/1995 - removes the mobile migration and legacy compatibility paths to reduce unnecessary complexity. This can remain a follow-up or be folded into this PR; I’ll leave that decision to you.

Please find the rest of my findings below. Most are suggestions for consideration, while a few identify behavior that I think should be addressed before merging.

JakeFernandes98 (approved):

PR #1970, opened 2026-07-23
Fix/wallet cleanup

JakeFernandes98:
- Adds issuer and request metadata to stored credential when receiving it in the wallet, with UI additions to example applications
- Extends walletsdk and examples apps to store this additional metadata
- Adds new methods to the PresentationHandler to be able to implement isolated flows in the APIs and Enterprise UI which better mirror the mobile wallet UX experience
- Extends the wallet-api2 with new endpoints implementing these new isolated steps for presentation

ES:https://github.com/walt-id/waltid-identity-enterprise/pull/532


waltid-libraries/protocols/waltid-openid4vci-wallet/src/commonMain/kotlin/id/waltid/openid4vci/wallet/token/TokenRequestBuilder.kt
```diff
@@ -226,7 +226,7 @@ class TokenRequestBuilder(
             log.error {
                 "Token request failed - Status: ${response.status.value} ${response.status.description}"
             }
-            throw Exception("Token request failed. Status: ${response.status}")
+            throw IllegalArgumentException("Token request failed. Status: ${response.status}")
```
JakeFernandes98, line 229: This was changed to align with the ES Integration tests, which are expecting a 401 and not 500 from failed token requests. However, if we want to revert it or handle it a different way then its ok as long as we update the integration tests

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -1078,30 +1200,48 @@ object WalletPresentationHandler {
...
+ * 3. `POST /present/send-response` - transmits the response to the verifier
+ *
+ * Prefer [selectedCredentialOptions] (+ optional [selectedDisclosureOptions]) for consent UIs.
+ * [selectedCredentialIds] remains for the legacy resolve → match → build path.
  */
 @Serializable
 data class BuildVpTokenRequest(
-    /** The resolved authorization request from step 1. */
+    /** The resolved authorization request from preview / resolve-request. */
     val authorizationRequest: AuthorizationRequest,
```
szijpeter: The new stateless `/preview` flow returns an `AuthorizationRequest` which the client echoes into `/build-vp-token` and `/send-response`. Those endpoints then trust client-controlled request fields, including DCQL, nonce, response URI/mode, and response-encryption parameters.

I'd suggest not use an echoed `AuthorizationRequest` as the continuation contract. Either retain a wallet-bound preview handle, or require the original `requestUrl` and re-resolve/revalidate it inside each sensitive step. The latter avoids client-side request tampering, though a retained handle is preferable when the implementation must bind submission to the exact reviewed request.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
```diff
@@ -960,26 +1016,59 @@ object WalletIssuanceHandler {
     private suspend fun Wallet.parseAndStore(
         issuedCredential: id.walt.openid4vci.responses.credential.IssuedCredential,
         label: String? = null,
+        metadata: JsonObject? = null,
```
szijpeter, line 1121: Is it intentional that issuer display metadata is only propagated through the pre-authorized full-receive path? 

`fetchCredential(storeInWallet = true)`, `pollDeferredFlow` and the authorization-code continuation all store credentials without the metadata (and label) as I see it.

szijpeter (commented):
couple of questions/suggestions, LGTM otherwise

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
JakeFernandes98: I originally used a handle implementation, similar to your mobile implementation - the problem is that for our APIs it now introduces state which if stored in memory will not work in horizontally scaled deployments, and if stored in redis or mongo would increase the complexity of the solution. This is why the stateless version was introduced as an alternative. 

I understand the security concerns, and will look at implementing re-resolution of the request in the flow

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
JakeFernandes98: This should now be extended via [c388e73](https://github.com/walt-id/waltid-identity/pull/1970/commits/c388e73e3f132792f488a4f74a0826321c73cf77)

chsavvaidis (approved):

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -900,6 +1050,33 @@ object WalletPresentationHandler {
...
+     * client-echoed [AuthorizationRequest].
+     */
+    private suspend fun resolveAndValidatePresentationRequest(
+        wallet: Wallet,
+        requestUrl: Url,
+        transactionDataTypeRegistry: TransactionDataTypeRegistry,
+        resolveAuthorizationRequest: suspend (Url) -> ResolvedAuthorizationRequest,
+    ): AuthorizationRequest {
+        val resolvedAuthorizationRequest = resolveAuthorizationRequest(requestUrl)
+        val key = wallet.resolveKey(keyId = null)
```
szijpeter: This validates presentation capabilities using the wallet's default key, while `buildVpToken` may subsequently sign using the explicitly supplied `request.key` or `request.keyId`.

This means validation and signing can use different key types and algorithm capabilities. Could the build path resolve the effective signing key first and validate against that key?

Also, `sendAuthorizationResponse` calls this helper and therefore requires a default key even though it does not perform signing. It may be cleaner to separate key-independent request validation from signing-capability validation.

JakeFernandes98: Addressed: `resolveAndValidatePresentationRequest` now runs key-independent request validation only by default.

- `buildVpToken` resolves the effective signing key first and validates format/algorithm capabilities against that key.
- `sendAuthorizationResponse` no longer requires a default wallet key; it only re-resolves and runs key-independent validation before transmitting.

`PresentationRequestValidator.validate` accepts `formatCapabilities = null` to skip signing-capability checks.

szijpeter (approved):
thanks for the update, LGTM now 👍🏼

PR #1980, opened 2026-07-24
extend did creation support in wallet2

### LINKED ISSUE #1965: Creating DIDs with method `did:web` using the Wallet2 API
Creating DIDs using the Wallet2 API does not currently support accepting additional arguments, which might be needed by a specific DID registrar:

https://github.com/walt-id/waltid-identity/blob/7b837281a652590bfe72cb63a4f9de2b429e6661/waltid-libraries/protocols/waltid-openid4vc-wallet-server/src/main/kotlin/id/walt/wallet2/server/handlers/Wallet2RouteHandler.kt#L384

Thus it is currently not possible to pass the `domain` and `path` for creating DIDs with the method `did:web`:

https://github.com/walt-id/waltid-identity/blob/7b837281a652590bfe72cb63a4f9de2b429e6661/waltid-libraries/waltid-did/src/commonMain/kotlin/id/walt/did/dids/DidService.kt#L197

JakeFernandes98:
Resolves walt-id/waltid-identity#1965 <br>Docs: [https://github.com/walt-id/waltid-docs/pull/327](<https://github.com/walt-id/waltid-docs/pull/327>)


waltid-libraries/protocols/waltid-openid4vc-wallet-server/src/main/kotlin/id/walt/wallet2/server/handlers/Wallet2RouteHandler.kt
```diff
@@ -86,7 +87,16 @@ data class ImportKeyRequest(
...
     val method: String,
-    val keyId: String? = null
+    val keyId: String? = null,
+    /**
+     * Method-specific registrar arguments forwarded to [DidService.registerDefaultDidMethodByKey].
+     *
+     * Examples:
+     * - did:web — `domain` (required), `path` (optional)
+     * - did:key — `useJwkJcsPub` (optional boolean)
+     * - did:cheqd — `network` (`testnet` or `mainnet`)
```
SuperBatata: The request to the cheqd registrar's /1.0/create endpoint is being sent as a GET with a JSON body:

GET /1.0/create
Content-Type: application/json

{ ... }

The registrar responds with 400 Bad Request (plain text), and we then attempt to deserialize "Bad request" into JobActionResponse, which results in a misleading JsonDecodingException.

Could you verify whether this endpoint should be invoked with POST instead of GET? If so, switching the request method should resolve the 400.

Additionally, it would be good to handle non-success responses before attempting deserialization, e.g. by checking response.status and surfacing the response body in the exception. That would expose the actual server error instead of a JSON parsing exception.

SuperBatata (changes requested):

waltid-libraries/protocols/waltid-openid4vc-wallet-server/src/main/kotlin/id/walt/wallet2/server/handlers/Wallet2RouteHandler.kt
JakeFernandes98: Looks like our cheqd implementation is out of date

SuperBatata (approved):

PR #1991, opened 2026-07-27
Align docker compose ports and ensure docker urls are consistent

JakeFernandes98:
fixes verifier2 docker compose config: urlPrefix is hardcoded to http://localhost:7003/verification-session

Aligns ports in docker-compose and the service config files so they are the same

Splits the compose profiles into identity-old and identity, based on the protocol versions

Disable client attestation in issuer2 (as our OSS walletapi does not have access to an attester by default)


docker-compose/.env
```diff
@@ -1,4 +1,4 @@
-VERSION_TAG=stable
+VERSION_TAG=latest
```
dinkar-jain: Is it intentional?

JakeFernandes98: reverted

dinkar-jain (approved):

PR #2014, opened 2026-07-30
wrap body request errors in friendlier error messages

JakeFernandes98:
Turns response bodies like:

```
{
  "exception": true,
  "id": "BadRequestException",
  "status": "Bad Request",
  "code": "400",
  "message": "Failed to convert request body to class id.walt.crypto.keys.TypedKeyGenerationRequest",
  "cause1_exception": "JsonConvertException",
  "cause1_message": "Illegal input: Class discriminator was missing and no default serializers were registered in the polymorphic scope of 'TypedKeyGenerationRequest'.\nJSON input: {\"keyType\":\"Ed25519\"}",
  "cause2_exception": "JsonDecodingException",
  "cause2_message": "Class discriminator was missing and no default serializers were registered in the polymorphic scope of 'TypedKeyGenerationRequest'.\nJSON input: {\"keyType\":\"Ed25519\"}"
}
```

into something like

```
{
  "exception": true,
  "id": "BadRequestException",
  "status": "Bad Request",
  "code": "400",
  "message": "Request body is missing required discriminator field 'backend' for TypedKeyGenerationRequest",
  "details": {
    "expectedType": "TypedKeyGenerationRequest",
    "discriminator": "backend"
  }
}
```

which can be useful for first time users and reduces the number of questions we get

In some cases, the number of causes may be too many that the actual reason why the body is rejected is not in the response but has to be searched for in the logs. This also tried to resolve this DevEx issue

ES: https://github.com/walt-id/waltid-identity-enterprise/pull/566 (fixes test which relied on older error message)


waltid-services/waltid-service-commons/src/main/kotlin/id/walt/commons/web/plugins/RequestBodyErrorMessages.kt
```diff
@@ -0,0 +1,241 @@
...
+
+private fun extractExpectedType(cause: Throwable): String? {
+    for (throwable in causeChain(cause)) {
+        val message = throwable.message ?: continue
+        FAILED_TO_CONVERT_REGEX.find(message)?.groupValues?.get(1)
+            ?.substringAfterLast('.')
+            ?.takeIf { it.isNotBlank() }
+            ?.let { return it }
+
+        Regex("""expected type:\s*([A-Za-z0-9_.$]+)""", RegexOption.IGNORE_CASE)
```
SuperBatata: Every other regex is already a top level constant i'd extract it alongside the others

SuperBatata (changes requested):

SuperBatata (approved):

PR #2045, opened 2026-08-05
[WAL-1211] feat(wallet): support signed issuer metadata in legacy wallet API

szijpeter:
## Summary

This follow-up brings the signed issuer-metadata trust boundary to the legacy `waltid-wallet-api` wallet1 issuance flows. It covers both ordinary issuance and externally signed issuance, so those active paths no longer bypass the resolver introduced by [walt-id/waltid-identity#2011](https://github.com/walt-id/waltid-identity/pull/2011).

The matched Enterprise companion is [walt-id/waltid-identity-enterprise#585](https://github.com/walt-id/waltid-identity-enterprise/pull/585).

## What Changed

- Added a shared modern-to-legacy metadata adapter for wallet1's `OpenIDProviderMetadata.Draft13` model.
- Added explicit issuer-to-public-JWK trust configuration, with optional `kid` and `alg` binding.
- Routed normal issuance plus external-signature prepare/submit through one configured resolver.
- Reused the resolved provider metadata across each issuance flow instead of resolving it again.
- Added focused adapter coverage for unsigned projection, trusted signed metadata, and rejected signed metadata.

## Architecture Notes

- The legacy adapter lives beside the modern wallet protocol code and is the compatibility boundary for wallet1 callers.
- A JWK supplied by a JWT is not treated as trusted authority; trust comes only from the embedding service's issuer-specific configuration.
- Existing deployments remain unsigned-only unless `TrustConfig.issuerMetadataSigners` is populated.

## Caveats and Follow-Ups

- This PR intentionally does not change the modern wallet2/mobile API or Enterprise wallet1 wiring. The matched Enterprise companion applies the same adapter and trust boundary to `WalletOldImplementation`.
- Public signed-metadata journeys still depend on the issuer services publishing signed metadata and suitable trust configuration.

## Breaking

- No existing wallet1 request shape changes. The new trust configuration is optional and defaults to the previous unsigned-metadata behavior.

### DIFF
--- a/waltid-libraries/protocols/waltid-openid4vc-wallet/build.gradle.kts
+++ b/waltid-libraries/protocols/waltid-openid4vc-wallet/build.gradle.kts
@@ -37,6 +37,8 @@ kotlin {
             // OpenID4VCI 1.0 — wallet-side client logic (offer parsing, token, proof)
             // package: id.waltid.openid4vci.wallet.*
             api(project(":waltid-libraries:protocols:waltid-openid4vci-wallet"))
+            // Legacy OpenID4VC models used by wallet1 compatibility flows.
+            api(project(":waltid-libraries:protocols:waltid-openid4vc"))
 
             // OpenID4VP 1.0 — core protocol types (AuthorizationRequest, response modes, etc.)
             // package: id.walt.verifier.openid.*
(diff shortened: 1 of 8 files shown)


szijpeter:
legacy support not needed

PR #1885, opened 2026-07-09
[WAL-896] feat(wallet): finalize authenticated OpenID4VP presentation flow

philpotisk:
## Summary

This is the WAL-896 wallet, protocol, and mobile consumer half, stacked directly on [the verifier prerequisite #2055](https://github.com/walt-id/waltid-identity/pull/2055). It is based on `feature/wal-896-verifier`, which now sits on current `main` after [walt-id/waltid-identity#2031](https://github.com/walt-id/waltid-identity/pull/2031).

The branch makes wallet2/mobile presentation consume authenticated Final Request Objects while preserving the stateless HTTP continuation API and the retained mobile preview flow. The coordinated Enterprise counterpart is [walt-id/waltid-identity-enterprise#553](https://github.com/walt-id/waltid-identity-enterprise/pull/553).

### Authenticated wallet2 and mobile flow

- Requires strict signed Request Object resolution for wallet2/mobile presentation, including the expected audience and wallet-controlled client trust configuration.
- Re-resolves the immutable signed request URL for preview, build, response, and rejection so caller-supplied request state cannot alter the reviewed transaction.
- Keeps both continuation models: single-use retained preview state for the mobile consent flow, and stateless preview/build/send/reject operations bound to the same signed request URL.
- Preserves the wallet2 routes and request-based APIs introduced by #1970; this branch wires trust and authenticated resolution through them rather than restoring caller-echoed `AuthorizationRequest` state.
- Exposes the negotiated response mode in the mobile presentation request model and refreshes the generated ABI/API snapshot.

### DCQL and Crypto2 response handling

- Enforces DCQL Final semantics, credential-set references, complete disclosure paths, trusted-authority handling, and typed required-credential availability failures.
- Treats a missing required credential as a protocol-level preview error that can be rejected safely; malformed or unsupported matcher constraints remain fail-closed.
- Makes Crypto2 response-encryption selection deterministic and binds the selected verifier key and encryption parameters to the authenticated request.

### X.509 and mobile fixtures

- Adds platform-specific X.509 certificate-chain validation needed by the wallet client-identifier path, with explicit trust-result handling and no system-trust fallback.
- Adds signed Request Object fixtures and Android/iOS mobile coverage for the signed inline URL and Request URI POST contracts.

## Stack and ownership

- Immediate base: `feature/wal-896-verifier` / [#2055](https://github.com/walt-id/waltid-identity/pull/2055) on current `main`.
- The native GitHub stack is `#2055 → #1885 → #2091`.
[...]

PR #2063, opened 2026-08-08
Harden and finish the native Digital Credentials presentation (review follow-up to #2062)

szijpeter:
Stacked on **#2062** (`feature/wal-752-dc-api-v2`), which it targets rather than `main`. Merge #2062 first, or merge
this into it.

Addresses two rounds of review on #2062. The goal was to finish, simplify and harden the native Digital Credentials
presentation while keeping the PR narrowly focused on the capabilities it actually intends to deliver. Nothing here
expands scope: the signed DC API variants remain explicitly unsupported and fail closed, and no dependency on
#1885/#1939, legacy crypto APIs or deprecated key fallbacks was reintroduced.

25 commits, 51 files, +5452/−567 against #2062's head.

## Android support status

| Capability | Status |
| --- | --- |
| `openid4vp-v1-unsigned` | supported |
| `mso_mdoc` (ISO mdoc) | supported |
| `dc+sd-jwt` (SD-JWT VC) | supported |
| DCQL, including multi-credential queries | supported |
| Multiple alternative DC requests in one envelope | supported |
| `transaction_data` | supported |
| `response_mode=dc_api` | supported |
| `response_mode=dc_api.jwt` | supported |
| `org-iso-mdoc` (ISO 18013-7 Annex C) | supported |
| `openid4vp-v1-signed` | **not supported** |
| `openid4vp-v1-multisigned` | **not supported** |

The two signed variants are reported `supported = false` with explicit reasons, and there is no partially functional
signed path. If a signed variant is the alternative the **user selected**, the request **fails** — it does not
silently fall back to a different alternative the user did not choose. An envelope offering only unsupported
protocols is rejected outright.

### Two capabilities were finished rather than descoped

- **`response_mode=dc_api.jwt`** (`36787c0d3`, `8b502d29c`). Response members are returned as a single `response`
  JWE encrypted to the verifier's `client_metadata` key, so neither the OS nor the calling website can read the
  disclosed claims. It fails closed without usable verifier encryption keys rather than degrading to a cleartext
  response the verifier would still accept — and that validation now runs during **preview**, before any credential
  is read, rather than at response construction. #2062's description said this was "not wired up"; that was stale.
- **Multi-request envelopes** (`1390d8833`, `d3b6c3f7f`). An OpenID4VP 1.0 Appendix A `requests` array is answered
  against the alternative the **matcher selection** names, not one the wallet picks by preference. The first
  version of this preferred OpenID4VP over Annex C regardless of what the user chose, which answers a request the
  user never saw; `resolveSelectedProtocolRequest` now derives the index from Credential Manager's attribution and
  permits inference only for a single-alternative envelope, or for a protocol matching exactly one offered
  alternative. The two registered matchers attribute differently — the AndroidX OpenID4VP matcher carries
  `dc_request_index` in `metadata`, while Multipaz's Annex C matcher has no metadata channel and attributes by
  protocol — and each is parsed on its own terms rather than through an invented shared convention.

### Correctness fixes

- The platform registry is a best-effort projection of committed wallet state, never an authority over it
  (`2bc5c7535`). A registry that rejects the projection can no longer turn a stored credential into a failed
  issuance or an unsuccessful deletion; the failure surfaces through `digitalCredentialRegistration`, which is
  where a stale projection belongs.
- Only the protocols the wallet can fulfill are advertised as supported (`5dd1f7af4`), on both platforms, so a
  verifier learns what is unavailable up front instead of mid-flow.
- Reader trust states are truthful (`276cbb7fa`). `Unverified` — which conflated "not checked yet" with "checked
  and not accepted" — is replaced by a five-state hierarchy: `NotApplicable`, `NotAuthenticated`,
  `PendingRawRequest`, and, only after cryptographic verification succeeded, `Untrusted(reason)` or
  `Trusted(certificateSubject)`. The latter two form a `MobileWalletReaderTrustDecision` sub-interface, so an
  evaluator cannot return a state meaning "not evaluated". The default evaluator answers `Untrusted`: a valid
  signature alone is not identification.
- The provider activity shares one wallet construction with the app (`370e8e103`) instead of building its own
  independently-configured wallet that could disagree on key type, locales or trust config.

### Simplification

- `DcApiWallet` is the only origin canonicalizer (`03717db9c`). The Android copy is gone. This was not cosmetic:
  the canonical origin is hashed into the mdoc session transcript, and the two implementations disagreed on host
  case and on bracketed IPv6 loopback — each disagreement yielding a device signature no verifier can reproduce.
  `canonicalizePlatformOrigin` is public and idempotent so adapters call it rather than reimplementing it.
- Annex C crypto lives in the layers that own it (`d5adaa715`). HPKE recipient-key shape rules moved into
  `waltid-crypto2` as `Hpke.validateRecipientPublicKey`, so the wallet cannot drift from what `Hpke.sealBase`
  accepts. No ISO structures were pushed into crypto2.
- The Annex C flow reads as its sequence (`205147795`) — validate, match, evaluate reader trust, confirm the
  request is unchanged after consent, build documents, encrypt — without splitting each step into a class.
- Dead surface removed and cross-language identifiers stabilised (`9d52e2769`): the never-thrown
  `MobileWalletDigitalCredentialCancellationException` is gone, the capability enums expose an explicit
  `identifier` instead of leaking a generated Objective-C class name through `String(describing:)`, and the Swift
  registration-status key is real API rather than a literal hand-copied from a Kotlin `internal` constant.

### Test coverage

Unit and host-test coverage for Android Annex C through Credential Manager, the mobile Digital Credentials
presentation facade, the unguarded Annex C rejection paths, and issuance-side registry synchronization failure
(`3a1734706`, `b1cdb87e2`, `f26cca539`, `350e95934`). The last needed a seam: every `MobileWallet` issuance entry
point performs real HTTP, and `WalletIssuanceSessionService` already accepted an `httpClient` override that simply
was not forwarded. It is now a defaulted, last-position parameter on `MobileWallet`'s **internal** constructor —
production omits it, and the published klib ABI is unchanged.

`DigitalCredentialSharingE2ETest` is now the single native sharing suite (`28f953db2`), one test per protocol
variant the wallet claims, each verified by the code that would verify it in production:

| Test | Covers |
| --- | --- |
| `sharesMdocThroughClearOpenId4VpDcApi` | baseline `dc_api` mdoc, verifier2's real policy set |
| `sharesSdJwtWithTransactionDataThroughCredentialManager` | payment fields on the consent surface **before** Share, and the `transaction_data_hashes` recomputed from the request and compared to the KB-JWT |
| `sharesMdocThroughEncryptedDcApiJwt` | `dc_api.jwt`; asserts the exact member set, so encrypting *and* leaving the clear members does not pass |
| `selectsNonZeroAnnexCAlternativeFromMultiProtocolRequest` | two alternatives, the non-zero Annex C one selected; absorbs the former `AnnexCSharingE2ETest` |

Annex C is verified in-process because it has no back-channel: the encrypted DeviceResponse returns through the OS
to whoever called `getCredential`, so reader and caller are necessarily the same party. The Annex C request comes
from the shared verifier-side `AnnexCRequestBuilder`, so no verifier session is faked.

**On-device limit, found by running this suite (`4f9681115`) and worth knowing before relying on the capability
table above.** An OpenID4VP alternative and an Annex C alternative cannot both surface candidates in one envelope on
Android today, in either array order, for two independent platform reasons:

1. **Annex C after OpenID4VP — mechanism identified.** Multipaz's Annex C matcher iterates `requests[]` and stops at
   the first entry whose `protocol` it recognises — and it recognises all three `openid4vp-v1-*` values as well as
   `org-iso-mdoc`. An OpenID4VP entry ahead of the Annex C one ends the scan before Annex C is reached, whether or
   not it matches anything. This is read off the decompiled `$matcher` function, not inferred: the `br_table`
   dispatch on protocol-string length, and the loop tail that exits once a match is recorded.
2. **Annex C before a matchable OpenID4VP entry — observed, not explained.** With Annex C at index 0 and a
   *matchable* OpenID4VP entry at index 1, no Annex C candidate reaches the picker either. Making that second entry
   unmatchable makes Annex C surface, and scrolled screenshots confirm only one candidate card exists rather than one
   below the fold. I could not establish *why*: the matchers' `printf` output never reached logcat, so there is no
   trace, and nothing in the decompiled matcher explains it. Registration order (OpenID4VP registry first, Annex C
   second) is a candidate but unverified. Treat this one as a reproducible observation with an unknown mechanism.

Neither is wallet behaviour and neither is fixable from wallet code — `OpenId4VpRegistry` takes no matcher and
accepts none, so AndroidX's embedded default matcher cannot be replaced. The on-device test therefore uses `preview`
(the legacy DC protocol identifier, which this wallet reports as unsupported) at index 0, which still proves what it
was written to prove: a non-zero selected index, and Multipaz's by-protocol attribution that carries no index at all.
Selection across a genuinely mixed OpenID4VP/Annex C envelope stays covered by
`AndroidDigitalCredentialProviderTest`, in both orders, without the platform picker. So "multiple alternative DC
requests in one envelope" is supported by the wallet's resolution logic; what a mixed-protocol envelope actually
shows the user is the platform's call.

`canonicalizePlatformOrigin`'s accepted and rejected matrices are now complete tables (`d14e838f0`), including
scheme case folding, default-port dropping, `.localhost` subdomains, and query/fragment/userinfo rejection. Every
accepted entry is additionally asserted idempotent.

The `dc-api-compose` CI phase lists every DC test class explicitly; a class omitted there runs nowhere, and because
skips are not surfaced it would read as green rather than as a gap.

### `transaction_data` on Android needs a type Credential Manager knows

AndroidX's default OpenID4VP matcher switches on the transaction-data `type`, and it recognises exactly two:
`payment_details`, for which it reads `payee_name`/`payment_amount`/`payment_currency`, and
`urn:eudi:sca:payment:1`, for which it reads a nested `payload.payee.name`/`amount_display`/`currency` structure.
For **any other** type it falls through to reading `merchant_name`, whose absence Credential Manager answers with
`Merchant name should not be null` and then emits **no candidate at all**. Ours was
`org.waltid.transaction-data.payment-authorization`, so the test found no candidate. `OpenId4VpRegistry` accepts no
matcher, so the default cannot be replaced and the type the verifier sends is the only lever.

`9d260f9a7` first worked around this by adding a `merchant_name` member to the item. **`05a10f85e` replaces that
with the proper fix**: a `payment_details` profile — with the three field names the matcher actually reads — added to
the verifier2 and wallet-api profile lists, in both the service defaults and the docker-compose configs. The profile
is what makes the two sides agree: the verifier accepts any type at session creation, while the wallet gates on
profile membership (`TransactionDataTypeRegistry.requireKnown`), so a shared profile is exactly the seam.
`urn:eudi:sca:payment:1` is deliberately left out — a flat `fields` list cannot describe its nested shape.

> **⚠️ This config must be deployed before the test can use it.** Verified against a local docker-compose stack
> (freshly built `verifier-api2` and `wallet-api` images, since `stable` predates the profiles endpoint): with these
> profiles served and the item switched to `payment_details` with no `merchant_name`,
> `sharesSdJwtWithTransactionDataThroughCredentialManager` passes on an API 37 Play Store emulator — candidate
> produced, all three fields on the consent surface before Share, and `transaction_data_hashes` matching the
> recomputed hash. But `DemoTestBackend` resolves its item from the **deployed public** profile list at run time, so
> until `wallet.demo.walt.id` and `verifier2.demo.walt.id` serve `payment_details`, the test still sends
> `org.waltid.transaction-data.payment-authorization` plus `merchant_name`. Dropping `merchant_name` is a one-line
> follow-up once deployed.

Nothing in the wallet changed for any of this: validation gates on `type` only, and `TransactionDataDecoding` routes
every non-standard member into `details`, so it reaches consent as a decoded detail alongside the profile fields.

## Verification

All library lanes run on `d14e838f0`; the device lanes on `4f9681115`. `05a10f85e` on top of them is config only
(four `transaction-data-profiles.conf` files), touching no source:

| Lane | Result |
| --- | --- |
| `waltid-openid4vp-wallet:jvmTest` | 94 tests, 0 failures |
| `waltid-openid4vc-wallet:jvmTest` | 106 tests, 0 failures |
| `waltid-mdoc-credentials2:jvmTest` | 38 tests, 0 failures |
| `waltid-openid4vc-wallet-mobile:testAndroidHostTest` | 83 tests, 0 failures |
| `waltid-openid4vc-wallet-mobile:iosSimulatorArm64Test` | 87 tests, 0 failures |
| `waltid-wallet-demo-compose:androidApp:compileDebugAndroidTestKotlin` | pass |
| `checkKotlinAbi` (`waltid-crypto2`, `waltid-openid4vc-wallet-mobile`, `waltid-openid4vc-wallet-persistence-mobile`) | pass |
| `DigitalCredentialSharingE2ETest`, local API 37 Play Store emulator | 4 tests, 0 failures |
| `sharesSdJwtWithTransactionDataThroughCredentialManager` against a local stack serving `05a10f85e`'s profiles, with the item switched to `payment_details` and no `merchant_name` | 1 test, 0 failures |

Two `checkKotlinAbi` failures exist repo-wide and are **pre-existing on `main`**, not from this branch:
`waltid-crypto2` and `waltid-crypto2-signum` baselines were dumped without iOS targets, so they fail when
`-PenableIosBuild=true` is passed. This branch modifies no source in either module.

**iOS 26 `IdentityDocumentServices` on a physically provisioned device remains a manual acceptance requirement**
and is currently **not** met: `org-iso-mdoc` fails on a real iPhone from both a local build and a distributed
TestFlight build. Simulator and compile-time tests cannot prove entitlement, provider-registration or Keychain
access-group behaviour. This is tracked as follow-up work, not resolved here.

## Decisions worth a look

1. **Android Annex C is retained as supported**, with a self-verifying Credential Manager E2E rather than being
   unexposed.
2. **`waltid-crypto2`'s public ABI is widened additively** with `Hpke.validateRecipientPublicKey`. Worth confirming
   that belongs in crypto2 rather than in the wallet.
3. **`androidx.credentials` is pinned to `1.7.0-alpha03`** and `androidx.credentials.registry` to `1.0.0-alpha05` —
   both prereleases with experimental APIs.
4. **New public KMP/Swift API:** `IosIdentityDocumentRegistrationStatus` and `reportRegistrationStatus`.
5. **Source-visible break for Swift callers:** `ReaderTrust.unverified` → `.untrusted(reason:)`, plus the new states.
6. **Public API rename:** `DcApiWallet.validatePlatformOrigin` → `canonicalizePlatformOrigin`. The old name promised
   a check it never performed — the function normalises syntax and does not compare against `expected_origins`,
   which the unsigned path strips.
7. **`05a10f85e`'s profile config needs deploying** to `wallet.demo.walt.id` and `verifier2.demo.walt.id`. Until it
   is, the on-device `transaction_data` test keeps the `merchant_name` workaround; after it, that member comes out.
   The config is additive — the two existing profiles are untouched — so deploying it changes nothing else.
8. **Two tests are marked `DEMONSTRATIVE - remove before merging`** (`DcApiDeploymentComparisonE2ETest`,
   `DeployedIssuerDocumentSignerTest`). They document a deployment-side finding rather than wallet behaviour: the
   certificate `issuer2.demo.walt.id` signs mdocs with is X.509 v1 with no extensions, so it carries neither
   `keyUsage:digitalSignature` nor EKU `1.0.18013.5.1.2`. `verifier2.demo.walt.id` 0.23.0 predates that
   enforcement and accepts it; a verifier that enforces the ISO 18013-5 profile rejects the same presentation.
   Keep or drop is an open call.

## Note on #2062's description

Several claims in it did not match the code: it described signed DC API routing that does not exist, said
`dc_api.jwt` was unimplemented, referenced a `validatePlatformOrigin()` performing an `expected_origins` exact
match with a `DcApiOriginMismatchException` (neither symbol exists; `expected_origins` is stripped, not matched),
said the wallet prefers OpenID4VP over Annex C (it answers the selected alternative), and said the `dc-api-compose`
CI lane had not been exercised on a runner. Most of its "known rough edges" are fixed by this branch. #2062's
description needs the same rewrite as this one; it is not updated by this PR.


JakeFernandes98:
DCAPI over OpenID4VP confirmed to be working with mutlipaz verifier in both plain and encypted variants. Signed requests are rejected as explained in the PR summary

DCAPI over 18013-7 does not work on either local build or properly distributed test flight build on an iphone device. However the same protocol works on my android phone - so as expected probably some extra set up/configuration required for Apple

The prompt we see in-app during the sharing could use some improvements - could we not reuse the layout + components from the plain openid4vp flow?

szijpeter:
Folded into #2062. Every commit on this branch is now contained in `feature/wal-752-dc-api-v2` — the fold was a fast-forward, so no SHA reviewed here was rewritten (`git merge-base --is-ancestor` verified against both pre-fold heads, which are also tagged locally).

#2062 is now the single DC API mobile support PR and its description has been rewritten to merge this one's content. Three changes were made on top of the fold:

- `DcApiDeploymentComparisonE2ETest` and `DeployedIssuerDocumentSignerTest` are dropped from the whole lineage. They asserted nothing a wallet change can fix or break; the deployment finding they recorded is kept as comments at the two places a reader would look for it.
- The two browser-mediated tests moved to #2064 (`feature/wal-752-dc-api-browser-tests`), where they run un-`@Ignore`d against their blockers. #2062 keeps only tests expected to pass today.
- The `merchant_name` workaround now carries a `TODO` naming the profile deployment as its removal condition, per review.

Verified on the folded head: all four `DigitalCredentialSharingE2ETest` device tests pass (0 skipped) on an API 37 Play Store emulator, and the library lanes are unchanged at 94/106/38/83/87 plus `waltid-x509` at 131. Closing in favour of #2062.

PR #2067, opened 2026-08-09
[WAL-752] feat(dc-api-ios): make both iOS demos real Apple mobile-document providers

szijpeter:
Part of the **WAL-752** stack. Targets `feature/wal-752-dc-api-v2` (#2062) rather than `main`, so **review that one
first** — this diff is only the iOS platform work on top of it. The Android sharing UX PR (#2066) has since been
folded into #2062, so this branch now sits directly on it.

The two workstreams were built in parallel from the same Phase 0 base and touch mostly disjoint files; the stacking is
a review order, not a code dependency of iOS on the Android UI.

## Problem

The Annex C core existed but the iOS platform integration was not production-correct, and it had never completed a
presentment on hardware. Concretely:

- Apple registrations collapsed all credentials of one doctype into a single entry, so the platform could offer only
  one mDL no matter how many the wallet held.
- Desired registration state was written **only** when Apple had already authorized the provider. Authorizing later
  therefore left Apple's registry empty until the credentials were reissued.
- The Compose demo had no cross-process wallet configuration and no provider extension at all, so it could not act as
  a document provider.
- Nothing verified that host app and extension actually resolved to the same App Group, the same Keychain access
  group, or the same doctype set — the three things that decide whether iOS treats them as one wallet.

## End goal

Make both iOS demos real Apple mobile-document providers, sharing one provider implementation rather than two that
drift. Per the directive, the **Compose demo was prioritized**; the native demo follows the same code path.

## How

**Per-credential registrations.** `IosIdentityDocumentRegistry` projects one `MobileDocumentRegistration` per mdoc
credential, keyed by registry entry ID, and writes the desired projection **regardless of authorization status** so
`performRegistrationUpdates()` can rebuild Apple's registry after the user authorizes without reissuance. The
projection is a blind overwrite of one key in the App Group `NSUserDefaults`, published by the host app only.

**The projection names one wallet, not a union.** Apple hands the extension only the parsed request, the origin,
`sendResponse` and `cancel` — no matched `documentIdentifier` — so an extension cannot tell which wallet a request was
registered for. `IosIdentityDocumentProjectionState` therefore carries exactly one `walletId` and its registrations,
and each publish replaces it. It carries the wallet id because `wallet_${walletId}` decides which database the
extension opens; without it, a host started on a non-default `WALLET_ID` was unpresentable, since the extension opened
an empty `wallet_default` and showed an empty picker instead of an error. Each registration carries only what Apple's
store takes — `documentIdentifier` and `documentType`. A credential id or registry id here would be shared
cross-process metadata with no reader: reconciliation never resolves a registration back to a credential, and Apple
does not hand `documentIdentifier` to the extension either.

**Absent and corrupt state are not "wants nothing".** Reads return `Missing`, `Malformed(reason)` or
`Published(state)`. Collapsing the first two into an empty list was data loss: reconciliation removes every managed
registration absent from the desired set, so an unwritten or corrupt container unregistered documents the wallet could
still present. Missing now leaves Apple's registrations untouched, malformed throws, and only a published projection
authorizes removals — an explicitly empty published projection stays authoritative and does unregister.

**The extension writes nothing.** `providerWallet()` no longer bootstraps: bootstrap creates key material and a DID
and republishes the desired projection, which is the least-entitled process making writes to shared state. Annex C
needs only credential reads and signing-key resolution, which the shared configuration alone provides. The demo tests
assert on a wallet that was never bootstrapped in-process, so they are also the test that no bootstrap is needed. No
filesystem lock is involved anywhere.

**Shared-container configuration is mandatory, not best-effort.** Both demos embed a provider extension, so there is
no useful degraded mode: an unresolved Keychain group or empty App Group silently gave the wallet process-local
storage, which looks healthy to the app and is invisible to the extension — the only symptom was a presentation
failing at signing time on a device, long after the cause. Both demos now stop at launch naming the Info.plist key the
build did not expand.

**Deterministic reconciliation.** `reconciliationPlan(desired:existing:supportedDocumentTypes:)` is pure and therefore
testable without an iOS 26 device. It drops a doctype outside the entitlement rather than registering it (Apple
rejects the addition, and one rejection would abort reconciliation for the valid ones); removes every registration
absent from the desired projection; and treats a doctype change under an unchanged identifier as remove+add, because
Apple has no update operation. Removal is unconditional because a published projection is authoritative for the whole
store: this integration is the only writer of Apple's store for this provider, and a projection that could not be read
never gets that far — `desiredRegistrations(from:)` fails closed first.

**A credential change asks the host to reconcile.** Publishing the desired projection is not the same as writing
Apple's `IdentityDocumentProviderRegistrationStore`, which only the app process may write, and
`performRegistrationUpdates()` runs periodically rather than on change. `MobileWallet.syncDigitalCredentialRegistration()`
— the single funnel every credential mutation already ends in — now emits `onDigitalCredentialRegistryChanged`, and the
Compose host turns that into `IdentityDocumentRegistrationCoordinator.reconcile()`. Reconciling on the notification
rather than on the next foreground transition is what makes a freshly issued credential presentable now; previously the
Compose demo reconciled only at launch and on becoming active. Apple API calls stay in the host: the wallet reports a
fact and does not wait. The callback is not suspending, because Kotlin/Native cannot export a suspending function type
to Swift and the host has nothing to report back — the credential change is already committed. A host callback that
throws is contained, so a failing host cannot roll back a committed credential change.

**The Swift boundary exposes values, not the wallet core's objects.** `desiredRegistrations` returns
`DesiredRegistrationProjection` (`.missing` / `.malformed(reason:)` / `.published(walletID:registrations:)`) with
`Equatable, Sendable` `DesiredIdentityDocumentRegistration` elements, mapped from the KMP projection inside
`WalletSdkBridge`. The demo package previously received the Kotlin objects and the persistence shape along with them.

**One shared provider implementation.** A new local Swift package `WalletDemoIdentityDocumentSupport` holds the
namespace definitions, request mapping, presentation orchestration and registration coordinator, and both demos'
extensions consume it. `IdentityDocumentNamespace` gives each demo its own App Group and Keychain suffix, so
installing both does not make them share credentials. The Keychain access group is read from the bundle
(`WALTKeychainAccessGroup`) rather than composed in code — only the build knows the Team ID, and guessing wrong yields
a group the process is not entitled to, which fails at first use rather than at launch.

**Request mapping that fails closed.** `parsedRequest(from:)` throws on alternative document request sets rather than
silently taking the first: choosing is a user decision, and picking one would disclose more than the verifier asked
for in the branch the user never saw. Origins are passed through verbatim — the KMP canonicalizer stays the sole
source, since its output is hashed into the session transcript.

**CI verifies built products, not sources.** `verify-ios-identity-document-provider.sh` builds the extension scheme
first (so a provider that cannot compile reads as an extension failure), then asserts on the built `.xcent` and
`Info.plist`: extension embedded, identical App Group, identical *first* `keychain-access-groups` entry matched by
suffix so no Team ID is hard-coded, the exact expected doctype set in both, and the correct extension point. Source
`.entitlements` cannot carry this check — `$(AppIdentifierPrefix)` and `INFOPLIST_KEY_` resolve at build time and can
resolve to nothing, which does not fail a build; it fails a presentation on a device.

## Reviewer attention

**The Kotlin ABI baseline was already stale before this branch.** `checkKotlinAbi` on
`:waltid-libraries:protocols:waltid-openid4vc-wallet-mobile` fails at this branch's merge base, so part of the
baseline diff here is a pre-existing repair rather than a change this PR introduces. It passes at this head.

**`CrossProcessWalletTests` was renamed, because its name overclaimed.** It built both wallets in one host process, so
it never covered a second process, while its name and comments said it did — which made physical-device acceptance look
redundant. It is now `SharedWalletConfigurationTests` and says what it actually covers: the namespace resolves to this
demo's App Group and Keychain group, a wallet opened the way the extension opens it finds the host's credentials and
signing key, and that key can sign. Extension-process entitlements stay with the built-product verify script and device
acceptance.

**The Compose iOS demo CI job now assembles `WalletCore`.** That demo embeds an extension linking `WalletSDK`, whose
local binary target is the assembled XCFramework, and the job skipped assembling it — so SwiftPM resolution failed
before any test ran (`local binary target 'WalletCore' does not contain a binary artifact`). The native demo job
already assembled it for the same reason. Both iOS demo jobs also got a longer step timeout: that assembly costs about
27 minutes on a cold Konan cache, which pushed the Compose job past its 40-minute limit with its tests still running
and passing, and left the native job 6 minutes of margin. The limits were measuring cache luck rather than test health.

**The provider verification steps run in this PR's CI, deliberately via a branch pin.** A workflow consumes a
composite action at the pinned SHA, so the steps added to `.github/actions/gradle-ios` would have been dead code here
until a follow-up pin bump on `main` — a green run that never executed them. `gradle-macos.yml` is therefore pinned at
this branch's action commit, and the flag is split per demo (`run-native-…` / `run-compose-…`) so each check runs in the
job that already has that project's build warm rather than making both slow jobs build the other demo. Because a rebase
rewrites that commit's SHA, the pin has to be re-pointed whenever this branch is rebased, or the workflow silently
consumes an orphaned action again. **Retarget the pin at the merged SHA when this lands.** `Verify native iOS identity
document provider` and `Verify Compose iOS identity document provider` should be treated as required.

**One presentation test now waits on a deadline instead of a yield count.** `crossProcessAccessConfiguration()` became
mandatory here, so opening a wallet does real Keychain and App Group work; a fixed 1000 `Task.yield()` is a
machine-speed deadline that a loaded runner can miss while the wallet is starting normally, and the spin starves the
work it waits for. It now polls with a sleep against a real deadline, matching `WalletViewModelReceiveTests`.

**Cancel is disabled while a response is in flight.** Only Share was, so a cancellation could tear the scene down while
`sendResponse` was still sealing a response, and the platform would see both a cancellation and a response for one
request.

## Not achievable in this scope

These are reported here rather than silently dropped.

- **B1 (partial) — Apple Developer portal work needs a human.** Registering the Mobile Document Provider capability,
  the two App Groups and the two Keychain groups, regenerating provisioning profiles, and setting `DEVELOPMENT_TEAM`
  for the new Compose targets all require signed-in access to the portal and Xcode. The entitlements, Info.plist keys
  and target wiring are all in place and CI-checked; the account-side registration is not something this branch can
  perform.
- **B24–B26, and the physical half of the acceptance criteria — no iOS 26 hardware available.** There is no physical
  iOS device attached to this environment, so real entitlement, provider-registration and Keychain access-group
  behaviour cannot be exercised. What stands in for it is weaker and worth stating exactly: `SharedWalletConfigurationTests`
  opens the wallet **the way the extension opens it, but in the host test process** — it covers the shared
  configuration and that the resolved signing key can sign, not a second process's entitlements — plus the
  built-product entitlement checks above. The physical Annex C acceptance run remains **outstanding** and must be
  performed before the iOS capability is claimed as verified.
- **B27 — diagnostic playbook.** Documentation-only; nothing to implement, and this repo does not carry that kind of
  runbook.
- **Release/TestFlight-equivalent acceptance.** Same blocker as above: requires provisioning and a device.


szijpeter:
Folded into #2062.

The whole WAL-752 implementation stack now lives on `feature/wal-752-dc-api-v2`, rebuilt as 14 semantic commits with the same final tree. The iOS `IdentityDocumentServices` work from this PR is carried by `feat(dc-api-ios): register and serve identity documents in both iOS demos` and `feat(dc-api-ios): share provider orchestration between the demo extensions`.

Closing here so #2062 is the single implementation PR; the review history on this PR stays attached to it.

PR #2068, opened 2026-08-09
[WAL-752] feat(ios): unify Digital Credentials sharing review UI

szijpeter:
Part of the **WAL-752** stack, and the last of the three. Targets the iOS platform PR
(`feature/wal-752-dc-api-ios-platform`) rather than `main`, because it builds directly on the provider path that PR
introduces and mirrors the interaction semantics the Android sharing review PR establishes. Review both of those first.

Rebased onto the platform PR's current head, which reworked the cross-process projection, the SDK boundary and the
registration reconciliation, so the diff here is this PR's UI and orchestration-test work only.

## Problem

With Apple's provider integration in place, iOS had three consent surfaces for one act of sharing: the native app's
`PresentationReviewView` for normal OpenID4VP, and `BasicAnnexCReviewView` — a deliberately plain `List` of doctype
strings and namespace names — for the Annex C provider in both demos. The provider surface showed neither credential
cards nor per-claim disclosure rows, and its claim rendering had nothing to do with the app's.

## End goal

One SwiftUI sharing-review system used by normal native OpenID4VP, the native Annex C provider, and the Compose demo's
Annex C provider.

## How

A new `WalletDemoSharingUI` library target in the existing `waltid-wallet-demo-shared-ios` package holds the
provider-safe UI: `SharingReviewModel` and its two adapters, the review view and its sections, and the claim/credential
rendering (`Display/`, the card and metadata components, `WalletAccessibilityID`, `WalletColors`) that previously lived
inside the native app target. `WalletDemoIdentityDocumentSupport` depends on it.

`SharingReviewModel` mirrors the Compose `WalletDemoSharingReview` in concepts, not type names, and stays UI adaptation
only — Annex C and normal OpenID4VP are not merged into one protocol-domain hierarchy. As on Android, sections render
only the concepts a transport has: `ReaderTrust.notApplicable` becomes `nil` rather than a section saying "not
applicable", and reader-trust copy describes who the reader is rather than whether the request is safe —
`pendingVerification` is normal Apple two-stage behaviour, `untrusted` is a trust-policy outcome and not a signature
failure.

Per **C4**, normal native OpenID4VP was migrated to the shared review *first* and its tests brought back green before
Annex C was wired into it; that is what proves the component replaced the old review instead of becoming a third copy.
`PresentationReviewView`, `VerifierReviewSections` and `BasicAnnexCReviewView` are all deleted.

Provider orchestration stays out of the UI package (**C10**): it holds no reference to
`ISO18013MobileDocumentRequestContext`, `context.sendResponse`, `RawAnnexCRequest` or `submitAnnexCPresentation`, and
does not import `IdentityDocumentServices`. Its only imports are `Foundation`, `SwiftUI`, `UIKit` and `WalletSDK`.
Actions are transport-aware (**C9**): normal OpenID4VP offers Share/Reject/Cancel, the Apple provider passes
`onReject: nil`, because IdentityDocumentServices has no channel for a protocol-level refusal and inventing one would
be a fake rejection protocol.

## Reviewer attention

**Two `public init`s were added to `WalletSDK/DigitalCredentials.swift`** (`AnnexCPresentationPreview`,
`DigitalCredentialResponse`), +38 lines. Both types were constructible only inside the package, which made provider
orchestration untestable without a real iOS 26 request — something no test can build. This widens a shipped library's
public API additively and is the one change here outside the demo apps.

**The two-stage submission is now covered end to end.** The stub request context previously only counted
`sendResponse` calls and never ran the closure the platform invokes, which left the whole releasing path untested. It
now executes that closure with known raw request data and the stub wallet records stage 2's arguments, so
`AnnexCPresentationModelTests` asserts that the request ID and origin come from stage 1's retained preview rather than
from the raw request, that `DeviceRequest` and `EncryptionInfo` arrive verbatim, that credentials are submitted in
request order, and that Apple receives the decoded sealed bytes rather than the wallet's base64url envelope. The
failure case covers a wallet that cannot seal: the failure surfaces, no bytes reach the platform, and `isSubmitting` is
cleared so the review stays usable.

Annex C `credentialRequirements` is derived from the matched `credentialOptions` rather than from
`parsedRequest.documents`. That is correct only because `MobileWalletAnnexC` already `require`s a matching option per
requested document, so an unsatisfiable request throws during preview and never reaches the UI. If that precondition is
ever relaxed to allow partial matching, this review would enable Share on a request the wallet cannot satisfy.

## Not achievable in this scope

**C12 — physical iOS acceptance was not run.** No physical iOS 26 device is available in this environment, and the
Annex C provider flow cannot be exercised in a simulator: the OS launches the app extension from a real reader
interaction. Everything below that platform boundary is covered by tests against a stand-in request context, and the
built-product entitlement checks from the platform PR still pass for both demos. **The physical acceptance run for both
demos remains outstanding** and should be performed before the iOS Annex C UX is treated as verified on hardware.

**Four pre-existing `MockWalletUITests` failures are not fixed here.** `testWalletReadyStateIsReachable`,
`testOfferSupportedClaims`, the credential-card identifier lookup and the read-only presentation review already fail
on this branch's merge base, so they are not caused by this PR. That XCUITest lane is also not run by CI — there is no
`xcodebuild test` in any workflow — so it was already unguarded. Fixing it is separate work and would expand this PR's
scope. The `iosAppTests` unit lane, which does cover the review model added here, is green.

One deliberate test-scope call: an assertion that Share stays unavailable when a requested document has no matching
credential was rewritten, because that state is unreachable — the KMP layer rejects such a request during preview. The
test now asserts the reachable equivalent (two requested documents, the user deselects one, Share is withdrawn) rather
than modelling a state the wallet core forbids.


szijpeter:
Folded into #2062.

The whole WAL-752 implementation stack now lives on `feature/wal-752-dc-api-v2`, rebuilt as 14 semantic commits with the same final tree. The shared SwiftUI sharing review from this PR is carried by `feat(ios): render one SwiftUI sharing review across the iOS transports`.

Closing here so #2062 is the single implementation PR; the review history on this PR stays attached to it.

PR #2073, opened 2026-08-10
build(dc-api): vendor the Annex C matcher instead of depending on Multipaz

szijpeter:
Follow-up to [#2062](https://github.com/walt-id/waltid-identity/pull/2062), scoped to the Android Annex C matcher dependency.

## Why the dependency was undesirable

`org.multipaz:multipaz-dcapi-android` was declared only to obtain `identitycredentialmatcher.wasm`, the Credential Manager matcher our `org-iso-mdoc` registry needs because AndroidX `credentials-registry` ships no Annex C registry or matcher. No Multipaz API is called, and `isTransitive = false` excluded the runtime its classes reference — so consumers received unreachable classes with unresolvable references, in exchange for one asset.

## What replaces it

The binary is vendored unmodified and the dependency is gone.

| | |
| --- | --- |
| Upstream repository | https://github.com/openwallet-foundation/multipaz |
| Release tag | `0.100.0` (annotated tag object `dae28ac7f7df30650a39258fcce8be813c46d53b`) |
| Commit | `7c0988bee3384d13a0732e0c33336ae0faf3b863` |
| Upstream path | `multipaz-dcapi/src/androidMain/assets/identitycredentialmatcher.wasm` |
| Git blob | `ebbc2f2671623f803c46c1a18d5f76e690c7e385` |
| Size / SHA-256 | 297,404 B / `420385a46bf554b34224c960051a0cd6c4ecff12ca2c3bdb8948a555afd6e0f8` |

Verified byte-identical to the upstream release blob. EUDI's `eudi-lib-android-wallet-core` vendors the same bytes.

## Licensing

`NOTICE-identitycredentialmatcher.txt` ships beside the binary in the published artifact, recording the licenses of the components compiled into it: Multipaz's matcher sources and libcppbor (Apache-2.0), cJSON (MIT), and `base64.h` (0BSD — the vendored copy is upstream `czkz/base64` verbatim plus an attribution comment, verified against that repository). Multipaz publishes no `NOTICE` file at the pinned commit. Since the matcher is no longer a Maven dependency, the generated `THIRD-PARTY-NOTICE.md` reports no longer cover it, so that file is the record.

## Packaging

It stays an ordinary Android asset at a package-qualified path, loaded with `assets.open(...)`. This requires `androidResources { enable = true }` in the module: asset and resource processing is off by default for Android KMP library targets, and without it `androidMain/assets` is not packaged into the AAR at all. The flag is module-scoped — a sibling mobile module's AAR is unchanged. The package-qualified path prevents collision with another library's copy during the application-level asset merge.

## Compatibility caveat

The matcher's credential-database input and emitted identifiers are Multipaz-private conventions rather than a specification, so `encodeAnnexCCredentialDatabase` and `parseMatcherCredentialId` are coupled to this binary. Both are unchanged here. `ANNEX-C-MATCHER.md` documents the contract and the refresh procedure.

Because the matcher is vendored rather than represented as a dependency, ordinary dependency-update tooling will not detect future Multipaz matcher releases. A follow-up should automate detection of newer Multipaz releases and, when the matcher bytes change, open a reviewable update PR. Such updates must not auto-merge, given the compatibility contract above.

## Guards

`AndroidVendoredMatcherTest` pins the SHA-256 read through the same asset API the registry uses, and checks the notice ships with it. `verifyAnnexCMatcherPackaging` (wired into `check`) asserts the published AAR carries both files, since the asset API alone still resolves when the AAR carries no assets.


szijpeter:
Superseded by #2062, which now carries this change as commit b6183be7c — the cherry-pick is content-identical (all six file blobs match this branch exactly).

Closing here so the vendored matcher lands once, together with the rest of the iOS/Android Digital Credentials work it depends on.

PR #2074, opened 2026-08-10
fix(issuer): enforce single-use pre-authorized codes

matei2803:
## Description

Enforces single-use behavior for pre-authorized codes in the legacy Issuer API.

When using a credential offer configured with:

- `authenticationMethod = PRE_AUTHORIZED`
- `standardVersion = DRAFT13`
- no `txCode`

the same credential offer could previously be claimed by multiple wallets.

I reproduced this with two different wallets / DIDs, both successfully receiving a credential from the same original credential offer.

At the Token Endpoint level, the same `pre-authorized_code` could be successfully exchanged multiple times.

This behavior is inconsistent with OID4VCI Draft 13, where the `pre-authorized_code` is defined as short-lived and single-use.

### Previous behavior

```text
same pre-authorized_code

first token request  -> success
second token request -> success
```

### New behavior

```text
same pre-authorized_code

first token request  -> success
second token request -> invalid_grant
```

The implementation stores active pre-authorized grants separately from the issuance session and atomically consumes the grant using the existing `getAndRemove()` persistence operation.

The issuance session itself remains available for the subsequent credential issuance flow.

`txCode` validation is performed before consuming the pre-authorized grant, so an invalid transaction code does not invalidate the offer for the legitimate holder.

Regression tests were added for:

- successful first use of a pre-authorized code;
- rejection of a second use with `invalid_grant`;
- invalid `txCode` not consuming the pre-authorized code;
- successful use with the correct `txCode` after a failed attempt;
- rejection of reuse after successful exchange.

The full Issuer API test suite passes:

```text
./gradlew :waltid-services:waltid-issuer-api:test

BUILD SUCCESSFUL
```

## Type of Change

- [x] bug fix - change which fixes an issue

## Checklist

## Breaking

No breaking changes.

## Summary by CodeRabbit

- **Enhancements**
  - Pre-authorized credential grants are now securely stored and automatically expire after five minutes.
  - Each pre-authorized code can be redeemed only once, preventing reuse.
  - Invalid transaction-code attempts no longer consume the pre-authorized code, allowing a valid retry.
  - Successfully redeemed codes are consumed atomically to prevent duplicate token issuance.

- **Tests**
  - Added coverage for single-use enforcement and transaction-code validation behavior.


matei2803:
@coderabbitai full review

JakeFernandes98:
Hi @matei2803, Thanks for the contribution. I just wanted to note that the issuer service you have fixed is running draft version of OpenID4VCI (Draft 13), which you have noted as part of your configuration.

If you are interested in the new OpenID4VCI 1.0, please use the issuer2 service instead where this is already implemented.

We will review this soon, Thank you!

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/CIProvider.kt
```diff
@@ -937,6 +953,18 @@ open class CIProvider(
...
         }
 
+        if (tokenRequest is TokenRequest.PreAuthorizedCode) {
+            val grant = preAuthorizedCodeGrants.getAndRemove(sessionId)
+
+            if (grant == null) {
+                throw TokenError(
+                    tokenRequest = tokenRequest,
+                    errorCode = TokenErrorCode.invalid_grant,
+                    message = "Pre-authorized code has already been used."
```
JakeFernandes98, line 963: `grant == null` covers more than reuse — the key may also be missing because it was never stored, or because it expired from persistence while the session is still resolvable.

Consider wording closer to issuer2, e.g. `"Pre-authorized code is invalid or has already been used"`, so the error message does not over-claim the failure reason.

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/CIProvider.kt
```diff
@@ -937,6 +953,18 @@ open class CIProvider(
...
+                )
+            }
+        }
```
JakeFernandes98, line 966: Minor edge case worth being aware of: the grant is consumed via `getAndRemove` before the subsequent token generation / `generateProofOfPossessionNonceFor` (and the callback) complete.

If anything after consume fails, the wallet cannot retry with the same pre-authorized code. That is acceptable for single-use semantics, but it means a transient failure after a successful consume becomes a hard failure for that offer. No change strictly required, just calling it out

waltid-services/waltid-issuer-api/src/main/kotlin/id/walt/issuer/issuance/CIProvider.kt
```diff
@@ -800,6 +808,14 @@ open class CIProvider(
             callbackUrl = callbackUrl
         ).also {
             putSession(it.id, it, expiresIn)
+
+            if (issuanceRequests.first().authenticationMethod == AuthenticationMethod.PRE_AUTHORIZED) {
+                preAuthorizedCodeGrants.set(
+                    it.id,
+                    it.id,
+                    expiresIn
+                )
```
JakeFernandes98, line 817: Nit: the stored value is unused. only presence/absence matters for the null check after `getAndRemove`. Storing `it.id` as both key and value is a bit confusing; a sentinel such as `"1"` (or similar) would make the intent clearer.

waltid-services/waltid-issuer-api/src/test/kotlin/id/walt/OidcIssuanceTest.kt
```diff
@@ -43,6 +50,123 @@ class OidcIssuanceTest {
...
+        val error = assertFailsWith<TokenError> {
+            ciTestProvider.processTokenRequest(tokenRequest)
+        }
+
+        assertEquals(
+            TokenErrorCode.invalid_grant,
+            error.errorCode
+        )
+    }
+
```
JakeFernandes98, line 96: Optional follow-up: it would be useful to add a concurrent double-redeem test (two parallel `processTokenRequest` calls for the same code, asserting only one succeeds).

Not blocking. Redis `GETDEL` / in-memory `getAndRemove` should already make this safe, but it would lock in the race behavior the same way issuer2 does.

JakeFernandes98 (commented):
Thanks for this fix — the approach looks solid overall (separate grant store, `txCode` validation before consume, atomic `getAndRemove`). A few small suggestions below.

JakeFernandes98:
Separate from the review comments above I have two related suggestions that are out of scope for this PR, but may be useful depending on your use case:

1. **`txCode` comparison is plaintext today.** issuer2 hashes the transaction code before storage/comparison. If you need stronger protection of PIN values at rest (or against accidental logging), hashing here would be a natural follow-up.

2. **Authorization-code grants are still not single-use via this new store.** This PR correctly scopes the fix to pre-authorized codes. If you also need one-time redemption for the authorization-code grant path, that would be a separate change.

matei2803:
Thanks for the review. I updated the invalid grant message to cover missing/expired/consumed grants, replaced the duplicated session ID value with a sentinel value, and added a concurrent double-redemption test verifying that exactly one request succeeds while the other receives invalid_grant.

I kept the consume-before-token-generation behavior as discussed, and left txCode hashing and authorization-code single-use handling out of scope for this PR.

The targeted tests and the full waltid-issuer-api test suite pass.

JakeFernandes98 (approved):

PR #2075, opened 2026-08-10
update issuer certs

JakeFernandes98:
ES: https://github.com/walt-id/waltid-identity-enterprise/pull/594

As Peter pointed out, our default certificates were not compliant to the DSC profile. These have been updated across the stack now and should reflect in deployed environments once released


szijpeter:
Verified the new certs locally (no changes made). The DS cert is a clear improvement: the one it replaces carried **no X509v3 extensions at all**, while the new one has critical `KeyUsage: Digital Signature`, critical `EKU 1.0.18013.5.1.2`, SKI, AKI, `issuerAltName`, and a CRL DP, with a 457-day validity. `openssl verify -CAfile <new root> <new DS>` returns OK, and the DS public key matches `defaultIssuerKey` in `issuer2-profiles.conf`, so signing works.

Two observations:

**1. The root and the DS share one key pair, and its private key is committed.** The new root and the new DS have byte-identical SubjectPublicKeyInfo (SHA-256 `9135426955...3337256e` for both), and reconstructing a key from the `defaultIssuerKey` JWK in `issuer2-profiles.conf` — private `d` included — yields that same SPKI. So while this is structurally a two-tier chain that verifies, cryptographically there is one key: the DS key *is* the root CA key, and the root CA private key is in the repo. `AKI == SKI` on the DS is a symptom of that rather than a coincidence. For demo/test material this may be an accepted trade-off, but a reader that expects the IACA to be a distinct trust anchor may object, and anything meant to look production-shaped would want a separate root key.

**2. `https://crl.walt.id/crl.der` returns 404.** The DS cert points there via CRL DP (checked three times just now; the host also serves a TLS chain that doesn't validate against the default CA store). Nothing to fix in the cert content — just that a reader doing revocation checking gets nothing usable today. Likely a deployment follow-up rather than something for this PR.

For reference, `defaultHaipIssuerX5chain` is untouched here and stays on the separate HAIP conformance root, which reads intentional.

szijpeter (approved):

PR #2079, opened 2026-08-11
Review: SCA payment transaction data type, nested labels, and mdoc key authorizations

szijpeter:
Follow-up to [#2076](https://github.com/walt-id/waltid-identity/pull/2076), addressing its review and completing the SCA payment demo. Targets its head, so the diff shows only what this review changes; meant to be folded into #2076 rather than merged separately.

## Summary

Makes the mdoc SCA payment demo actually presentable: the transaction data type is the one implementations recognise, issuance authorizes the holder to sign it, the wallet review says what is being authorized, and the request examples say which of them Android can run.

## Resulting behavior

**Transaction data type.** The demo SCA profile is `urn:eudi:sca:payment:1` with the nested payload that type defines — `payload.transaction_id`, `payload.payee.name`, `payload.payee.id`, `payload.currency` and a numeric `payload.amount`. It replaces the previous `payment_details` profile, whose matcher handling expects a different, flat field shape and therefore does not represent this SCA payment payload. The type matters beyond naming: `urn:eudi:sca:payment:1` has dedicated handling in the matcher Android Credential Manager runs, which reads exactly those nested fields — including `amount` as a JSON number, skipping the entry if it is a string.

**Issuance authorizes signing.** `CredentialProfileConfig.authorizedTransactionDataTypes`, also available as a runtime offer override, reaches the mdoc MSO. Presenting transaction data on an mdoc requires its type to appear in `DeviceKeyInfo.KeyAuthorizations`, so without this the holder cannot sign it at all. Each type is authorized granularly — `dataElements[type] = ["transaction_data_hash", "transaction_data_hash_alg"]`, the elements our `MdocPresenter` device-signs — rather than as a blanket `nameSpaces` grant, which would also permit arbitrary future device-signed elements under that namespace. Blank types are dropped, duplicates collapsed, and a profile with nothing valid gets no `KeyAuthorizations` at all.

Both the element names and the namespace are walt.id's, not the spec's. OpenID4VP 1.0 covers transaction data in §8.4 and its mdoc encoding in Appendix B.2.1, which deliberately defines no element names: each transaction data type defines the `(NameSpace, DataElementIdentifier, DataElementValue)` it contributes, and the issuer authorizes those. Using the type itself as the DeviceSigned namespace is our convention, matching what `MdocPresenter` emits. That is now stated where the conversion happens and where the profile field is declared.

`MdocIssuer` gained only the generic ability to place a `KeyAuthorization` into `DeviceKeyInfo`; `MdocCredentialSigner` owns the issuer-facing conversion; issuer2 only transports profile, session and runtime-override data to it. The old `waltid-issuer-api` is untouched.

**Wallet review.** Nested transaction data leaves inherit the label of the object they came from, so the sharing screen reads "Payee name" rather than a bare "Name". Scoped to transaction data presentation — credential claims keep their plain labels, because their vocabulary already disambiguates them.

**Request examples.** Two, both mdoc:

- `[openid4vp-dc_api][sca] urn:eudi:sca:payment:1` — one SCA payment card credential. Runnable through current Android Credential Manager.
- `[openid4vp-dc_api][sca + age] urn:eudi:sca:payment:1 (not on Android)` — SCA payment card plus age verification, transaction data bound to the payment card alone. Kept as a reference shape for combined presentation; the label and KDoc state the Android limitation up front.

The payment card is a demo credential and is now described as one everywhere, including in the credential issuer metadata a wallet actually reads — which still advertised "Strong Customer Authentication attestation for payment cards per EUDI TS-12 v1.0.1". TS-12 v1.0.1 scopes itself to SD-JWT VC over OpenID4VP and delegates attestation structure to rulebooks, and `eu.europa.ec.eudi.sca.payment_card.1` is not an EU registered doctype, so that was a conformance claim the credential does not meet. The doctype itself is kept: it is what the France Identité playground uses for the same demo.

**Test matrix fix.** #2076 adds `scaPaymentCardMdoc` to the profile config and credential metadata but not to the test scenario matrix, leaving `assertProfilesCoverCredentialMetadata` red on its head. Registered, in its own commit.

## Android Credential Manager limitation

**Android Credential Manager's embedded matcher cannot simultaneously handle this transaction data prompt and a second credential.** Two gates in the OpenID4VP matcher wasm that AndroidX embeds:

- `report_matched_credential` compares a candidate against only the **first** child of `transaction_data[0].credential_ids` — an `i32.load offset=8` on the cJSON array node, with no sibling walk.
- The transaction data rendering path is gated on `cJSON_GetArraySize(transaction_data) == 1`.

Verified on an emulator, dumping the picker before any click:

| Request | Result |
|---|---|
| 1 credential + 1 transaction data entry | works — candidate surfaces with the payment prompt rendered |
| 2 credentials + 1 entry bound to the payment card | zero candidates — "Your info wasn't found", not a partial list |
| 2 credentials + 1 entry per credential | both credentials surface, no transaction data rendered at all |
| 2 credentials, no transaction data | works — this is the France Identité Wine scenario |

So the two gates make *this prompt* and a second credential mutually exclusive; combined presentation as such is fine, and listing both credential ids in one entry does not get around it.

`urn:eudi:sca:payment:1` has dedicated handling for the nested TS-12 payload this demo sends, and the same nested transaction works with one credential and fails once a second credential query is introduced — so this is not a payload-shape problem.

It is also not a format problem. In the matcher, `format`, `mso_mdoc`, `dc+sd-jwt`, `vct_values` and `doctype_value` are each referenced exactly once, all inside `MatchCredential`; `transaction_data` and `credential_ids` once each, in `__original_main`. The two paths never meet, so transaction data handling never learns a candidate's format. The failure is independent of credential format. We temporarily tested an SD-JWT VC SCA credential specifically to verify that hypothesis; it failed identically, confirming that the limitation is in AndroidX's transaction data candidate matching rather than the mdoc path. The experimental SD-JWT additions were therefore removed from the final PR.

Substituting the Multipaz matcher we already vendor for Annex C does not help either. It does implement OpenID4VP and DCQL for both formats — more than its Annex C role requires — but it implements no `transaction_data` at all and imports none of the payment host functions, so it would drop the payment prompt entirely: the split-entry outcome by a more invasive route. `ANNEX-C-MATCHER.md` now records that, so the next person to reach for it knows before trying.

`OpenId4VpDefaults` is `internal` and absent from `api/current.txt`, so the embedded matcher is not replaceable from wallet code; alpha05 is the newest release and androidx-main HEAD ships a byte-identical binary. Nothing below the platform is affected: browsers on other platforms and the wallet's own review screen build the prompt from the request itself.

## Tests

- `MdocCredentialSignerKeyAuthorizationsTest` — the signer's conversion: granular data elements and no blanket namespace, per-type authorization, blanks and duplicates dropped, no `KeyAuthorizations` when nothing valid remains.
- `Issuer2TransactionDataAuthorizationTest` — the same grant observed through issuer2's HTTP wallet flow by decoding the issued MSO: from the configured profile, from a runtime override replacing it, and absent for a profile that authorizes nothing.
- `CredentialDisplayNormalizerTest` — nested transaction data labels, including that credential claim rendering is unchanged.
- `DigitalCredentialSharingE2ETest.sharesMdocWithScaPaymentTransactionDataThroughCredentialManager` — the full round trip, asserting the payment details on Credential Manager's own prompt (the only place the matcher's output is observable), the qualified labels on the wallet review, and `mso_mdoc/transaction-data-hash-check` on the verifier. `@Ignore`d until the profiles in this branch are deployed to the demo issuer and verifier; the `@Ignore` reason says what to deploy. Everything above it runs unconditionally.

## Known and deferred

- SCA transaction data alongside a second credential on Android, per the section above. Not a wallet-core limitation and not fixable from wallet code.
- A combined request without transaction data — the France Identité Wine shape — works on Android and would make a good interoperability example, but is left out of this cleanup.
- The demo portal cannot author a nested `payload`: `VerificationSection.tsx` renders flat inputs for transaction data fields, so the new profile's single `payload` field has no usable editor. That is [#1899](https://github.com/walt-id/waltid-identity/pull/1899)'s territory.
- Full TS-12 SD-JWT VC SCA Attestation support is a separate feature: it needs VC Type Metadata `category`/`transaction_data_types`, metadata-driven schema and rendering, and SCA-specific KB-JWT semantics such as `amr`.
- Already tracked elsewhere: mdoc DS certificate conformance in [#2075](https://github.com/walt-id/waltid-identity/pull/2075), the deployed payment-authorization profile field skew in [#2062](https://github.com/walt-id/waltid-identity/pull/2062).


szijpeter:
Folded into #2076.

PR #2082, opened 2026-08-12
Register Google's newer OpenID4VP matcher for Android Credential Manager

szijpeter:
Targets [#2079](https://github.com/walt-id/waltid-identity/pull/2079)'s head, so the diff shows only the matcher change.

A valid OpenID4VP request for two DCQL credentials with one `transaction_data` entry bound to one of them yields *zero* candidates on Android — the picker shows "Your info wasn't found". This registers a matcher that handles it, and nothing else changes.

## Why

Once `transaction_data` is present, the matcher AndroidX embeds takes a payment-only emit path whose sole exits are `AddPaymentEntryToSetV2` / `AddPaymentEntryToSet` / `AddPaymentEntry`. The ordinary `AddStringIdEntry` / `AddEntryToSet` path lives in a sibling loop reached only when there is no transaction data, with no fallthrough, so a credential the transaction data does not name emits nothing at all — while the option was already declared with `AddEntrySet(setId, 2)`. Declared two, delivered one, so the platform discards the whole option. That is why the failure is total rather than degrading to the payment credential alone. The fault is specific to the transaction data reporting path, not to generic multi-credential DCQL matching.

Google's newer Rust matcher decides per credential: it iterates all of `credential_ids`, reports transaction-bound credentials as payment entries, and reports the remaining matched credentials as standard verification entries through an explicit `if !reported` fallback.

Not fixable by upgrading — `1.0.0-alpha05` is the newest published `registry-digitalcredentials-openid` and `androidx-main` HEAD embeds a byte-identical matcher — and not fixable through `OpenId4VpRegistry`, whose `OpenId4VpDefaults` is `internal` and absent from `api/current.txt`. Nor by reusing the Annex C matcher already vendored here, which implements no `transaction_data` at all.

## What changes

`AndroidDigitalCredentialRegistry` builds the same `OpenId4VpRegistry` as before and re-registers its `credentials` bytes verbatim through a private `DigitalCredentialRegistry` subclass carrying the vendored matcher. AndroidX remains the sole serializer of the registry format; only the matcher binary is ours. The Annex C registry is untouched, and the two share no abstraction beyond both being raw registries.

Registering only `openid4vp-v1-unsigned` keeps the wallet out of signed and multisigned requests it cannot fulfill: the matcher iterates `registry.supported_protocols` and skips anything unlisted. This is load-bearing rather than incidental — the matcher recognises all three `openid4vp-v1-*` protocols — so it is asserted against the real platform rather than assumed.

The `[openid4vp-dc_api][sca + age]` example is no longer labelled "not on Android", and its KDoc now describes the limitation as belonging to the matcher AndroidX currently embeds rather than to the platform.

## Provenance

`openid4vpmatcher.wasm` is **our build, compiled from unmodified sources** from [android/identity-samples@5d966fc](https://github.com/android/identity-samples/commit/5d966fc4913cac93f3b3b11e11bdd44d3e0b5c9e) (branch `wasm`, `CredentialProvider/wasm/matcher-rs`, Apache-2.0) using upstream's own `build.sh`. Upstream ships the Rust sources only, so this is not a Google-published artifact and there is no official binary for this commit to verify it against.

The build is not bit-reproducible across toolchains — `build.sh` uses `cargo +nightly -Z build-std`, so the bytes depend on the nightly — so `OPENID4VP-MATCHER.md` and the shipped notice record the exact commit, source path, toolchain, size and SHA-256, and `AndroidVendoredMatcherTest` pins that hash. The documented maintenance model is: the pinned hash verifies the committed binary; an intentional update picks an upstream commit, records the toolchain it was built with, runs upstream's test suite and our Android matcher regressions, then updates asset, hash and provenance together.

## Tests

- `DigitalCredentialSharingE2ETest.sharesMdocWithScaPaymentTransactionDataAndSecondCredential` — the shape this exists for. Asserts that Credential Manager surfaces the combined option, that the payment payee and amount are visible on its own prompt, that the second credential is present in that same option, that the wallet review receives both, that both DCQL query ids appear in `vp_token`, that `mso_mdoc/transaction-data-hash-check` executes and passes, and that the verifier independently reports the age credential.
- `doesNotSurfaceForSignedOrMultisignedRequests` — the gate for retaining the matcher, and active in committed E2E. Each unsupported request carries the *same* payload the unsigned baseline matches, re-wrapped into the signed and multisigned shapes the matcher's own parsers accept, so it isolates `supported_protocols` rather than payload parsing: were the gate to break, the matcher would decode the payload, run the same mDL DCQL query and surface the credential. Verified by mutating `supportedProtocols` to advertise both, which does fail the test. Asserted through the caller, because "no provider" is what the caller observes — when nothing can answer, Credential Manager puts up its own empty state and only fails `getCredential` once dismissed, so the test dismisses it.
- `sharesTheSatisfiableCredentialSetOption` — `credential_sets`, a materially separate matcher code path where parity cannot be assumed: the unsatisfiable option must not be offered, and the satisfiable one must answer exactly its own query.
- `AndroidVendoredMatcherTest` — pins the binary's hash and the notice's provenance and license text, through the same asset API the registry uses.

`sharesMdocWithScaPaymentTransactionDataAndSecondCredential` was executed successfully against a local issuer2/verifier2 stack built from this PR lineage, on an emulator with real Google Play services. The committed test remains `@Ignore` only because the required SCA and age-verification profiles are not yet deployed to the public demo stack, so CI does not run it today; each `@Ignore` reason names what to deploy.

## Known behaviour differences

No full parity audit against AndroidX's matcher was performed, so this is not a closed list. Identified during this review:

- `evaluate_explicit_credential_sets` filters on `required.unwrap_or(true)`, so `required: false` credential sets are never evaluated, where AndroidX's matcher reports optional sets. Nothing in this repo sends optional sets and no test covers them.
- The matcher imports its set functions from module `credman`, where AndroidX's imports them from `credman_v2`/`credman_v5`. Play services resolves both today; a future release that stopped aliasing `credman` would break this binary and not AndroidX's.

## Removal condition

Delete the asset, the notice, the registry subclass and its branch once a released `registry-digitalcredentials-openid` embeds a matcher that reports a payment entry and a standard entry from one request — test by dropping the custom registry and re-running `sharesMdocWithScaPaymentTransactionDataAndSecondCredential`. Because the matcher is vendored rather than a declared dependency, dependency-update tooling will not flag upstream changes to it.


szijpeter:
Folded into #2076.

PR #2086, opened 2026-08-13
[WAL-752] test: stabilize Android Digital Credentials E2Es

szijpeter:
## Summary

Stabilize the Android DC API sharing E2Es without weakening their end-to-end coverage, and make the dedicated API 37 Play Store test environment reproducible.

## What changed

- Live-issue the canonical mDL and SD-JWT once per suite through `MobileWallet`, then reuse and verify that shared state across presentation tests.
- Replace the global verifier handoff with request-scoped handles and lifecycle-bound platform calls, with explicit cleanup for completion, cancellation, back/re-entry, and Activity destruction.
- Make picker interaction state-driven and fail with focused diagnostics instead of relying on stale UI state or cross-test cleanup.
- Add a real Credential Manager/GMS platform smoke gate and require instrumentation results, so a setup failure cannot be reported as a passing zero-test lane.
- Cold-boot and cache a validated API 37 / Pixel 7 / Play Store AVD. The lane fingerprints the emulator, platform-tools, system image, AVD config, device build, page size, GMS version/code, and preparation scripts, and rejects stale or mismatched cache state.
- Exclude the class-level GMS-gated sharing suite from the default API 34 `compose-demo` lane and run it in the dedicated Play Store lane.
- Run the now-enabled SCA + EU age presentation case in the stabilized DC API lane.

## Caveats and follow-ups

- The dedicated lane remains tied to the current API 37 Play Store / x86_64 emulator substrate and its pinned package revisions; a platform upgrade should change the cache identity and baseline together.

## Breaking

None.


szijpeter:
Reopening immediately to trigger restored CI on the cleaned main-based SHA; the PR remains draft.

PR #2077, opened 2026-08-11
Android Credential Manager adapter for OpenID4VCI CREATE_CREDENTIAL

JakeFernandes98:
## Summary
- Adds an Android Credential Manager **provider adapter** for Digital Credentials API issuance (`CREATE_CREDENTIAL` / `openid4vci-v1`), feeding offers into the existing platform-neutral OpenID4VCI issuance engine (`MobileWallet` / session persistence).
- Compose demo fulfills create requests in a translucent Activity + bottom sheet (offer review, pre-authorized + tx-code). Authorization-code grants reuse the shared **external browser** + `openid://` callback path (no embedded WebView); durable handoff reconnects the persisted issuance session.
- Typed `MobileWalletCredentialOffer` (`Uri` | `InlineJson`) replaces nullable offer XOR fields. Creation-option registration is separated from presentation registry `replace()`. Provider branding uses the host app label/icon.
- Vendored CMWallet provision matcher provenance/docs corrected (C `matcher/issuance/provision.c`, immutable commit pin). Redistribution remains a **release gate** until upstream licensing is explicit.

### Platform scope
OpenID4VCI issuance itself is platform-neutral. What this PR adds on Android is the **native Credential Manager holder/provider integration**. Apple’s public Identity Document provider APIs currently cover presentment, not an equivalent create/issuance handoff—so iOS issuance-provider wiring is deferred, not missing from the common engine.

### HCE / default wallet app (demo-only)
The stub NFC Host Card Emulation / payment-category AID is **intentionally kept for the demo** so Android 15+ can list the app under **Default apps → Wallet app** (`ROLE_WALLET`). It declines APDUs and is **not** a payments or contactless-credential implementation; it is unrelated to Credential Manager Digital Credentials flows. Treat it as demo discoverability only.

### Interop notes
| Issuer / wallet | walt.id | CM Wallet |
| --- | --- | --- |
| walt.id Portal 2 | ✅ | ✅ |
| Google DC API portal | ❌ (investigating) | ✅ |

Google portal incompatibility is still under investigation (offer/credential shape differences observed vs CM Wallet). This PR does **not** claim the Google credential is malformed.

## Test plan
- [x] Host unit tests: create-provider parse/`openid4vci-v1`-only, registry encoding, matcher hash + provision contract, AuthHandoff delivery/orphan queue
- [x] Common `MobileWallet` inline-offer issuance tests
[...]

PR #2089, opened 2026-08-13
[WAL-820] feat(mobile): Replace CMWallet matcher with Google OpenID4VCI issuance matcher

szijpeter:
## Summary

This is a strictly scoped follow-up to [walt-id/waltid-identity#2077](https://github.com/walt-id/waltid-identity/pull/2077). It replaces the vendored CMWallet OpenID4VCI matcher with Google's Rust issuance matcher so Android Credential Manager can surface the wallet for valid OpenID4VCI issuance requests.

The PR is stacked on [walt-id/waltid-identity#2094](https://github.com/walt-id/waltid-identity/pull/2094). The platform-neutral wallet engine, issuer flow, provider API, and iOS/KMP paths remain outside this change.

## What changed

- Replaced the old provision matcher asset with a reproducible `issuance.wasm` built from immutable `android/identity-samples` commit `d5a8adc1b84061a4e3a9581cdaf867df89fb1f19`.
- Kept the AndroidX registration contract: packed icon/JSON layout, canonical `openid4vci-v1`, issuance entries, `package_info`, and the broad `Pass` filter.
- Applied one local Rust parser change: remove the unused string-typed `credential_signing_alg_values_supported` field. Valid mdoc metadata carries numeric COSE identifiers such as `[-7, -9]`, and the matcher does not use this field for matching.
- Added the source-level Rust regression and retained one authoritative AndroidX serialization test plus one vendored-asset/provenance test. Removed the duplicate issuance registration test.
- Made the Portal-shaped Android fixture keep embedded metadata issuer identities consistent with the generated inline offer.
- Documented the pinned source, local patch, toolchain, licensing notice, binary size, and hash. No public SDK API or platform-neutral wallet behavior changes.

## Validation

- Red-green proof against pristine pinned upstream: the regression fails with `Unexpected token I64(-7) expected string`; after the one-line parser fix it passes.
- Clean full Rust suite: 84 tests passed. Clean release build reproduced the committed Wasm exactly: 91,976 bytes, SHA-256 `3a745c778f1881365cced1f4215b5e7e80f673928052f9d3883eb3d596d42c95`.
- `:waltid-libraries:protocols:waltid-openid4vc-wallet-mobile:testAndroidHostTest` passed, including registry serialization, vendored hash, NOTICE/provenance, source, and license assertions. `:waltid-libraries:protocols:waltid-openid4vc-wallet-mobile:verifyMatcherPackaging` passed, confirming the vendored matcher binaries and notices are present in the Android AAR.
- On emulator `emulator-5554` (Pixel 9, API 35), all three `DigitalCredentialIssuanceE2ETest` cases passed, including `acceptsPortalShapedOfferThroughCreateCredential`.

## Provenance / local patch

The checked-in binary is built from Google's `CredentialProvider/wasm/matcher-rs/src/bin/issuance.rs` at commit `d5a8adc1b84061a4e3a9581cdaf867df89fb1f19` using `rustc 1.99.0-nightly (12c36e253 2026-08-10)` via the `nightly-2026-08-11` channel for `wasm32-unknown-unknown`. The only source deviation is the unused numeric-mdoc metadata field removal, with its regression test kept in `matcher-patches/`. `NOTICE-issuance.txt` records the source and compiled-component licenses; its dependency list is a build-graph record, not a claim that every dependency's bytes are present in the final Wasm.


JakeFernandes98 (approved):

.github/workflows/android-device-tests.yml
```diff
@@ -20,9 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # dc-api-compose is disabled: Play Store AVD GMS prep is flaky in CI
```
JakeFernandes98, line 23: should we redisable this before merging to main?

szijpeter: nice catch, thx! 🎣 
dropped this change

PR #2115, opened 2026-08-17
Fix did:jwk issuer kid qualification and key resolution

JakeFernandes98:
## Summary

Issuer2 and other `getKidHeader` callers now emit the `did:jwk` verification-method id (`{did}#0`) instead of a stored key id or JWK thumbprint fragment. Verifier signature checks were failing with `Could not resolve issuer key` because `did:jwk` documents only publish `#0`, so a non-method `kid` never matched the resolved key.

This unblocks SD-JWT VC (and other JWT) verification for credentials whose issuer DID is `did:jwk`. `did:key` was already rewritten to its method verification-method id and is unchanged. Tracked as WAL-1307.

Related PRs:
- [walt-id/waltid-identity-enterprise#612](https://github.com/walt-id/waltid-identity-enterprise/pull/612) (WAL-1307-related, independent of these changes)
- [walt-id/waltid-identity-enterprise-license#4](https://github.com/walt-id/waltid-identity-enterprise-license/pull/4)

### Issuer kid construction
- `did:jwk` is treated as a self-identifying DID method: `kid` is always `{did}#0`, including when the signing key id is already a DID URL or starts with `#`.
- Thumbprint substitution is skipped for `did:jwk`, matching the existing `did:key` exception. `did:web` still uses a thumbprint or explicit verification-method id.

### Key resolution
- `did:jwk` resolution requires exactly one verification key and `kid` `{did}#0`.
- A KMS-path, thumbprint, or DID-without-fragment `kid` does not resolve. Omitted `kid` still selects the single key.
- Multi-key methods such as `did:web` still require a matching verification-method id.

### Tests
- Kid qualification covers KMS-style key ids, DID-URL key ids, `{did}#0`, and `#0`.
- Round-trip signature verification succeeds for `{did}#0` and fails for a non-method fragment.
- Resolver tests cover `#0` success, omitted `kid` success, and rejection of DID-as-kid and KMS-path fragments.

## Architecture Notes

- `did:jwk` encodes exactly one public key in the DID and always exposes verification method `#0`. A non-`#0` fragment cannot select a different key.
- Issuance always writes `{did}#0`. Resolution is strict; there is no unmatched-kid fallback.
- OpenID4VCI SD-JWT signing already goes through `Issuer.getKidHeader`, so Issuer2 inherits this without an enterprise-only change.

## Caveats and Follow-Ups

- Credentials already issued with a non-`#0` `did:jwk` kid will fail signature-key resolution after this change and need re-issuance.
- No Verifier2 trust-list, JWKS, or DID-store configuration is required for `did:jwk`.

## Breaking

- Newly issued `did:jwk` JWTs use `kid` `{did}#0` instead of `{did}#{keyId}` or `{did}#{thumbprint}`. Relying parties that required a non-`#0` fragment for `did:jwk` will need to accept `#0`.
- Verification now rejects `did:jwk` JWTs whose `kid` is not `{did}#0` (except when `kid` is omitted and the document has a single key).

### DIFF
--- a/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonMain/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolver.kt
+++ b/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonMain/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolver.kt
@@ -125,7 +125,11 @@ class Crypto2JwtKeyResolver(
         val kidCandidates = idCandidates(kid)
         val matches = keys.filter { key -> idCandidates(key.id.value).any(kidCandidates::contains) }
         require(matches.size <= 1) { "Multiple DID verification keys match kid: $kid" }
-        return matches.singleOrNull() ?: throw NoSuchElementException("No DID verification key matches kid: $kid")
+        matches.singleOrNull()?.let { return it }
+        if (keys.size == 1 && did.startsWith("did:jwk:")) {
+            return keys.single()
+        }
+        throw NoSuchElementException("No DID verification key matches kid: $kid")
     }
 
     private fun idCandidates(id: String): Set<String> =
--- a/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonMain/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolver.kt
+++ b/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonMain/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolver.kt
@@ -119,17 +119,19 @@ class Crypto2JwtKeyResolver(
             require(keys.size == 1) { "JWT with multiple DID verification keys must include kid" }
             return keys.single()
         }
-        if (keys.size == 1 && kid == did && (did.startsWith("did:key:") || did.startsWith("did:jwk:"))) {
+        if (did.startsWith("did:jwk:")) {
+            require(keys.size == 1) { "did:jwk must resolve to exactly one verification key" }
+            require(kid == "$did#0") { "did:jwk kid must be '$did#0', but was '$kid'" }
+            return keys.single()
+        }
+        if (keys.size == 1 && kid == did && did.startsWith("did:key:")) {
             return keys.single()
         }
         val kidCandidates = idCandidates(kid)
         val matches = keys.filter { key -> idCandidates(key.id.value).any(kidCandidates::contains) }
         require(matches.size <= 1) { "Multiple DID verification keys match kid: $kid" }
-        matches.singleOrNull()?.let { return it }
-        if (keys.size == 1 && did.startsWith("did:jwk:")) {
-            return keys.single()
-        }
-        throw NoSuchElementException("No DID verification key matches kid: $kid")
+        return matches.singleOrNull()
+            ?: throw NoSuchElementException("No DID verification key matches kid: $kid")
     }
 
     private fun idCandidates(id: String): Set<String> =
--- a/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonTest/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolverTest.kt
+++ b/waltid-libraries/credentials/waltid-credential-key-resolver/src/commonTest/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolverTest.kt
@@ -83,23 +83,23 @@ class Crypto2JwtKeyResolverTest {
     }
 
     @Test
-    fun `did jwk key is selected when kid is the DID`() = runTest {
-        val resolved = resolver("$DID_JWK#0").resolveFromJwt(
-            jwtHeader = buildJsonObject { put("kid", DID_JWK) },
-            jwtPayload = buildJsonObject { put("iss", DID_JWK) },
+    fun `did jwk resolution rejects kid that is the DID without fragment`() = runTest {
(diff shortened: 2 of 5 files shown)


waltid-libraries/credentials/waltid-w3c-credentials/src/commonMain/kotlin/id/walt/w3c/issuance/Issuer.kt
```diff
@@ -349,6 +351,7 @@ object Issuer {
         DidUtils.isDidUrl(keyId) -> keyId
         keyId.startsWith("#") -> issuerDid + keyId
         issuerDid.startsWith("did:key:") -> "$issuerDid#${issuerDid.removePrefix("did:key:")}"
+        issuerDid.startsWith("did:jwk:") -> "$issuerDid#0"
```
chsavvaidis: Move the did:jwk condition at line 354 above the generic key-ID conditions (Perhaps it's better to force did:jwk to #0 at the moment.):

  private fun qualifyDidKeyId(keyId: String, issuerDid: String?): String = when {
      issuerDid.isNullOrEmpty() -> keyId
      issuerDid.startsWith("did:jwk:") -> "$issuerDid#0"
      DidUtils.isDidUrl(keyId) -> keyId
      keyId.startsWith("#") -> issuerDid + keyId
      issuerDid.startsWith("did:key:") -> "$issuerDid#${issuerDid.removePrefix("did:key:")}"
      else -> "$issuerDid#$keyId"
  }

waltid-libraries/credentials/waltid-credential-key-resolver/src/commonMain/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolver.kt
```diff
@@ -125,7 +125,11 @@ class Crypto2JwtKeyResolver(
         val kidCandidates = idCandidates(kid)
         val matches = keys.filter { key -> idCandidates(key.id.value).any(kidCandidates::contains) }
         require(matches.size <= 1) { "Multiple DID verification keys match kid: $kid" }
-        return matches.singleOrNull() ?: throw NoSuchElementException("No DID verification key matches kid: $kid")
+        matches.singleOrNull()?.let { return it }
```
chsavvaidis: Perhaps you could clean the selectDidKey() like:
```

  private fun selectDidKey(keys: Set<Key>, did: String, kid: String?): Key {
      require(keys.isNotEmpty()) {
          "No verification keys resolved for DID: $did"
      }

      if (kid == null) {
          require(keys.size == 1) {
              "JWT with multiple DID verification keys must include kid"
          }
          return keys.single()
      }

      if (did.startsWith("did:jwk:")) {
          require(keys.size == 1) {
              "did:jwk must resolve to exactly one verification key"
          }
          require(kid == "$did#0") {
              "did:jwk kid must be '$did#0', but was '$kid'"
          }
          return keys.single()
      }

      if (keys.size == 1 && kid == did && did.startsWith("did:key:")) {
          return keys.single()
      }

      val kidCandidates = idCandidates(kid)
      val matches = keys.filter { key ->
          idCandidates(key.id.value).any(kidCandidates::contains)
      }

      require(matches.size <= 1) {
          "Multiple DID verification keys match kid: $kid"
      }

      return matches.singleOrNull()
          ?: throw NoSuchElementException(
              "No DID verification key matches kid: $kid"
          )
  }
```

waltid-libraries/credentials/waltid-credential-key-resolver/src/commonTest/kotlin/id/walt/credentials/keyresolver/Crypto2JwtKeyResolverTest.kt
```diff
@@ -71,6 +71,47 @@ class Crypto2JwtKeyResolverTest {
...
+            jwtPayload = buildJsonObject { put("iss", DID_JWK) },
+        )
+
+        assertEquals(KeyId("$DID_JWK#0"), resolved?.key?.id)
+    }
+
+    @Test
+    fun `did jwk single key is used when kid is a kms resource path`() = runTest {
+        val resolved = resolver("$DID_JWK#0").resolveFromJwt(
+            jwtHeader = buildJsonObject { put("kid", "$DID_JWK#org.tenant.kms.key_issuer") },
```
chsavvaidis: Should this be considered invalid and expected to fail resolution? The did:jwk document only defines ${did}#0. ${did}#org.tenant.kms.key_issuer does not identify a verification method in that document. Since backward compatibility is not required, this test should assert resolution failure.

waltid-libraries/credentials/waltid-w3c-credentials/src/commonTest/kotlin/Crypto2W3cCredentialTest.kt
```diff
@@ -162,6 +165,52 @@ class Crypto2W3cCredentialTest {
...
+            },
+        )
+        assertTrue(scheme.verifyCrypto2(issued, setOf(JwsAlgorithm.ES256)).isSuccess)
+
+        val alreadyIssued = CompactJws.sign(
+            payload = payload,
+            key = key,
+            algorithm = JwsAlgorithm.ES256,
+            protectedHeader = buildJsonObject {
+                put("kid", "$didJwk#org.tenant.kms.key_issuer")
```
chsavvaidis, line 208: Should this be considered invalid and expected to fail resolution? or we need something like backward compatibility here?

chsavvaidis (changes requested):

chsavvaidis (approved):

PR #2118, opened 2026-08-17
Wal 1249

chsavvaidis:
## Summary

Issuer2 demo and test profiles now issue mdocs and SD-JWT VCs with purpose-specific end-entity certificates instead of one shared chain. Mdoc profiles use a Document Signer leaf with the ISO mdoc DS EKU. SD-JWT profiles use a separate leaf for the same issuer key without that EKU. The IACA remains an external trust anchor and is not embedded in `x5Chain`.

This is the community-stack half of [WAL-1249](https://linear.app/walt-new/issue/WAL-1249/clarify-the-use-of-leaf-certificates-vs-root-certificates-for-mdocs). The Enterprise OpenAPI counterpart is [walt-id/waltid-identity-enterprise#608](https://github.com/walt-id/waltid-identity-enterprise/pull/608).

### Issuer2 profile config
- Add `defaultMdocIssuerX5chain` and `defaultHaipMdocIssuerX5chain` alongside the existing SD-JWT chains.
- Point catalog mdoc profiles, including HAIP mDL, at the Document Signer chain. Leave SD-JWT profiles on the SD-JWT chain.
- Replace the previous mixed-purpose / self-signed fixtures with AT test leaves issued by a committed IACA.
- Align a few mdoc `issuing_country` values with those AT certificates.

### Tests and HAIP fixtures
- Add a fixture test that checks leaf-versus-root, country, validity, key match, and EKU (mdoc DS present only on Document Signer leaves).
- Update profile endpoint tests so each profile resolves the matching chain.
- Refresh the HAIP runner leaf and root PEMs, and document that the SD-JWT leaf must match `defaultHaipIssuerKey` while HAIP mdoc uses a separate Document Signer leaf for the same key.

## Architecture Notes

- One issuer key can back two leaves. The mdoc Document Signer EKU (`1.0.18013.5.1.2`) must not be reused on the SD-JWT signer.
- `x5Chain` / `x5c` carries the leaf (and intermediates if any). Verifiers and the HAIP suite receive the IACA separately.
- The committed PEMs are test-only. The CA private key is not in the repository.

## Caveats and Follow-Ups

- Custom issuer2 configs that still substitute `${defaultIssuerX5chain}` into mdoc profiles, or that omit the new chain keys, will not pick up the Document Signer leaf automatically.
- When the test certificates expire, generate a new test IACA and purpose-specific leaves offline, then update the inline issuer2 chains and the public HAIP runner fixtures together.

## Breaking

- Default issuer2 demo/test certificate fixtures changed: new PEMs, AT IACA, and split SD-JWT versus mdoc chains.
- HAIP conformance runner `issuer2-haip-leaf.pem` and `issuer2-haip-root-ca.pem` changed. Trust stores and suite config that pinned the previous root need to follow.
- New required HOCON keys `defaultMdocIssuerX5chain` and `defaultHaipMdocIssuerX5chain` for any config that copies the updated profile substitutions.


waltid-services/waltid-issuer-api2/src/test/kotlin/id/walt/issuer2/profile/Issuer2ProfileEndpointTest.kt
```diff
@@ -622,6 +640,15 @@ class Issuer2ProfileEndpointTest {
         const val ISSUER_DID =
             "did:jwk:eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2IiwieCI6IkcwUklOQmlGLW9RVUQzZDVER25lZ1F1WGVuSTI5SkRhTUdvTXZpb0tSQk0iLCJ5IjoiZWQzZUZHczJwRXRycDd2QVo3QkxjYnJVdHBLa1lXQVQySlBVUUs0bE40RSJ9"
         val HAIP_PROFILE_IDS = setOf(ISO_MDL_HAIP_PROFILE_ID, IDENTITY_HAIP_SD_JWT_PROFILE_ID)
+        val MDOC_PROFILE_IDS = setOf(
```
JakeFernandes98, line 644: SCA credential is missing and is using the non-ISO certificate

JakeFernandes98 (changes requested):

JakeFernandes98 (approved):

PR #2125, opened 2026-08-18
feat(mobile): Present issuer and requester metadata consistently

szijpeter:
## Summary

This presents issuer and requester metadata consistently in the mobile wallet review screens. When an OpenID4VCI offer or OpenID4VP request supplies a friendly name and image, those values now lead the review instead of protocol identifiers, while the underlying technical values remain available on demand.

The same information hierarchy is applied across Compose Android, Compose iOS, and native iOS, with safe identifier fallbacks when display metadata is absent.

### Issuance review

- Show the issuer image and friendly name as the primary identity.
- Move the raw credential-issuer identifier into a collapsed `Issuer details` disclosure when a friendly name is available.
- Keep the identifier as the primary fallback when display metadata is absent.
- Render offered-credential image, name, and description consistently in native iOS.

### Presentation review

- Show requester image and friendly name as the primary identity.
- Keep an authenticated Digital Credentials API origin distinct from self-asserted requester metadata.
- Move additional client metadata into a collapsed `Requester details` disclosure.
- Preserve the existing separate disclosure for protocol-level technical request details.

### Cross-platform accessibility

- Add stable Compose test tags and native iOS accessibility identifiers for the new disclosures and their content.

## Architecture Notes

- This is a presentation-only change over the wallet's existing normalized issuer and requester models.
- Metadata parsing, locale selection, trust evaluation, protocol behavior, and request generation are unchanged.
- Missing or incomplete metadata continues to use the existing identifier-based fallbacks.

## Screenshots

| Review state | Compose Android | Compose iOS | Native iOS |
| --- | --- | --- | --- |
| Issuer metadata (collapsed) | Compose Android credential offer showing issuer logo and name with issuer details collapsed | Compose iOS credential offer showing issuer logo and name with issuer details collapsed | Native iOS credential offer showing issuer logo and name with issuer details collapsed |
| Issuer metadata (expanded) | Compose Android credential offer showing issuer logo, name, and expanded technical issuer field | Compose iOS credential offer showing issuer logo, name, and expanded technical issuer field | Native iOS credential offer showing issuer logo, name, and expanded technical issuer field |
| Requester metadata (collapsed) | Compose Android presentation request showing requester logo and name with requester details collapsed | Compose iOS presentation request showing requester logo and name with requester details collapsed | Native iOS presentation request showing requester logo and name with requester details collapsed |
| Requester metadata (expanded) | Compose Android presentation request showing requester logo, name, and expanded client field | Compose iOS presentation request showing requester logo, name, and expanded client field | Native iOS presentation request showing requester logo, name, and expanded client field |

## Breaking

None.


JakeFernandes98:
Theres a bit of inconsistency regarding the UI here. We can take this into this PR, raise as a follow up, or document for further work. But I think it will clean up the look and feel massively for our upcoming demos

- During the issuance flow, the offer input and related buttons are not persisted to further steps. Whereas for the presentation flow, they are present -> I like the issuer flow better. Get rid of the input section after beginning a presentation flow
- The action buttons are not immediately visible (this only affects presentation flow now, but could be the case down the line for batch issuance). Lets pin the buttons to the bottom of the content and have a scrollable view above with all the details (issuer/verifier, credentials etc.)
- Move things like technical response protection, technical request details etc. to be under the requester details dropdown. It does not need to be nested dropdowns
- When selecting credentials to share, have the claim selection screen in a dropdown. By default we should only see the credentials to select from and then selecting/seeing claims be more of an "advanced" view
- Remove the final state where we see the credential received and are left with the "New receive" option. We should just be redirected to the credentials page at the end of a successful issuance flow (or stay on the page and display the error if required)
- 

All of these changes also lend themselves well to when/if we decide to move over to review sheets.

szijpeter:
> Theres a bit of inconsistency regarding the UI here. We can take this into this PR, raise as a follow up, or document for further work. But I think it will clean up the look and feel massively for our upcoming demos
> 
> * During the issuance flow, the offer input and related buttons are not persisted to further steps. Whereas for the presentation flow, they are present -> I like the issuer flow better. Get rid of the input section after beginning a presentation flow
> * The action buttons are not immediately visible (this only affects presentation flow now, but could be the case down the line for batch issuance). Lets pin the buttons to the bottom of the content and have a scrollable view above with all the details (issuer/verifier, credentials etc.)
> * Move things like technical response protection, technical request details etc. to be under the requester details dropdown. It does not need to be nested dropdowns
> * When selecting credentials to share, have the claim selection screen in a dropdown. By default we should only see the credentials to select from and then selecting/seeing claims be more of an "advanced" view
> 
> All of these changes also lend themselves well to when/if we decide to move over to review sheets.

Thanks for the feedback @JakeFernandes98! I agree on most/all points and that the UI needs an overhaul in general. that being said unless the inconsistencies you mention are introduced here I'd rather keep this PR narrow with the current scope and treat them in other well scoped PR(s).

JakeFernandes98 (approved):

PR #2132, opened 2026-08-19
ci: cut PR queue time with path-split macOS lanes

JakeFernandes98:
## Summary

This changes Identity CI so pull requests spend less time queued on the five-job GitHub-hosted macOS cap, without dropping relevant iOS, Android, or Enterprise coverage.

A single broad macOS predicate currently fans out to simulator, native, Compose, Enterprise, and SDK docs jobs. This PR splits that into path-based lanes, assembles WalletCore once for its real consumers, and splits live OpenID conformance out of the default Linux Gradle job. Labels still add coverage; they are not the only way to get it.

### macOS eligibility
- Replaces one all-or-nothing iOS predicate with five lanes: Kotlin simulator, native WalletCore/Xcode demo, Compose iOS demo, Enterprise iOS, and SDK docs.
- Automatic path matching is the default. `ci:macos` / `ci:mobile` force every macOS lane; `ci:sdk-docs` forces docs only.
- Fixture tests cover the predicates so `commonMain` no longer launches WalletCore consumers, while wallet-mobile / Swift package / demo paths still do.

### WalletCore and SDK docs
- Native and Compose share one XCFramework assemble. SDK DocC reuses that artifact when those consumers ran, and assembles locally otherwise.
- Dokka and snippet checks run on Linux. `checkKotlinAbi` stays on macOS with iOS targets enabled. Swift DocC remains macOS work.

### Linux Gradle
- Live conformance (and the Cloudflare tunnel) move to a dedicated job. The default job skips those suites.
- PRs keep `clean cleanAllTests --no-daemon` and drop only `--rerun-tasks`. Main, release, and `ci:cacheless` keep the hermetic `--rerun-tasks` rebuild.
- Android host compilation and the unified Enterprise graph stay in the Linux job. Heap, swap, apt-get, and `cache-overwrite-existing` are unchanged.

### Merge gate
- Adds a stable `ci-gate` job that treats path-skipped lanes as passing and fails on any failed or cancelled lane. Coverage contract is in `.github/ci-coverage.md`.
- Cancels in-progress runs on pull requests only, so main is not cancelled by a later push.

## Architecture Notes

- Combined native+Compose plus a DocC artifact was chosen to cut macOS slot use; a separate producer job would add another queued Mac.
- `ci-gate` is the intended required check. Individual path-filtered jobs should not be required by name.
- Gradle task-outcome and memory summaries are recorded so later cache/heap experiments can be compared on evidence, not a promised 15–25 minute warm-build target.

## Caveats and Follow-Ups

- Combined versus artifact WalletCore topology is a slot-count choice, not a timed benchmark on this branch.
- Branch protection is not changed here; requiring `ci-gate` is a repo-settings follow-up.
- Removing `clean` / `--no-daemon`, remote build cache, heap/swap tuning, and larger runners remain later experiments.
- Node 20 / setup-java warnings are intentionally out of scope.

## Breaking

- Check names change: native and Compose demo jobs become `wallet-core-consumer-tests`. If anything required the old job names, point it at `ci-gate`.
- PR Gradle no longer passes `--rerun-tasks`, so it can use up-to-date and cache hits. Use `ci:cacheless` for a hermetic rebuild.


.github/workflows/gradle.yml
```diff
@@ -74,12 +84,126 @@ jobs:
...
+          detailed_summary: true
+          include_passed: false
+          include_skipped: false
+
+      - name: Set version
+        if: ${{ inputs.version != '' && runner.os == 'linux' }}
+        run: |
+          sed -i "s/1.0.0-SNAPSHOT/${{ inputs.version }}/g" build.gradle.kts
+
+      - name: Publish
```
szijpeter: Conformance now runs independently while this job skips it, so release and fix-release can publish Maven artifacts before a requested conformance job finishes or fails. Please make publication depend on the Gradle job and any requested conformance job.

.github/ci/macos-lanes.json
```diff
@@ -0,0 +1,77 @@
...
+        "waltid-applications/waltid-wallet-demo-ios/*",
+        "waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/*",
+        "waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-mobile/*",
+        "waltid-libraries/protocols/waltid-wallet-sdk-ios/*"
+      ]
+    },
+    "compose": {
+      "description": "Compose iOS demo and its WalletCore/Swift integration",
+      "patterns": [
+        "waltid-applications/waltid-wallet-demo-compose/*",
```
szijpeter, line 51: Both native and Compose Xcode projects consume waltid-wallet-demo-shared-ios, but changes under its Sources directory select no lane and its Package.swift selects native only. Please route this package tree to both lanes and add source/manifest fixtures expecting both.

.github/workflows/mobile-sdk-docs.yml
```diff
@@ -35,6 +79,25 @@ jobs:
       - name: Configure gradle
         uses: walt-id/waltid-identity/.github/actions/gradle-setup-action@864f84343fdeea17f6f63434bc8036c39f178dc7
 
+      - name: Restore WalletCore XCFramework
+        if: ${{ inputs.restore-wallet-core-artifact }}
+        uses: actions/download-artifact@v4
+        with:
+          name: wallet-core-release-xcframework
+          path: waltid-identity/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/build/XCFrameworks/release
```
szijpeter: The uploaded artifact contains the contents of WalletCore.xcframework at its root, so downloading it into the release directory does not recreate the WalletCore.xcframework directory checked below. Please preserve that directory during upload or download.

.github/workflows/build.yml
```diff
@@ -111,9 +134,47 @@ jobs:
...
-    needs: [android-eligibility, issuer-conformance-eligibility]
+    needs: [android-eligibility, issuer-conformance-eligibility, cacheless-eligibility]
     with:
       enable-android-build: ${{ needs.android-eligibility.outputs.should-run == 'true' }}
+      cacheless: ${{ needs.cacheless-eligibility.outputs.should-run == 'true' }}
       report-check-name: 'gradle-build / gradle (ubuntu-22.04)'
+      conformance-report-check-name: 'gradle-build / conformance (ubuntu-22.04)'
       run-issuer-conformance: ${{ needs.issuer-conformance-eligibility.outputs.should-run == 'true' }}
       run-haip-conformance: ${{ needs.issuer-conformance-eligibility.outputs.should-run == 'true' }}
+  ci-gate:
```
szijpeter, line 152: Once ci-gate is required, documentation or asset-only PRs will skip this entire workflow through paths-ignore, leaving the required check pending. Please emit ci-gate from an always-triggered workflow or move path filtering inside the workflow.

szijpeter (changes requested):
Nice improvements overall—just a few concerns; please see the inline comments.

.github/workflows/gradle.yml
JakeFernandes98: Done in `6608638a9`. Maven publish is now its own job in `gradle.yml`. It `needs` both `gradle` and `conformance`, and only runs when Gradle succeeded and conformance is `success` or `skipped` (skipped when those suites were not requested). Release and fix-release can no longer upload artifacts while a requested conformance job is still running or after it failed.

.github/ci/macos-lanes.json
JakeFernandes98: Done in `6608638a9`. `waltid-applications/waltid-wallet-demo-shared-ios/*` is now in both the native and Compose lane patterns. Fixtures cover the `Package.swift` manifest and a `Sources/` path, both expecting `native` and `compose`. Local predicate fixtures: 13 passed.

.github/workflows/mobile-sdk-docs.yml
JakeFernandes98: Done in `6608638a9`. `upload-artifact@v4` stores the *contents* of `WalletCore.xcframework` at the artifact root, so the DocC job now downloads into `.../release/WalletCore.xcframework` instead of `.../release`. The directory check below then sees `Info.plist` and the slices in the right place.

.github/workflows/build.yml
JakeFernandes98: Done in `6608638a9`. Path filtering moved inside the workflow: `paths-ignore` is gone so Build always starts and `ci-gate` can run as a required check. Docs/asset-only changes skip `gradle-build` via `gradle-eligibility` instead of skipping the whole workflow.

.github/workflows/gradle.yml
```diff
@@ -247,13 +375,33 @@ jobs:
...
+      && needs.gradle.result == 'success'
+      && (needs.conformance.result == 'success' || needs.conformance.result == 'skipped')
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout all repos with ref fallback
+        uses: walt-id/waltid-identity/.github/actions/checkout-repos@d123dc039bd55e193b3d54350791868fcfa6ac04
+        with:
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
+
+      - name: Configure gradle
```
szijpeter, line 402: Moving publication to its own job fixes the conformance ordering, but this fresh checkout no longer applies inputs.enable-android-build. Both waltid-identity/gradle.properties and the unified-build root default enableAndroidBuild=false, while release.yml and fix-release.yml pass the input as true; therefore ./gradlew publish can configure Android targets and publications out. Please repeat both Android-enablement property steps in this job before publishing.

szijpeter (changes requested):
Nice improvements overall—one remaining concern; please see below.

.github/workflows/gradle.yml
JakeFernandes98: Done in `cbc055822`. The publish job now repeats both Android-enablement property steps (`waltid-identity/gradle.properties` and the unified-build root `gradle.properties`) when `inputs.enable-android-build` is true, matching the Gradle job. Release and fix-release can therefore still configure Android targets and publications before `./gradlew publish`.

szijpeter (approved):

.github/workflows/build.yml
```diff
@@ -3,31 +3,25 @@ name: Build and test
...
-      - '**/*.webp'
-      - '**/*.svg'
-      - '**/*.ico'
-      - '**/LICENSE'
-      - '.github/ISSUE_TEMPLATE/**'
-      - '.github/PULL_REQUEST_TEMPLATE.md'
-      - 'docs/**'
   push:
     branches: [main]
-    paths-ignore: *docs-and-assets
```
dinkar-jain, line 19: I think we should revert this and avoid running CI for changes that only affect these files.

.github/workflows/mobile-sdk-docs.yml
```diff
@@ -56,10 +122,13 @@ jobs:
 
       - name: Generate mobile SDK API docs
         timeout-minutes: 45
-        uses: walt-id/waltid-identity/.github/actions/gradle-mobile-sdk-docs@88ae3106d2882f728b7c79f16295dcf665506af3
+        uses: ./waltid-identity/.github/actions/gradle-mobile-sdk-docs
         with:
           waltid-directory: 'waltid-identity'
           xcode-developer-dir: ${{ env.XCODE_DEVELOPER_DIR }}
+          generate-dokka: ${{ inputs.generate-linux-docs && 'false' || 'true' }}
```
dinkar-jain: `build.yml` always passes `generate-linux-docs: true`, which flips `generate-dokka` to `false` for the macOS job. If someone adds a public iOS function, the PR will pass without any checks

.github/workflows/build.yml
JakeFernandes98: We kept the workflow starting so `ci-gate` can still emit as a required check (restoring `paths-ignore` would leave that check pending on docs/asset PRs, and would also skip `docs/mobile-sdk-api` SDK-docs work).

`f82cd21a2` reuses the existing docs/asset classifier: `gradle-build` and `macos-predicate-tests` skip on those PRs. Path eligibility still runs so a `docs/mobile-sdk-api` change can start `sdk-docs`.

.github/workflows/mobile-sdk-docs.yml
JakeFernandes98: Done in `f82cd21a2`. `generate-dokka` is no longer tied to `generate-linux-docs`. Linux still does Android Dokka plus snippets; macOS now always runs iOS-enabled Dokka (with `checkKotlinAbi` and Swift DocC). Snippet checks stay Linux-only when that job already ran.

PR #2133, opened 2026-08-19
 Resolve public-key-only and DID-based issuers for etsi-trust-list

philpotisk:
## Summary
ETSI TS 119 602 allows a trust list entity to be registered by public key alone
(`PublicKeyValues`, no `X509Certificates`) — the standard way to register a
DID-based issuer that holds no certificate. Two things blocked this:

- `PublicKeyValues` were parsed but never made resolvable: no thumbprint was
  computed, no store query existed, and `resolveByPublicKey` was a stub that
  always returned `UNSUPPORTED_SOURCE`.
- `etsi-trust-list` hard-required an `x5c` header, so a JWT/SD-JWT credential
  signed by a DID-based issuer (e.g. `did:web`, no certificate) failed
  verification outright with "JWT has no x5c certificate chain in header."

This PR implements both sides:
- `HashUtils.computeJwkSha256Thumbprint` (RFC 7638), used by `LoteJsonParser`
  to populate `ServiceIdentity.publicKeyJwkThumbprint`, `TrustStore` to query
  by it, and `DefaultTrustRegistryService.resolveByPublicKey` to resolve it.
- `ETSITrustListPolicy`: when a JWT credential has no `x5c`, resolve the
  issuer's public key via `Crypto2JwtKeyResolver` (DID resolution or HTTPS
  well-known JWT VC issuer metadata — the same resolver already used for the
  credential's base signature check) and match it by JWK thumbprint instead
  of failing. An inline `jwk` header is intentionally not trusted for this,
  since it's self-asserted. Wired through all three resolution modes (remote
  service, inline trust lists, injected resolver).

## ES
https://github.com/walt-id/waltid-identity-enterprise/pull/623
## Architecture
- https://github.com/walt-id/waltid-architecture/pull/59
## Docs
- https://github.com/walt-id/waltid-docs/pull/366


waltid-libraries/credentials/waltid-verification-policies2/src/commonMain/kotlin/id/walt/policies2/vc/policies/ETSITrustListPolicy.kt
```diff
@@ -535,9 +614,27 @@ data class ETSITrustListPolicy(
                 }
             }
         }
+
+        /**
+         * Resolves the issuer's public key from a JWT header/payload that has no x5c chain
+         * (DID, or HTTPS well-known JWT VC issuer metadata). Inline `jwk` headers are not
+         * trusted for trust-list matching, as they are self-asserted.
+         */
+        private val keyResolver: Crypto2JwtKeyResolver by lazy { Crypto2JwtKeyResolver(allowInlineJwk = false) }
```
JakeFernandes98, line 633: **Signer binding — please confirm before merge.** Not trusting the inline `jwk` here (`allowInlineJwk = false`) is the right call. But this policy establishes issuer *trust* without verifying the credential signature itself, so trust is only sound if the key matched here is the same key that actually signed the credential.

Concretely: if the base signature-verification policy resolves its key with `allowInlineJwk = true` (or any different resolution strategy) while this trust policy resolves the DID/well-known key, an attacker could sign with a self-asserted inline `jwk` and have trust matched against a *different*, trust-listed DID key — i.e. trust asserted for a key that did not sign.

Could you confirm the base signature policy and this trust policy resolve to the **same** key material for the no-`x5c` case (same `iss`/`kid` resolution, same inline-jwk stance)? If the binding is guaranteed by pipeline ordering / a shared resolver, a short code comment stating that invariant would help, since it's load-bearing for the security of this feature. A test where the credential is signed by key A but `iss`/`kid` resolves to trust-listed key B (expected: reject) would nail it down.

waltid-libraries/credentials/waltid-trust-registry/src/commonMain/kotlin/id/walt/trust/store/InMemoryTrustStore.kt
```diff
@@ -120,6 +120,13 @@ class InMemoryTrustStore : TrustStore {
             }
         }
 
+    override suspend fun findIdentitiesByJwkThumbprint(sha256Thumbprint: String): Flow<ServiceIdentity> =
+        snapshotFlow {
+            identities.values.filter {
+                it.publicKeyJwkThumbprint?.equals(sha256Thumbprint, ignoreCase = true) == true
```
JakeFernandes98: `ignoreCase = true` is semantically wrong for a JWK thumbprint. RFC 7638 thumbprints are base64url-encoded and **case-sensitive** — `A` and `a` are distinct characters carrying distinct bits — so case-insensitive comparison widens the match set for a trust decision. This looks copied from the certificate-SHA-256 path, where the value is hex and `ignoreCase` is harmless. Since both sides here are produced by the same canonical encoder, suggest `ignoreCase = false`. (Same issue in the enterprise `EnterpriseTrustStore.findIdentitiesByJwkThumbprint`.)

waltid-libraries/credentials/waltid-trust-registry/src/jvmTest/kotlin/id/walt/trust/PublicKeyOnlyIssuerTrustTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Regression tests covering ETSI TS 119 602 trust list entities that are registered by
+ * public key only, with no X.509 certificate — a registration method the standard
+ * explicitly allows for `ServiceDigitalIdentity` (`PublicKeyValues`), but which was not
+ * resolvable by the trust registry: the LoTE JSON parser never populated
+ * `ServiceIdentity.publicKeyJwkThumbprint`, and `resolveByPublicKey` was an unimplemented
+ * stub that always returned `UNSUPPORTED_SOURCE`.
+ */
+class PublicKeyOnlyIssuerTrustTest {
```
JakeFernandes98, line 25: These unit tests (thumbprint / parser / `resolveByPublicKey`) are good, and the enterprise integration test covers `resolveByPublicKey` over HTTP. What's still untested is the most complex and security-sensitive addition in this PR: the `ETSITrustListPolicy` branch that, when a JWT/SD-JWT has **no `x5c`**, resolves the issuer key via DID/well-known metadata and matches it by thumbprint.

Could we add a policy-level test through `ETSITrustListPolicy.verify(...)` covering: (a) no-`x5c` credential whose issuer resolves and **is** in the trust list → TRUSTED; (b) resolvable issuer **not** in the list → NOT_TRUSTED; (c) issuer key unresolvable → the new error message. This also directly exercises the signer-binding concern raised on `ETSITrustListPolicy.kt`.

waltid-libraries/credentials/waltid-verification-policies2/src/commonMain/kotlin/id/walt/policies2/vc/policies/ETSITrustListPolicy.kt
philpotisk: Traced the code to confirm the binding holds structurally: both the base signature policy (JwsSignatureScheme.getIssuerCrypto2KeyInfo) and the trust policy resolve via the same Crypto2JwtKeyResolver class with the same allowInlineJwk = false default, over the same JWT header/payload. Added a code comment documenting this invariant.

waltid-libraries/credentials/waltid-trust-registry/src/commonMain/kotlin/id/walt/trust/store/InMemoryTrustStore.kt
philpotisk: Unlike the hex-identifiers, RFC 7638 thumbprints are case-sensitive! Good catch!

waltid-libraries/credentials/waltid-trust-registry/src/jvmTest/kotlin/id/walt/trust/PublicKeyOnlyIssuerTrustTest.kt
philpotisk: Added 3 new tests to ETSITrustListPolicyTest.kt exercising ETSITrustListPolicy.verify(...) directly with a real no-x5c, did:jwk-signed credential: trusted-and-listed → success, resolvable-but-unlisted → failure, unresolvable issuer → the new error message.

JakeFernandes98 (approved):

PR #2137, opened 2026-08-20
feat(android): accept historical OpenID4VCI DC API protocol aliases

JakeFernandes98:
## Summary

Android Digital Credentials issuance previously advertised and accepted only `openid4vci-v1`. Relying parties that still send Chrome's historical protocol strings (`openid4vci1.0`, `openid4vci-1.0`, `openid4vci1.1`, `openid4vci-1.1`) never reached the wallet: Google's issuance matcher treats `preferred_protocols` as exclusive, and the create provider rejected the aliases even if a request arrived.

This PR registers those aliases in the matcher and accepts them in the create provider so Android wallets can complete OpenID4VCI Digital Credentials issuance when the browser/OS forwards a historical protocol string. Canonical `openid4vci-v1` remains first in the list and is still the capability-row key. Bare `openid4vci` stays rejected.

This overrules the `openid4vci-v1`-only provider contract that landed in [walt-id/waltid-identity#2101](https://github.com/walt-id/waltid-identity/pull/2101). The alias list is Android-internal; it is not public commonMain/KLIB API.

Linear: [WAL-1322](https://linear.app/walt-new/issue/WAL-1322/accept-historical-openid4vci-dc-api-protocol-aliases)

### Android protocol list
- Added ordered `OPENID4VCI_CREATE_PROTOCOLS` as an `internal` androidMain constant shared by matcher registration and request matching.
- Canonical `openid4vci-v1` stays first; historical aliases follow.
- Tests pin the required ordered literals independently so an empty production list cannot pass.

### Android matcher
- `encodeOpenId4VciCreationOptions` now writes the full alias list into `preferred_protocols`.
- Capability rows remain a single OpenID4VCI issuance entry keyed by `openid4vci-v1`.
- No WASM rebuild: the compiled matcher already iterates `preferred_protocols`.

### Android create provider
- `resolveCreateRequest` matches against the shared alias list in preferred-protocol order, so dual-protocol requests still select `openid4vci-v1`.
- The selected protocol string is preserved and echoed in the create ack instead of being rewritten to `openid4vci-v1`.

### Docs
- Matcher contract and wallet README text now describe the alias list and selection order without exposing a public Kotlin API.

## Architecture Notes

- Alias acceptance is a registration/provider concern, not a WASM change. Google's issuance matcher only considers protocols listed in `preferred_protocols`.
- Selection uses preferred-protocol order rather than the relying party's `requests` array order, so a mixed request that includes both `openid4vci-v1` and a historical alias still picks the canonical string.
- Portal2 and other first-party relying parties stay on `openid4vci-v1`; this only widens wallet compatibility for callers that still use historical identifiers.

## Caveats and Follow-Ups

- Chrome may still refuse `openid4vci1.0` via `DigitalCredential.userAgentAllowsProtocol` before Credential Manager. This change only helps when the browser/OS forwards that protocol string.
- Device Digital Credentials E2E for historical aliases is out of scope here.
- iOS Digital Credentials create matching is unchanged; this is an Android matcher/provider expansion.

## Breaking

- None. Additional protocol strings are accepted; existing `openid4vci-v1` behavior is preserved. No public KLIB/commonMain API was added.

### DIFF
--- a/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
+++ b/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
@@ -50,8 +50,17 @@ class AndroidDigitalCredentialCreateProviderTest {
 
     @Test
     fun acceptsHistoricalOpenId4VciProtocolAliasesAndEchoesThem() {
-        val aliases = MobileWalletDigitalCredentialProtocols.OPENID4VCI_CREATE_PROTOCOLS
-            .filter { it != MobileWalletDigitalCredentialProtocols.OPENID4VCI_V1 }
+        val expectedCreateProtocols = listOf(
+            MobileWalletDigitalCredentialProtocols.OPENID4VCI_V1,
+            "openid4vci1.0",
+            "openid4vci-1.0",
+            "openid4vci1.1",
+            "openid4vci-1.1",
+        )
+        assertEquals(expectedCreateProtocols, OPENID4VCI_CREATE_PROTOCOLS)
+        val aliases = expectedCreateProtocols.filter {
+            it != MobileWalletDigitalCredentialProtocols.OPENID4VCI_V1
+        }
         for (protocol in aliases) {
             val request = AndroidDigitalCredentialCreateProvider.resolveCreateRequest(
                 requestJson = """{"requests":[{"protocol":"$protocol","data":{"credential_issuer":"https://i.example","credential_configuration_ids":["c"]}}]}""",
--- a/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
+++ b/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
@@ -49,23 +49,42 @@ class AndroidDigitalCredentialCreateProviderTest {
     }
 
     @Test
-    fun rejectsHistoricalOpenId4VciProtocolAlias() {
-        assertFailsWith<IllegalArgumentException> {
-            AndroidDigitalCredentialCreateProvider.resolveCreateRequest(
-                requestJson = """{"requests":[{"protocol":"openid4vci1.0","data":{"credential_issuer":"https://i.example","credential_configuration_ids":["c"]}}]}""",
+    fun acceptsHistoricalOpenId4VciProtocolAliasesAndEchoesThem() {
+        val aliases = MobileWalletDigitalCredentialProtocols.OPENID4VCI_CREATE_PROTOCOLS
+            .filter { it != MobileWalletDigitalCredentialProtocols.OPENID4VCI_V1 }
+        for (protocol in aliases) {
+            val request = AndroidDigitalCredentialCreateProvider.resolveCreateRequest(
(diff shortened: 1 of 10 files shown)


waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletDigitalCredentials.kt
```diff
@@ -10,8 +10,23 @@ public object MobileWalletDigitalCredentialProtocols {
...
+     * sent by some issuer pages; they are accepted and echoed, not advertised as separate
+     * capabilities.
+     */
+    public val OPENID4VCI_CREATE_PROTOCOLS: List<String> = listOf(
+        OPENID4VCI_V1,
+        "openid4vci1.0",
+        "openid4vci-1.0",
+        "openid4vci1.1",
+        "openid4vci-1.1",
+    )
```
szijpeter: This list is consumed only from `androidMain`, but declaring it public in `commonMain` adds a cross-platform API and KLIB ABI entry. That currently trips the Swift-parity gate and exposes Android matcher policy without a demonstrated non-Android consumer. Please make the shared list internal and Android-scoped, removing the KLIB/public README changes. If public API is intentional, add the required `SwiftParityDecisions.md` rationale instead.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
```diff
@@ -49,17 +49,20 @@ class AndroidDigitalCredentialCreateProviderTest {
...
 
     @Test
-    fun rejectsHistoricalOpenId4VciProtocolAlias() {
-        assertFailsWith<IllegalArgumentException> {
-            AndroidDigitalCredentialCreateProvider.resolveCreateRequest(
-                requestJson = """{"requests":[{"protocol":"openid4vci1.0","data":{"credential_issuer":"https://i.example","credential_configuration_ids":["c"]}}]}""",
+    fun acceptsHistoricalOpenId4VciProtocolAliasesAndEchoesThem() {
+        val aliases = MobileWalletDigitalCredentialProtocols.OPENID4VCI_CREATE_PROTOCOLS
+            .filter { it != MobileWalletDigitalCredentialProtocols.OPENID4VCI_V1 }
+        for (protocol in aliases) {
```
szijpeter: This test derives its cases from the production list it is meant to protect. If every historical alias is accidentally removed, `aliases` becomes empty and the test still passes. Please define the required ordered literals in the test, assert that the production list equals them, and iterate those literals when verifying acceptance and echoing.

szijpeter (changes requested):
Good, focused change. Two issues to address before merge; details inline.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletDigitalCredentials.kt
JakeFernandes98: Moved `OPENID4VCI_CREATE_PROTOCOLS` to an `internal` androidMain constant and reverted the KLIB dump. Matcher registration and the create provider still share that one list; it is no longer public commonMain API.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidHostTest/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialCreateProviderTest.kt
JakeFernandes98: The host tests now define the required ordered literals, assert the production list equals them, and iterate those literals for acceptance/echo (and for `preferred_protocols` registration).

szijpeter (approved):

PR #2091, opened 2026-08-13
[WAL-896] feat: authenticated OpenID4VP presentation and signed DC API

JakeFernandes98:
## Summary

Single identity PR for WAL-896. It now targets `main` and includes the former stacked PRs [#2055](https://github.com/walt-id/waltid-identity/pull/2055) (verifier Request Objects), [#1885](https://github.com/walt-id/waltid-identity/pull/1885) (wallet presentation), and this DC API work.

Enterprise counterpart: [walt-id/waltid-identity-enterprise#553](https://github.com/walt-id/waltid-identity-enterprise/pull/553).

- Verifier2 signed cross-device sessions expose an authenticated compact Request Object on the full authorization URL (DC API stays `request_uri`).
- Wallet2/mobile consume authenticated Final Request Objects: signed-only by default, `redirect_uri` unsigned carve-out, `aud` must contain `https://self-issued.me/v2`.
- First pass of OpenID4VP DC API signed-request resolution (`openid4vp-v1-signed`) and encrypted `dc_api.jwt` responses through `openid4vp-wallet`, `openid4vc-wallet`, `openid4vc-wallet-mobile`, and the Compose demo.
- SDK init now has `allowUnsignedRequests` (default **false**). Signed compact OpenID4VP Request Objects are always accepted. Unsigned requests (DC API `openid4vp-v1-unsigned`, and JAR `alg=none`) are accepted only when the flag is true.
- Android Credential Manager advertises signed always, and unsigned only when the flag is set. `openid4vp-v1-multisigned` still fails closed.
- Compose Android/iOS demo wallets and the native iOS demo set `allowUnsignedRequests = true` so they accept both signed and unsigned.
- `WalletPresentationHandler` can take `unsignedRequestObjectPolicy`; MobileWallet passes `ALLOW_UNSIGNED` when the flag is true, otherwise `REQUIRE_SIGNED`. MobileWallet preview rejects unsigned DC API when the flag is false.
- Compose pins copyable Verifier2 X.509 example trust anchors so `x509_san_dns` / `x509_hash` clients are not empty-list fail-closed. Wallet API 2 keeps the same PEM examples commented in `wallet-service.conf`.

## Test plan
- [x] SDK unit tests for signed DC API resolve, origin mismatch, bad signature, and signed+encrypted JWE
- [x] AuthorizationRequestResolver JVM tests
- [x] MobileWallet + Android registry/provider host tests
[...]

Issue #633, opened 2024-07-18
Java implementation issue with CredentialWrapperValidatorPolicy abtract class from waltid-veriable-credential package


vinayak0792:
Hi Team,
I'm having trouble implementing the `CredentialWrapperValidatorPolicy` class in a Java project because Java cannot directly override the `abstract suspend fun verify(data: JsonElement, args: Any? = null, context: Map<String, Any>): Result<Any>` method. 

I would appreciate any help with an adapter method or class to make it usable in a Java project.

PR #2141, opened 2026-08-21
[WAL-896] feat: authenticated OpenID4VP presentation and signed DC API

JakeFernandes98:
## Summary

This is the WAL-896 identity PR on `feature/wal-896-final`. It supersedes [walt-id/waltid-identity#2091](https://github.com/walt-id/waltid-identity/pull/2091) (`feature/wal-896-dc-api`), which is closed; that branch is kept.

Wallets and verifiers can complete authenticated OpenID4VP presentations: Verifier2 exposes signed Request Objects, Wallet2 and the mobile SDK authenticate them, and the first signed Digital Credentials API path (`openid4vp-v1-signed` with encrypted `dc_api.jwt`) is wired through shared KMP, Android, and Compose.

Default wallet behavior accepts signed Request Objects with a signable client ID, unsigned JSON fetched via `request_uri` with `redirect_uri` (the Verifier2 bootstrap, including omit-`clientId` → `redirect_uri:<response_uri>`), and unsigned `pre-registered` identifiers only when stored metadata includes matching `redirect_uris`. There is no operator flag for HAIP signed-only; that can be added later.

**Related PRs**

- Enterprise: [walt-id/waltid-identity-enterprise#633](https://github.com/walt-id/waltid-identity-enterprise/pull/633)
- Docs: [walt-id/waltid-docs#358](https://github.com/walt-id/waltid-docs/pull/358)
- License: [walt-id/waltid-identity-enterprise-license#15](https://github.com/walt-id/waltid-identity-enterprise-license/pull/15)
- Quickstart: [walt-id/waltid-enterprise-quickstart#55](https://github.com/walt-id/waltid-enterprise-quickstart/pull/55)
- Closed predecessor: [walt-id/waltid-identity#2091](https://github.com/walt-id/waltid-identity/pull/2091)

## OpenID4VP request authentication

Wallet2 authenticates the final Authorization Request before match/present. There is no `allowUnsignedRequests` (or similar) product config.

| Request | Client ID | Default |
| --- | --- | --- |
| Signed JAR | `x509_san_dns`, `x509_hash`, DID, `verifier_attestation`, `pre-registered` | Accept if trust matches (fail-closed without trust material) |
| Signed JAR | `redirect_uri` | Reject |
| Unsigned JSON from `request_uri` | `redirect_uri` | Accept |
| Unsigned JSON from `request_uri` | `pre-registered` with matching `redirect_uris` | Accept |
| Unsigned JSON from `request_uri` | `pre-registered` without `redirect_uris`, unknown id, or in-band `client_metadata` | Reject (`invalid_client`) |
| Unsigned JSON from `request_uri` | any other prefix | Reject |
| Query parameters / `alg=none` | `redirect_uri` | Accept |
| Query parameters / `alg=none` | any other prefix | Reject |

URL-shaped identifiers parse as `Unsupported` first, then fall back to `pre-registered` only when the raw string is a map key. Empty `preRegisteredClients` fails those identifiers closed.

DC API always advertises and accepts `openid4vp-v1-signed` and `openid4vp-v1-unsigned`. `openid4vp-v1-multisigned` stays rejected.

`UnsignedRequestObjectPolicy` remains inside `AuthorizationRequestResolver` (`ALLOW_UNSIGNED` default; `REQUIRE_SIGNED` for tests and a possible later HAIP profile). It is not threaded through Wallet2 routes, `wallet-service.conf`, or the mobile SDK.

### Verifier2
- Signed cross-device sessions expose an authenticated compact Request Object on the full authorization URL (DC API stays `request_uri`).
- Unsigned Verifier2 sessions that omit `clientId` bind as `redirect_uri:<response_uri>` so license heartbeats and similar bootstraps stay on the unsigned `redirect_uri` path.

### Wallet2 and shared resolver
- Final Request Objects are authenticated before match/present.
- `AuthorizationRequestResolver` implements the table above, including spec-complete unsigned `pre-registered` (registered `redirect_uris` must bind the destination; registration plus request `client_metadata` is `invalid_client`).
- Signed `pre-registered` verifies against stored JWKS. Destination URI checks run only when a redirect/response URI is present and the registered list is non-empty (DC API signed has no destination).
- Wallet API v2 loads `clientIdTrust` from `wallet-service.conf` (`x509TrustAnchors` as an in-memory PKIX trust store, plus `preRegisteredClients`). X.509 prefixes fail closed when anchors are empty.

### Mobile SDK
- Shared KMP, Android Credential Manager, and iOS Swift bridge authenticate Request Objects the same way.
- Android Credential Manager advertises signed and unsigned DC API protocols.
- `present(requestUrl)` and `previewPresentation(requestUrl)` use the same resolver policy as DC API.

### Trust material
- Compose pins copyable Verifier2 X.509 example trust anchors so `x509_san_dns` / `x509_hash` are not empty-list fail-closed.
- Wallet API v2 keeps the same PEM examples commented in `wallet-service.conf`.
- `ClientIdTrustConfiguration.x509TrustAnchors` is the wallet-controlled PKIX store. Enterprise Wallet2 may populate it from PEM pins unioned with Trust Registry relying-party identities.

## Architecture Notes

- Default HTTP OpenID4VP is signed client-ID prefixes **or** unsigned `redirect_uri` **or** unsigned `pre-registered` bound by registered `redirect_uris`. It is not a HAIP signed-only profile.
- Unsigned JSON at `request_uri` is the Verifier2 bootstrap path. OpenID4VP JAR `request_uri` is `application/oauth-authz-req+jwt`.
- A later HAIP flag can flip the resolver to `REQUIRE_SIGNED` without adding a Wallet2 config today.

## Caveats and Follow-Ups

- iOS IdentityDocumentServices OpenID4VP DC API is out of scope.
- `openid4vp-v1-multisigned` is out of scope.
- `haip-vp://` OS scheme registration is out of scope; protocol parse is scheme-agnostic and demos still register `openid4vp`.
- Wallet2 Trust Registry–derived JAR anchors land in [enterprise #633](https://github.com/walt-id/waltid-identity-enterprise/pull/633). Live issuer `resolveCertificateChain` on receive remains a later WAL-1326 slice.
- This PR does not make a wallet HAIP-complete (HAIP still expects `x509_hash` for signed requests, encrypted `direct_post.jwt`, and a signed-only HTTP profile).

## Breaking

- Wallet2 now authenticates Request Objects. Signable prefixes (`x509_*`, DID, attestation, pre-registered) fail closed without matching trust material.
- Unsigned `redirect_uri` via `request_uri` remains accepted. Unsigned `pre-registered` is accepted only with stored `redirect_uris` that match the request destination.
- Operators who want HAIP signed-only have no config switch yet.


waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/PreRegistered.kt
```diff
@@ -43,8 +43,25 @@ data class PreRegistered(override val rawValue: String) : ClientId {
                 ClientIdError.InvalidMetadata("Stored metadata is malformed: ${it.message}")
             )
         }
+        if (context.clientMetadata != null) {
```
szijpeter, line 46: **[P1] Align Verifier2's registered-client requests with this metadata guard**

Could we update the Verifier2 producer alongside this guard? `VerificationSessionCreator` always assigns `effectiveClientMetadata` to the request (line 442), including default presentation formats when the caller supplies no metadata. As a result, its generated signed and unsigned registered-client requests are rejected here with `InvalidClient`, even with the correct wallet registration.

I reproduced this locally at `bc16e7a` using real generated sessions, the exact registered client ID, a matching response URI, and the signing key's public JWK in the registration. Both cases fail authentication. Removing only the unsigned request's inline metadata makes that case succeed; the signed generated-request control passes on base `032e507`.

Please preserve this guard and adapt request generation to the out-of-band registration contract. For encrypted responses, the verifier's decryption key must also agree with the registered encryption metadata. Coverage using actual generated signed/unsigned requests, plus a registered signed/encrypted round trip, would protect the explicitly supported flow.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -827,12 +848,14 @@ object WalletPresentationHandler {
...
         onEvent: suspend (WalletSessionEvent) -> Unit = {},
         transactionDataTypeRegistry: TransactionDataTypeRegistry,
+        clientIdTrustConfiguration: ClientIdTrustConfiguration = ClientIdTrustConfiguration(),
     ): PreviewDcApiPresentationResult {
         onEvent(WalletSessionEvent.presentation_request_parsed)
         val resolvedRequest = DcApiWallet.resolveRequest(
             protocol = request.protocol,
             data = request.data,
             origin = request.origin,
+            trustConfiguration = clientIdTrustConfiguration,
```
szijpeter, line 860: **[P2] Validate registered presentation algorithms on the signed DC API path**

Could we apply the shared format/capability checks before this path creates a preview? The resolved registered metadata reaches response encryption, but its `vp_formats_supported` restrictions are not checked here as they are in HTTP's `PresentationRequestValidator`.

Locally at `bc16e7a`, I used a genuinely signed request with a trusted registered key and matching `expected_origins`, registered `jwt_vc_json.alg_values = ["ES256"]`, and a wallet with only an Ed25519 holder key. Actual DC API preview and submission succeed and return a VP with `alg=Ed25519`; I verified that outer signature with the holder key. The HTTP control with the same effective metadata and capabilities rejects with `VP_FORMATS_NOT_SUPPORTED`.

This allows consent to succeed while producing a presentation outside the verifier's advertised algorithms. Please share the transport-independent validation, while keeping DC API origin rules separate from HTTP destination rules, and cover both incompatible and compatible registered algorithms.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/DcApi.kt
```diff
@@ -218,7 +250,49 @@ public object DcApiWallet {
             data.filterKeys { it != "client_id" && it != "expected_origins" && it != "request" },
         )
         return json.decodeFromJsonElement(AuthorizationRequest.serializer(), effectiveData)
```
szijpeter, line 253: **[P2] Restore DCQL structural validation for DC API requests**

Could we retain `DcqlQuery.precheck()` in a validation stage used by both signed and unsigned DC API? Its removal here, together with the relocation of signed-request validation into the HTTP validator, leaves DC API checking query presence but not its structure.

For example, an otherwise valid request containing this query now resolves successfully:

```json
{"credentials":[{"id":"pid","format":"jwt_vc_json","meta":{},"claims":[]}]}
```

I reproduced this at `bc16e7a`: unsigned resolution accepts the empty `claims`, and an authenticated signed request reaches actual wallet preview with one credential option. The unsigned rejection test passes on base `032e507`. The existing `precheck()` rejects this malformed shape; candidate matching does not replace that validation.

Please validate before credential lookup and cover both protocols' rejection paths, including that an invalid request cannot create a preview session.

szijpeter (commented):
Thanks for the updates; these five locally reproduced findings concern the supported WAL-896 flows and consumers affected by this PR.

waltid-libraries/credentials/waltid-dcql/src/commonMain/kotlin/id/walt/dcql/DcqlMatcher.kt
```diff
@@ -98,15 +94,30 @@ object DcqlMatcher {
...
+     */
+    fun match(
+        query: DcqlQuery,
+        availableCredentials: List<DcqlCredential>,
+        trustedAuthoritiesChecker: ((DcqlCredential, List<TrustedAuthoritiesQuery>) -> Boolean)? = null,
+    ): Result<Map<String, List<DcqlMatchResult>>> {
+        val discovered = findMatches(query, availableCredentials, trustedAuthoritiesChecker)
+        if (discovered.isFailure) return discovered
+        val finalIndividualMatches = discovered.getOrThrow()
+        enforceSatisfaction(query, finalIndividualMatches.keys)?.let { return Result.failure(it) }
```
szijpeter, line 120: **[P2] Update CLI candidate discovery for the stricter matcher contract**

Could we migrate `CliDcql.match` to `findMatches` for its per-credential discovery calls? It currently calls `DcqlMatcher.match` separately for each credential/query pair and only checks the original credential sets afterward. This new satisfaction check aborts on an unavailable alternative before that final check can run.

I exercised the real `VPCreateCmd` with the repository's OpenBadge/key fixtures and a required credential set whose options are `[["badge"], ["missing"]]`. The credential satisfies `badge`; `missing` asks for an absent claim. Base `032e507` exits 0 and creates the presentation, while `bc16e7a` exits 1 with `No matches found for required credential queries: [missing]`.

Please keep strict final satisfaction and use partial discovery in the CLI, retaining its cardinality and original credential-set checks. A satisfiable-alternative test and a pool containing irrelevant credentials would cover the missed consumer integration. The existing `DcqlPresentationCommandTest` also regresses on this head.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/request/AuthorizationRequestResolver.kt
```diff
@@ -352,13 +526,17 @@ object AuthorizationRequestResolver {
             requireUnsignedRequestObjectAllowed(
                 clientId = authReqJws.payload["client_id"]?.jsonPrimitive?.contentOrNull,
                 policy = unsignedRequestObjectPolicy,
+                trustConfiguration = trustConfiguration,
+            )
+            val authorizationRequest = json.decodeFromJsonElement(
+                deserializer = AuthorizationRequest.serializer(),
+                element = applyRedirectUriPrefixBinding(authReqJws.payload),
             )
+            val registeredMetadata = authenticatePreRegisteredAuthorizationRequest(authorizationRequest, trustConfiguration)
```
szijpeter, line 535: **[P2] Adapt the existing wallet API tests to the new registered-request policy**

Could we update the affected fixtures and assertions with this policy change? Six existing tests pass on base `032e507` and fail locally at `bc16e7a` with the same causes seen in [the current Identity CI run](https://github.com/walt-id/waltid-identity/actions/runs/35349902502):

- `Wallet2IssuerVerifier2IntegrationTest`: both `eu pid sd-jwt vc full flow` and `openbadge jwt vc json full flow`.
- `Wallet2AdditionalUseCasesTest.testPresentationIsolatedStepsAndMatching`.
- `Wallet2MoreUseCasesTest`: `testIsolatedVPPresentation` and `testMultipleFormatsInOneWallet`.
- `OpenId4VpPresentationServiceTest.normalized request URL rejects unsigned request objects by default`.

The five positive cases still pin plain client IDs without registering them, so they receive HTTP 400 / `PreRegisteredClientNotFound` instead of 200. The negative case expects `UnsignedAuthorizationRequestNotAllowedException` but now receives registered-client validation failure. These are local/in-process reproductions, independent of public demo availability.

Please align positive fixtures with their intended flow: omit `clientId` for the unsigned `redirect_uri` bootstrap, or configure the required registration/trust for registered scenarios; update the negative assertion accordingly. This is a fixture/coverage fix, preserving strict authentication. It is separate from the generated-request metadata issue, which reproduces even with a correct registration.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
JakeFernandes98: Addressed in `511e0b779f`.

`previewDcApiPresentation` now runs `PresentationRequestValidator` with `PresentationValidationTransport.DigitalCredentialsApi` after resolve and before credential lookup. That shares the HTTP format/`vp_formats_supported` checks. DC API origin / `expected_origins` stay in `DcApi`; HTTP destination binding is not applied.

Covered incompatible vs compatible registered algorithms in `MobileWalletDigitalCredentialPresentationTest`.

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/PreRegistered.kt
JakeFernandes98: Addressed in `511e0b779f`.

The `PreRegistered` in-band `client_metadata` guard is unchanged. `VerificationSessionCreator` now strips that metadata from the wallet-facing JAR and URL for pre-registered client IDs, so generated signed and unsigned registered sessions authenticate against out-of-band registration. Encrypted registered sessions decrypt with the registered enc key when the caller supplies the matching private JWK.

Coverage is in `VerificationSessionCreatorMetadataTest` (unsigned/signed omit in-band metadata; signed+encrypted round-trip with the registered key).

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/DcApi.kt
JakeFernandes98: Addressed in `511e0b779f`.

`DcApi.validateAuthorizationRequest` calls `DcqlQuery.precheck()` for both signed and unsigned DC API after the null check. Preview also re-applies `precheck()` through the shared validator, so an empty `claims` query is rejected before credential lookup and cannot create a preview session.

Covered in `DcApiWalletTest` and `MobileWalletDigitalCredentialPresentationTest`.

waltid-libraries/credentials/waltid-dcql/src/commonMain/kotlin/id/walt/dcql/DcqlMatcher.kt
JakeFernandes98: Addressed in `511e0b779f`.

`CliDcql.match` now uses `DcqlMatcher.findMatches` for per-credential discovery. Final satisfaction is unchanged: `enforceCredentialSets` and `requireExactResponseCardinality` still run. `VPCreateCmd` with `[[badge],[missing]]` succeeds again (`DcqlPresentationCommandTest`).

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/request/AuthorizationRequestResolver.kt
JakeFernandes98: Addressed in `511e0b779f`.

The five positive Wallet2 fixtures now omit `clientId` so the unsigned flow uses `redirect_uri`. The negative `OpenId4VpPresentationServiceTest` asserts `PreRegisteredClientNotFound` / `invalid_client` instead of `UnsignedAuthorizationRequestNotAllowedException`. Authentication policy is unchanged.

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/PreRegistered.kt
szijpeter: Thanks, the generated signed and unsigned requests now work for plain registered names. One narrower **P2** case remains at `511e0b779f`: registered URL identifiers.

With `clientId = "https://verifier.example"`, the exact ID registered in the wallet, a matching response URI, and the signing key's public JWK in that registration, both generated signed and unsigned requests still fail with `InvalidClient`. I reproduced this through the actual wallet-facing JWT/authorization-request handler response and the shared prefix parser/authenticator; the equivalent `registered-review` controls pass.

The new [classification helper](https://github.com/walt-id/waltid-identity/blob/511e0b779f236396f0019dcf12493b5754c08dee/waltid-libraries/protocols/waltid-openid4vp-verifier/src/commonMain/kotlin/id/walt/verifier2/handlers/sessioncreation/PreRegisteredRequestMetadata.kt#L9) treats every colon as a client-ID prefix, so it retains inline metadata for the URL. The wallet intentionally supports this URL through registration fallback and correctly rejects that metadata. The PR description and companion configuration guide explicitly include URL-shaped registrations.

Could we align the verifier's classification with the recognized-prefix/fallback rules and cover generated signed/unsigned URL registrations as well? Please apply the same classification to registered encryption-key selection. I’m keeping this thread open for that remaining supported case; the original plain-name reproduction is fixed.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
szijpeter: Thanks, the single-format incompatible-algorithm case now rejects correctly. A **mixed-format registered request** still reproduces the underlying issue at `511e0b779f`.

Local reproduction: a genuinely signed, trusted DC API request with matching `expected_origins` allows either `pid` (`jwt_vc_json`) or `sd` (`dc+sd-jwt`), using required alternatives `[["pid"], ["sd"]]`. Registered metadata is:

```json
{
  "jwt_vc_json": {"alg_values": ["ES256"]},
  "dc+sd-jwt": {
    "sd-jwt_alg_values": ["Ed25519"],
    "kb-jwt_alg_values": ["Ed25519"]
  }
}
```

The wallet has only the W3C credential and an Ed25519 holder key, with no SD-JWT credential. Preview nevertheless offers the W3C option, and actual submission returns a W3C VP with `alg=Ed25519`. I cryptographically verified that outer signature. Changing only the W3C allowed algorithm to `Ed25519` gives a passing control.

`supportsAny` accepts the compatible but unavailable SD-JWT alternative. Then [buildVpToken](https://github.com/walt-id/waltid-identity/blob/511e0b779f236396f0019dcf12493b5754c08dee/waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/WalletPresentFunctionality2.kt#L502) uses effective registered metadata only for the encryption thumbprint and passes the original request to the presenters. That request correctly has no inline metadata, so the W3C presenter misses the registered algorithm restriction.

Could we carry effective metadata through per-format signing and exclude incompatible credential options before consent? This remains P2 and within the signed DC API scope. I’m leaving the thread open for this case.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/DcApi.kt
szijpeter: Verified locally at `511e0b779f`: both signed and unsigned `claims: []` reproductions now reject, including the signed wallet-preview path. The DC API suite also passes. This is addressed; resolving. Thanks!

waltid-libraries/credentials/waltid-dcql/src/commonMain/kotlin/id/walt/dcql/DcqlMatcher.kt
szijpeter: Verified locally at `511e0b779f`: the real `VPCreateCmd` reproduction with `[["badge"], ["missing"]]` succeeds again, and all seven existing `DcqlPresentationCommandTest` cases pass. The discovery change preserves final satisfaction checks. Resolving, thanks!

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/request/AuthorizationRequestResolver.kt
szijpeter: Verified locally at `511e0b779f`: all six originally reported tests now pass, including the five positive Wallet API2 flows and the updated negative exception assertion. Strict authentication remains intact. This is addressed; resolving. Thanks!

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/PreRegistered.kt
JakeFernandes98: Addressed in `0480238378`.

`isPreRegisteredVerifierClientId` now follows the recognized-prefix rules: a client id is pre-registered when it has no colon or when the text before the first colon is not a recognized OpenID4VP prefix. `https://verifier.example` therefore drops in-band metadata on generated signed and unsigned requests, and the same classification selects a registered encryption key when one is supplied. Recognized prefixes (`redirect_uri`, `x509_*`, DID, attestation, federation) still keep in-band metadata.

Covered by `VerificationSessionCreatorMetadataTest` for unsigned and signed URL client ids.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
JakeFernandes98: Addressed in `0480238378`.

`supportsFormat` now decides each credential option on its own. A mixed request can still pass `supportsAny` when another format is compatible, but preview drops options whose registered holder algorithms the wallet key cannot satisfy before consent. `buildVpToken` also copies effective registered metadata onto the local request used by the presenters, so signing enforces those algorithms even when the wire request correctly has no inline `client_metadata`.

Covered by `WalletPresentationFormatRegistryJvmTest` and `MobileWalletDigitalCredentialPresentationTest.mixedFormatPreviewDoesNotOfferACredentialOutsideItsRegisteredAlgorithms`.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -886,7 +928,11 @@ object WalletPresentationHandler {
             }
         }.filter { option ->
             request.eligibleCredentialIds?.let { option.credentialId in it } ?: true
-        }
+        }.compatibleWithVerifierFormats(
+            verifierFormats = resolvedRequest.effectiveClientMetadata?.vpFormatsSupported,
+            capabilities = keyMaterial?.presentationCapabilities()
+                ?: WalletPresentationFormatRegistry.defaultCapabilities(),
```
szijpeter: **[P2] Filter mdoc options using their credential-bound key**

Could we evaluate each mdoc against its resolved holder key here? `keyMaterial` comes from `wallet.resolveKeyMaterial(null, SIGN)`, so these capabilities describe the wallet's default key. Submission instead uses `wallet.mdocHolderKeyResolver()` for the mdoc's bound key. Applying the default key's algorithms to every option can therefore remove an mdoc that the wallet can present.

I reproduced this at `0480238378` through the MobileWallet preview/submission APIs with an Ed25519 default key and a separate P-256 key bound to an issued mdoc. A signed registered request allows either mdoc (`deviceauth_alg_values: [-7]`) or SD-JWT (`kb-jwt_alg_values: ["Ed25519"]`), with only the mdoc stored. The request-level check accepts the SD-JWT alternative, but this filter removes the compatible mdoc and preview throws `The selected registry entries no longer satisfy this request`.

The identical case completes preview and submission on parent `511e0b779f`. On the current head, changing only the unrelated default key to P-256 also makes it succeed. This is a regression in the supported signed DC API flow; the bound mdoc key already satisfies the verifier's restriction.

Please retain per-format algorithm enforcement, derive mdoc eligibility from its bound key, and add a regression with different default/bound-key algorithms. The existing mdoc test uses two P-256 keys, which cannot expose this mismatch. Thanks!

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/PreRegistered.kt
szijpeter: Verified locally at `0480238378`: generated signed and unsigned requests now authenticate for both plain and URL-shaped registrations. The URL-registered encryption test also decrypts with the configured key. This addresses the remaining case; resolving. Thanks!

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
szijpeter: Verified locally at `0480238378`: the original incompatible W3C mixed-format case now rejects before consent, while the compatible control still submits and its outer signature verifies. Effective registered metadata reaches presentation generation. This finding is addressed; resolving. The separate mdoc key-selection regression is tracked in a new inline comment on the added filter.

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/AttestationX509CertificateValidator.kt
```diff
@@ -23,25 +23,23 @@ class AttestationX509CertificateValidator : X509CertificateValidator {
...
         }
 
-        if (x509Certificate.data.extensionExtendedKeyUsage
-                ?.keyPurposeList
-                ?.contains(ExtendedKeyUsageExtension.KeyUsage.clientAuth) != true
+        // RFC 5280 §4.2.1.12: ExtendedKeyUsage is unrestricted when absent. When present it must
+        // include clientAuth for this Request Object signing profile.
+        val extendedKeyUsage = x509Certificate.data.extensionExtendedKeyUsage
+        if (extendedKeyUsage != null &&
+            ExtendedKeyUsageExtension.KeyUsage.clientAuth !in extendedKeyUsage.keyPurposeList
```
szijpeter: **[P2] Respect the verifier certificate’s intended purpose**

This check rejects the WE BUILD ITB verifier certificate because it has the reader-authentication EKU (`mdlReaderAuth`, `1.0.18013.5.1.6`) instead of TLS `clientAuth`. Making this fatal regresses VP001/002/003 and blocks the signed DC API cases.

[OpenID4VP §5.9.3](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.9.3) requires certificate-chain, signature and client-ID validation; it does not mandate TLS `clientAuth`. [RFC 5280 §4.2.1.12](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.2.1.12) requires respecting the intended EKU and defines `clientAuth` specifically for TLS client authentication. Please validate the purpose required by the wallet’s ecosystem/profile, retaining the other certificate and signature checks.

Confirmed locally against `0480238378`: explicitly selecting reader authentication makes **VP001/002/003, CS07 `tc15`, and TS12 `ts12_pay_dc_api_01/02/03` pass end to end**, including ITB `SUCCESS`. VP007 also gets past authentication but remains blocked by its missing mdoc credential.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
JakeFernandes98: Addressed in `72f71e8cc5`.

DC API preview still enforces registered algorithms per format. An `mso_mdoc` option is now judged against `mdocHolderKeyResolver` for that credential, which is the key submission uses. Other formats still use the wallet default signing key.

`signedDcApiPreviewKeepsAnMdocSignedByItsBoundKeyWhenTheDefaultKeyCannot` covers an Ed25519 default key and a P-256 key bound to the mdoc, with registered `deviceauth_alg_values: [-7]` and an Ed25519 SD-JWT alternative.

waltid-libraries/protocols/waltid-openid4vp-clientidprefix/src/commonMain/kotlin/id/walt/openid4vp/clientidprefix/prefixes/AttestationX509CertificateValidator.kt
JakeFernandes98: Addressed in `13e24d3603`.

When Extended Key Usage is present, the leaf is accepted if it names TLS `clientAuth` (`1.3.6.1.5.5.7.3.2`), ISO 18013-5 reader authentication (`1.0.18013.5.1.6`), or ISO 23220-4 reader authentication (`1.0.23220.4.1.6`). A missing EKU stays unrestricted. Key usage, chain, and signature checks are unchanged. An EKU that names none of those purposes, such as `serverAuth` only, is still rejected.

The verifier docs now describe the same accepted purposes.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -827,20 +850,42 @@ object WalletPresentationHandler {
...
+                client = AuthenticatedClientFacts(
+                    effectiveClientMetadata = resolvedRequest.effectiveClientMetadata,
+                    responseDestinationAuthenticated = true,
+                    boundResponseDestination = resolvedRequest.origin,
+                ),
+            ),
+            transactionDataTypeRegistry = transactionDataTypeRegistry,
+            formatCapabilities = {
+                keyMaterial?.presentationCapabilities()
+                    ?: WalletPresentationFormatRegistry.defaultCapabilities()
```
szijpeter: **[P2] Use the mdoc bound key in request-level format validation too**

Thanks, the per-option fix now lets the previously reported mixed-format mdoc request complete preview and submission. A request for **only that same mdoc** still fails at this earlier check, before reaching the corrected option filter.

At `13e24d3603`, I reproduced this with an Ed25519 default wallet key, a P-256 key bound to the available mdoc, and a signed registered request whose sole DCQL query is `mso_mdoc` and whose registered metadata allows `deviceauth_alg_values: [-7]`. `formatCapabilities` describes the default key, so `PresentationRequestValidator` returns `vp_formats_not_supported` and preview throws `The wallet supports none of the presentation formats requested by the verifier`. The mdoc's actual bound key supports the requested algorithm.

Two controls pass on the same head: adding the unavailable Ed25519 SD-JWT alternative, or changing only the unrelated default key to P-256. Both complete actual preview and submission. This is the same supported signed DC API path, with a simpler single-credential request.

Could we make this early format decision respect credential-bound mdoc capabilities as well, or defer that decision to the per-credential eligibility stage while retaining request structure, authentication, and algorithm checks? A single-mdoc regression alongside the new mixed-format test would protect this remaining case.

JakeFernandes98: Addressed in `5600ddf3cd`.

DC API preview still checks request structure and authentication before reading credentials. The format decision now includes device-authentication algorithms from each stored mdoc's bound key, so a sole `mso_mdoc` request is not rejected because the wallet default key cannot satisfy `deviceauth_alg_values`. Per-option filtering is unchanged: mdoc options still use that bound key, and other formats still use the default key.

`signedDcApiPreviewKeepsASoleMdocSignedByItsBoundKeyWhenTheDefaultKeyCannot` covers an Ed25519 default key, a P-256 key bound to the only mdoc, and registered `deviceauth_alg_values: [-7]`.

szijpeter (approved):

PR #2142, opened 2026-08-21
feat(wallet-demo): settings, credential actions, branding, and flush bottom nav

JakeFernandes98:
## Summary

[WAL-1188](https://linear.app/walt-new/issue/WAL-1188/improved-devex-for-wallet)

This adds a Settings surface to the Compose and native iOS wallet demos so developers can inspect the wallet DID, key ID, and public JWK, reset local wallet state, and (Compose only) lock the PIN session without crowding the header. Credential details also gain copy-raw and delete, and status banners no longer sit on screen forever. Also fixes the Android and Compose iOS bottom navigation gap.

Both demos now have a single white-label branding object for the in-app wallet title and brand colours (primary, secondary, container, and on-colours), so a rebrand does not require hunting through individual screens.

Related PRs: stacked native PIN and biometric unlock lives in [walt-id/waltid-identity#2143](https://github.com/walt-id/waltid-identity/pull/2143).

### White-label branding
- Compose: `WalletDemoBranding` plus `WalletDemoTheme` drive the header/PIN title and Material3 palette. Hosts can pass a custom instance into `WalletDemoApp`.
- Native iOS: `WalletDemoBranding.default` in the shared demo package drives the Credentials title, Settings name, app tint, and success-banner colours. Views read `@Environment(\.walletDemoBranding)`; the old `.waltBlue*` aliases are removed.
- Launcher names stay in platform manifests (`app_name`, `CFBundleDisplayName`).

### Compose demo
- Header Lock and DID are replaced by a Settings icon. Settings shows DID, key ID, public JWK, Lock, and Reset Wallet (with confirm).
- Credential details can copy the raw credential and delete it (with confirm). Deleting a credential that is in an open presentation review discards that preview and selections.
- Status banners hide when idle. Success auto-hides after 4s and can be dismissed. Errors stay collapsed until expanded, and can be dismissed. Dismissal is occurrence-local, so a later identical outcome can show again.
- Reset wipes wallet data first and always returns to Setup after a successful wipe. A later PIN-clear failure surfaces an error instead of leaving Ready with stale credentials.
- Bottom nav no longer sits above a safe-area gap: the app still insets the status bar and sides, but leaves bottom insets for `NavigationBar` so the bar is flush with the screen.

### Native iOS demo
- Each tab has a Settings screen with DID, key ID, public JWK, and Reset Wallet. There is no PIN lock on this app.
- Credential copy/delete and dismissable / expandable status banners match the Compose behavior, including occurrence-local dismissal and discarding an open presentation review after delete.

### Wallet SDK
- `MobileWallet.deleteCredential` is exposed through the iOS SDK bridge so the native demo can delete a stored credential.
- Bootstrap now also returns the public JWK (`MobileWalletBootstrapResult.publicJwk` / `WalletBootstrapResult.publicJWK`). Private material is never exported.

## Architecture Notes

- Compose Settings is an in-app overlay that hides the tab scaffold; Lock still unmounts it by returning to PIN entry.
- Native Settings is a navigation destination from each tab. Reset clears wallet data through the existing client APIs rather than a separate persistence path.
- Branding is demo-scoped, not a Wallet SDK API. Compose theme and native environment injection stay in the demo modules / shared demo package.
- The bottom-nav fix is Compose-only (Android and iOS hosts). The native `TabView` was already edge-connected.

## Caveats and Follow-Ups

- Compose Lock remains Compose-only; native iOS has no PIN lock by design.
- Wasm still uses `MockDemoWallet` and is a UI preview only.
- E2E that waits for `"Wallet ready"` needs to catch it while the success banner is visible (it auto-hides after 4s).
- Changing the in-app title does not rename the Android or iOS home-screen label; those remain manifest values.

## Breaking

- `MobileWalletBootstrapResult` now requires `publicJwk`. The Swift `WalletBootstrapResult` adds the same field with a test-only default so existing fakes still compile.
- None otherwise for published APIs. Demo UI chrome, status-banner lifetime, and default Compose palette (now walt-blue instead of stock Material3 purple) change for local wallet testing.


JakeFernandes98:
Wallet key in the settings is just a string. The private key is not exportable, so we should provide maybe just a public jwk to copy?

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoUiStateDerivedProperties.kt
```diff
@@ -1,5 +1,19 @@
...
+    Info,
+    Success,
+    Error,
+}
+
+data class WalletStatusBanner(
+    val message: String,
+    val kind: WalletStatusKind,
+) {
+    val key: String get() = "$kind:$message"
```
szijpeter: [P2] Treat dismissal as occurrence-local rather than keying it only by kind and message. Once an outcome is dismissed or auto-hidden, a later identical outcome gets the same key and remains invisible. The native `WalletStatusBannerModel` has the same issue.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/WalletScreen.kt
```diff
@@ -120,6 +137,7 @@ internal fun WalletScreen(controller: WalletDemoController, state: WalletDemoUiS
                             onCancel = controller::cancelPresentationReview,
                         )
                     },
+                    onDeleteCredential = controller::deleteCredential,
```
szijpeter, line 140: [P2] Please either hide deletion while presentation review is active or invalidate/discard the preview and selections after deletion. Currently only the ready credential list is refreshed, so the review can still display and submit the deleted credential. Native iOS has the same stale-state path.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoController.kt
```diff
@@ -97,6 +123,66 @@ class WalletDemoController(
...
+
+    fun resetWallet() {
+        receiveJob?.cancel()
+        presentationJob?.cancel()
+        val previous = getAndUpdateState { it }
+        cancelIssuance()
+        discardPresentationPreview(previous.activePresentationPreviewHandle())
+        scope.launch(dispatcher) {
+            runCatching {
+                wallet.deleteWallet()
```
szijpeter: [P2] Please handle partial completion explicitly. Android PIN clearing throws when the `SharedPreferences` commit fails; if that happens after `deleteWallet()` succeeds, the UI remains Ready with the pre-deletion credentials even though the wallet data is gone.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/WalletColors.swift
```diff
@@ -1,8 +1,8 @@
 import SwiftUI
 
 public extension Color {
-    static let waltBlue = Color(red: 5/255, green: 115/255, blue: 240/255)
-    static let waltBlueLight = Color(red: 173/255, green: 198/255, blue: 255/255)
-    static let waltBlueContainer = Color(red: 216/255, green: 226/255, blue: 255/255)
-    static let waltBlueDark = Color(red: 0, green: 46/255, blue: 105/255)
+    static var waltBlue: Color { WalletDemoBranding.default.primary }
```
szijpeter: [P2] These aliases always read `WalletDemoBranding.default`, so views using `.waltBlue*` override an injected environment palette. Please consume the environment branding at those call sites or remove/migrate the aliases so the advertised single-object rebrand is complete.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidHostTest/kotlin/id/walt/walletdemo/compose/ui/WalletDemoAppAndroidTest.kt
```diff
@@ -101,4 +101,19 @@ class WalletDemoAppAndroidTest {
     @Test
     fun credentialsPersistAcrossControllerRecreation() =
         scenarios.credentialsPersistAcrossControllerRecreation()
+
+    @Test
+    fun customBrandingTitleAppearsInTheHeader() =
+        scenarios.customBrandingTitleAppearsInTheHeader()
+
+    fun settingsReplacesHeaderLockAndShowsDidAndKey() =
```
szijpeter, line 110: [P2] Please add `@Test` here and to the Compose iOS host wrapper. The scenario exists, but neither host registers it; it was absent from the exact-head generated test results.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletClient.swift
```diff
@@ -20,6 +20,13 @@ protocol WalletClient {
     ) async throws -> PresentationResult
     func rejectPresentation(previewHandle: PresentationPreviewHandle) async throws -> PresentationResult
     func discardPresentationPreview(_ previewHandle: PresentationPreviewHandle) async throws
+    func deleteCredential(id: String) async throws -> Bool
+    func deleteLocalData() async throws
+}
+
+extension WalletClient {
```
szijpeter: [P3] Please remove these false/no-op defaults and implement the requirements in `TransactionCodeWalletClient`. This internal protocol needs no compatibility fallback, and silent defaults can make a test conformer appear to support delete/reset.

szijpeter (changes requested):
Nice improvements overall. I have a couple of concerns; please see the inline comments below.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoUiStateDerivedProperties.kt
JakeFernandes98: Done in `ffc149b5d`. Dismissal is now occurrence-local in Compose and native: the banner key includes an incrementing occurrence id, so a dismissed or auto-hidden outcome does not hide a later identical one. Covered by `dismissedSuccessReappearsOnLaterIdenticalOutcome` and the native reset-after-dismiss assertion.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/WalletScreen.kt
JakeFernandes98: Done in `ffc149b5d`. Deleting a credential now clears the presentation review and selections and discards the preview handle on both Compose and native, so the deleted credential cannot still be displayed or submitted. Covered by `deleteCredentialDiscardsAnActivePresentationReview` on both sides.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoController.kt
JakeFernandes98: Done in `ffc149b5d`. `deleteWallet()` and PIN clear are split. After a successful wipe we always return to Setup; a later PIN-clear failure surfaces `ResetWalletFailed` instead of leaving Ready with stale credentials. Covered by `resetWalletLeavesSetupWhenPinClearFailsAfterDelete`.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/WalletColors.swift
JakeFernandes98: Done in `ffc149b5d`. The remaining `.waltBlue*` call sites now read `@Environment(\.walletDemoBranding)`, and `WalletColors.swift` / the aliases are removed so an injected palette is not overridden.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidHostTest/kotlin/id/walt/walletdemo/compose/ui/WalletDemoAppAndroidTest.kt
JakeFernandes98: Done in `ffc149b5d`. `@Test` is on `settingsReplacesHeaderLockAndShowsDidAndKey` for both the Android and Compose iOS host wrappers.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletClient.swift
JakeFernandes98: Done in `ffc149b5d`. The protocol-extension no-ops are gone, and `TransactionCodeWalletClient` implements `deleteCredential` / `deleteLocalData`.

JakeFernandes98:
Implemented on the public-JWK question: Settings now shows and copies the bootstrap public JWK (`MobileWalletBootstrapResult.publicJwk` / native `WalletBootstrapResult.publicJWK`) on Compose and native. The key ID row is still there. Private material is never exported.

waltid-libraries/protocols/waltid-wallet-sdk-ios/Sources/WalletSDK/WalletModels.swift
```diff
@@ -712,14 +712,19 @@ public struct WalletBootstrapResult: Equatable, Sendable {
...
+    public let publicJWK: String
+
     /// Creates a bootstrap result.
     ///
     /// - Parameters:
     ///   - keyID: Identifier of the created or selected wallet key.
     ///   - did: DID created for the wallet.
-    public init(keyID: String, did: String) {
+    ///   - publicJWK: Public JWK of the wallet key as a JSON object string.
+    public init(keyID: String, did: String, publicJWK: String = #"{"kty":"OKP","crv":"Ed25519","x":"test"}"#) {
```
szijpeter: **[P2] Remove the fixture JWK default from this public initializer.** `publicJWK` must correspond to `keyID`, but callers that omit it currently receive fabricated test key material. The production bridge supplying the real value does not make the public initializer safe, and Kotlin already requires this argument. Since mobile backward compatibility is not required here, please make it mandatory and update Swift test fixtures explicitly.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
```diff
@@ -187,6 +210,83 @@ class WalletViewModel: ObservableObject {
...
+        guard statusBanner(for: selectedTab)?.kind == .error, isStatusVisible(for: selectedTab) else { return }
+        statusExpanded.toggle()
+    }
+
+    func deleteCredential(id: String) {
+        Task {
+            do {
+                let removed = try await walletClient.deleteCredential(id: id)
+                guard removed else { return }
+                credentials = try await walletClient.credentials()
```
szijpeter, line 240: **[P2] Reconcile Apple registrations after destructive mutations.** Unlike Compose iOS, this native path does not trigger `DemoIdentityDocumentRegistration.update()` after deletion. The core writes the reduced desired projection to the App Group, but Apple’s `IdentityDocumentProviderRegistrationStore` can therefore keep offering the deleted document until the next foreground transition or periodic provider callback. Please reconcile immediately after the destructive mutation. The reset path should likewise reconcile directly after `deleteLocalData()` rather than relying on a later successful bootstrap, so a bootstrap or credential-refresh failure cannot leave stale registrations behind.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoController.kt
```diff
@@ -97,6 +123,80 @@ class WalletDemoController(
...
+            val deleted = runCatching { wallet.deleteWallet() }
+            deleted.exceptionOrNull()?.let { error ->
+                setOperationError(WalletDisplayText.ResetWalletFailed, error, _state.value.selectedTab)
+                return@launch
+            }
+            val pinCleared = runCatching { pinStore.clear() }
+            statusHideJob?.cancel()
+            _state.value = WalletDemoUiState(auth = WalletAuthState.Setup())
+            pinCleared.exceptionOrNull()?.let { error ->
+                setOperationError(WalletDisplayText.ResetWalletFailed, error, WalletDemoTab.Credentials)
```
szijpeter: **[P3][Non-blocking] Surface this PIN-clear failure on the Setup route.** After the state is replaced with `WalletDemoUiState(auth = Setup())`, `setOperationError` stores the failure only in `operation`, while `PinScreen` reads `WalletAuthState.Setup.error`. The user is therefore asked to create a PIN without being told that the old PIN remains stored. This requires a storage-clear failure, so I consider it an edge case, but the recovery state should still expose the failure.

szijpeter (changes requested):
Nice improvements, but a few concerns remain; please see below.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/WalletScreen.kt
szijpeter: Still reproducible on current head `b43a5ba9` through the actual presentation-review details flow. The presentation option is mapped to details using `selection.id`—the composite query/credential UI identifier—and `WalletTabNavDisplay` passes that value to `deleteCredential`. The store expects the wallet-local `credentialId`, so removal returns `false`; the credential and review remain, and the presentation can still be submitted. Compose iOS shares this path, while native similarly maps `option.selection.id` into the details ID and then deletes by that value. The current tests call `deleteCredential("cred-1")` directly and miss this UI boundary. Please pass the real credential ID to deletion on both implementations and cover the presentation-details flow.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoUiStateDerivedProperties.kt
szijpeter: Partially addressed: successful outcomes now receive fresh occurrence IDs, but several terminal Compose error paths still assign `WalletOperationState.Failed` without `withPublishedStatus()`—for example offer resolution. After dismissing such an error, retrying the same failing request can reuse the dismissed key and leave the new error hidden. Retrying the same issuer or request is a normal user flow, so please bump the occurrence for every published terminal result and add coverage for a repeated identical error.

JakeFernandes98: Done in `a03e204af`. Every published Compose `Failed` path now goes through `withFailedOperation()` / `withPublishedStatus()`, so a dismissed identical error gets a new occurrence id. Covered by `dismissedErrorReappearsOnLaterIdenticalFailure`.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/WalletScreen.kt
JakeFernandes98: Done in `a03e204af`. Presentation details still navigate by `selection.id`. Delete now uses the store `credentialId` on Compose and native. Covered by `presentationDetailsDeleteUsesStoreCredentialId` on the controller, Compose hosts, and native ViewModel.

waltid-libraries/protocols/waltid-wallet-sdk-ios/Sources/WalletSDK/WalletModels.swift
JakeFernandes98: Done in `a03e204af`. `WalletBootstrapResult.publicJWK` is required. Swift fixtures pass it explicitly.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
JakeFernandes98: Done in `a03e204af`. Native delete reconciles after a successful removal. Reset reconciles after `deleteLocalData()` and before bootstrap; a reconcile failure is surfaced, then bootstrap still proceeds. Covered by `testDeleteAndResetReconcileIdentityDocumentRegistrations`.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoController.kt
JakeFernandes98: Done in `a03e204af`. After a successful wipe and a failed PIN clear, `WalletAuthState.Setup.error` now shows the reset-failed message. `resetWalletLeavesSetupWhenPinClearFailsAfterDelete` asserts that text.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
szijpeter: **[P1] The production calls are present, but the cited reset coverage is currently red.** After the preceding deletion, `isReady` is still `true` and `credentials` is already empty, so the wait after `resetWallet()` returns before its asynchronous task starts. The unchanged counter then fails with `2 is not greater than 2`. This reproduces both locally and in the exact-head native iOS CI run (121 tests, 120 passed, 1 failed). Please synchronize on an observable reset operation and verify the direct reset reconciliation independently from the later bootstrap reconciliation.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
```diff
@@ -187,7 +210,91 @@ class WalletViewModel: ObservableObject {
...
+                try await reconcileIdentityDocumentRegistrations()
```
szijpeter, line 241: **[P3][Non-blocking] Invalidate the active review before the fallible registry reconciliation.** The credential is already deleted and the local list refreshed here. If Apple reconciliation throws, control reaches the catch before the presentation review and selections are cleared, leaving a stale review while reporting that deletion failed. Submission later fails because core no longer finds the credential. Please complete the in-app deletion cleanup before platform reconciliation and report any reconciliation failure separately.

szijpeter (changes requested):
Nice improvements, couple of concerns, pls see below.

szijpeter (approved):
thanks for the updates, lgtm now, nice work 👏🏼 

(https://github.com/walt-id/waltid-identity/pull/2142#discussion_r3852045040 remains, but it's non-blocking imo)

PR #2143, opened 2026-08-21
feat(wallet-demo): native PIN lock and biometric unlock

JakeFernandes98:
## Summary

This adds the missing PIN gate to the native iOS wallet demo and lets both demo apps opt into biometric unlock during first-run PIN setup. The PIN is still required; Face ID or fingerprint is an additional unlock path, with PIN as the fallback when biometrics fails.

Stacked on [walt-id/waltid-identity#2142](https://github.com/walt-id/waltid-identity/pull/2142). Tracks [WAL-1189](https://linear.app/walt-new/issue/WAL-1189/add-native-ios-pin-lock-and-biometric-unlock-for-both-wallet-demos).

### Native iOS demo
- First-run PIN setup (4–8 digits, confirmation) and unlock on later launches.
- Settings now has Lock, which returns to the PIN screen without destroying the wallet session.
- Reset wallet clears PIN and the biometric-unlock preference and returns to PIN setup. Bootstrap waits until unlock.

### Biometric unlock (Compose Android, Compose iOS, native iOS)
- PIN setup shows an **Unlock with biometrics** toggle. It is disabled when the device cannot offer biometrics.
- Cold launch / process start still auto-prompts biometrics when the preference is on.
- Settings → Lock returns to the PIN screen without auto-prompting. The **Unlock with biometrics** button remains available.
- Failing biometrics leaves the PIN fields as the fallback.

### Wallet-key biometrics
- Unchanged. `WALLET_BIOMETRIC_ENABLED` still only controls signing/key-use authorization, not demo unlock.

## Architecture Notes

- Compose keeps PIN persistence in `DemoPinStore` and adds a `DemoBiometricAuthenticator` injected from the Android Activity and iOS host.
- Native iOS mirrors that with `DemoPinStore` / `DemoBiometricAuthenticator` on `WalletViewModel`. Injected mock clients still use an in-memory PIN store so unit tests do not leak UserDefaults state.
- This is demo unlock only. It does not change Secure Enclave or platform key authorization.

## Caveats and Follow-Ups

- Simulator and hosts without enrolled biometrics stay PIN-only; the toggle is disabled.
- Wasm remains a mock UI preview with no real biometrics.
- Native demo PIN records now use the same PBKDF2-SHA256 verifier as Compose (`1:<iterations>:<saltB64>:<verifierB64>`, 210,000 iterations). Existing native demo PIN records from the previous hash format will not verify; reset or re-setup the PIN.
- Settings → Lock marks the biometric prompt as already consumed so Face ID / fingerprint does not fire immediately after lock. Cold launch still auto-prompts.

## Breaking

- Native iOS demo now requires PIN setup before the wallet bootstraps. Existing native E2E helpers enter `1234` on launch, matching Compose.
- Native iOS demo PIN storage format changed to match Compose. Previously stored native demo PINs must be set again.


waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/androidMain/kotlin/id/walt/walletdemo/compose/logic/AndroidDemoBiometricAuthenticator.kt
```diff
@@ -0,0 +1,65 @@
...
+                            BiometricPrompt.ERROR_CANCELED,
+                            -> DemoBiometricResult.Cancelled
+                            else -> DemoBiometricResult.Failed
+                        }
+                        continuation.resume(result)
+                    }
+                },
+            )
+            continuation.invokeOnCancellation { prompt.cancelAuthentication() }
+            prompt.authenticate(
```
szijpeter: **[P1] Run the biometric prompt on the main thread.** This is reached from the controller’s `Dispatchers.Default` scope, but `BiometricPrompt.authenticate()` requires the fragment host’s main thread. On the physical device this throws `IllegalStateException: Must be called from main thread of fragment host`; persisted auto-unlock then makes every subsequent launch crash. Please enforce `Dispatchers.Main.immediate` inside this adapter, verify that the activity is resumed, and add coverage for the real platform boundary.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
```diff
@@ -268,14 +285,80 @@ class WalletViewModel: ObservableObject {
...
                 setError(WalletStatusText.failure(WalletStatusText.resetWalletFailed, error))
             }
         }
     }
 
+    func lock() {
+        receiveTask?.cancel()
+        presentationTask?.cancel()
+        cancelIssuanceIfPresent()
+        discardPresentationPreviewIfPresent()
```
szijpeter, line 325: **[P1] Clear the active flow state when locking.** This cancels or discards the underlying operations but leaves the published offer/presentation state, loading/status state, selections, completion flags, and pending continuation data intact. Because the wallet remains initialized, unlocking skips bootstrap and exposes stale state backed by cancelled sessions or discarded preview handles. Please mirror the Compose lock reset semantics, including invalidating navigation requests and clearing pending continuations, and cover locking during an active review or operation.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/PinScreen.kt
```diff
@@ -38,10 +46,19 @@ internal fun PinScreen(
         is WalletAuthState.Setup -> auth.error
         is WalletAuthState.Login -> auth.error
     }
+    val biometricAvailable = controller.isBiometricUnlockAvailable()
+    val biometricUnlockEnabled = controller.isBiometricUnlockEnabled()
+
+    LaunchedEffect(login != null) {
+        if (login != null) {
+            controller.unlockWithBiometrics()
```
szijpeter, line 62: **[P1] Do not automatically undo an app-initiated lock.** This automatic prompt also runs when the PIN screen reappears after Settings → Lock, because the fresh login state resets the prompt-consumed flag. Please suppress automatic prompting for app-initiated locks and apply the same policy to SwiftUI’s `onAppear`; the explicit biometric button should remain available.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/DemoPinStore.swift
```diff
@@ -0,0 +1,86 @@
...
+        return Self.derive(pin: pin, salt: salt) == expected
+    }
+
+    func clear() {
+        defaults.removeObject(forKey: recordKey)
+        defaults.removeObject(forKey: biometricKey)
+    }
+
+    private static func derive(pin: String, salt: Data) -> Data {
+        var hash = Data(SHA256.hash(data: Data(pin.utf8) + salt))
```
szijpeter: **[P2] Avoid a separate home-grown PIN verifier on native iOS.** Compose uses PBKDF2-SHA256 with 210,000 iterations, a 16-byte salt, a 32-byte verifier, and constant-time comparison, while this implements a custom 10,000-round hash and incompatible record format. Please align native iOS with the same standard contract, check random-generation failures, compare in constant time, and cover both implementations with independent parity vectors. This should remain demo-owned rather than moving into the mobile SDK.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/DemoBiometricAuthenticator.kt
```diff
@@ -0,0 +1,18 @@
+package id.walt.walletdemo.compose.logic
+
+interface DemoBiometricAuthenticator {
+    fun isAvailable(): Boolean
+    suspend fun authenticate(reason: String): DemoBiometricResult
+}
+
+enum class DemoBiometricResult {
```
szijpeter, line 8: **[P3] Keep the biometric contract to the behavior currently consumed.** Every caller handles failed, cancelled, and unavailable identically, while availability is already queried separately. Unless the UI needs to distinguish these outcomes now, please reduce this to success/failure and keep platform error mapping local; simplify the duplicated Swift contract likewise.

szijpeter (changes requested):
The demo-owned biometric boundary is sound, but Android’s off-main prompt, incomplete lock cleanup and automatic re-unlock, and the divergent native PIN verifier need correction before this is ready.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/PinScreen.kt
JakeFernandes98: Addressed in 32e501298.

App-initiated lock now sets `biometricPromptConsumed = true` on both Compose and native, so `LaunchedEffect` / `onAppear` do not auto-prompt after Settings → Lock. Cold launch / process start still uses a fresh login and auto-prompts. The **Unlock with biometrics** button still calls `unlockWithBiometrics(force = true)`. Covered by Compose controller + UI tests and a native unit test.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
JakeFernandes98: Addressed in 32e501298.

`WalletViewModel.lock()` now mirrors Compose: it idles status, clears offer/presentation/continuation/completion state, and increments the receive/present navigation reset keys, while still cancelling/discarding the in-flight handles. The native test starts an offer preview, locks, asserts the published review state is gone, then PIN-unlocks without a second bootstrap.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/androidMain/kotlin/id/walt/walletdemo/compose/logic/AndroidDemoBiometricAuthenticator.kt
JakeFernandes98: Addressed in 32e501298.

`authenticate()` now hops to `Dispatchers.Main.immediate` inside the Android adapter and only shows the prompt when the host `FragmentActivity` is at least `RESUMED`; otherwise it returns `Failed`. The adapter-boundary host test invokes `authenticate` from `Dispatchers.Default` against a resumed activity and asserts it completes without the main-thread `IllegalStateException`. Result may still be `Failed` when the host has no enrolled biometrics.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/DemoPinStore.swift
JakeFernandes98: Addressed in 32e501298.

Native `UserDefaultsDemoPinStore` now uses CommonCrypto PBKDF2-SHA256 with the same demo-owned record contract as Compose: version `1`, 210,000 iterations, 16-byte salt, 32-byte verifier, `1:<iterations>:<saltB64>:<verifierB64>`, and a constant-time compare. `setPin` fails if `SecRandomCopyBytes` is not `errSecSuccess`. Independent parity vectors are in both test trees (same PIN/salt/iterations/expected verifier), plus RNG-failure and truncated/wrong-version record coverage. Existing native demo PIN records from the previous hash format will not verify and need a reset/re-setup. Still demo-owned; not moved into the Wallet SDK.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/DemoBiometricAuthenticator.kt
JakeFernandes98: Addressed in 32e501298.

Kotlin and Swift `DemoBiometricResult` is now `Succeeded` / `Failed`. Android and iOS adapters still map cancel, not enrolled, and not-interactive locally to `Failed`. Enable-toggle and unlock callers persist or unlock only on `Succeeded`; any `Failed` leaves PIN fallback and unchecks the toggle. Fakes and the previous cancel/unavailable tests now use `Failed`.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
szijpeter: **[P2] Clear the complete pending continuation state.** `lock()` clears the published URL and HTML but leaves `pendingPresentationSuccessMessage`, which is the guard used by `completePresentationContinuation()` and `failPresentationContinuation()`. A callback already queued before locking can therefore restore a stale presentation result. Please call `clearPendingPresentationContinuation()` here—and in `resetWallet()`—and extend the lock test to prove late callbacks are inert.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosAppTests/DemoPinStoreTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        let defaults = UserDefaults(suiteName: "pin-parity-\(UUID().uuidString)")!
+        let store = UserDefaultsDemoPinStore(
+            walletID: "parity",
+            defaults: defaults,
+            randomSalt: { salt }
+        )
+        try await store.setPin(Self.parityPin)
+
+        XCTAssertEqual(defaults.string(forKey: "id.walt.walletdemo.pin.parity"), Self.parityRecord)
+        XCTAssertTrue(await store.verifyPin(Self.parityPin))
```
szijpeter: **[P1] Await before entering the XCTest assertion.** The native test target does not compile because XCTest assertion arguments are non-async autoclosures. Evaluate `store.verifyPin(...)` first and assert the resulting value. The same pattern occurs on lines 25, 54, 57, and 63, plus `WalletViewModelReceiveTests.swift:165` and `WalletViewModelPinTests.swift:181`.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/PinView.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+        .onChange(of: scenePhase) { phase in
+            if phase == .active {
+                promptBiometricsIfSceneActive()
+            }
+        }
+    }
+
+    private func promptBiometricsIfSceneActive() {
+        guard scenePhase == .active else { return }
```
szijpeter: **[P1] Refresh biometric availability independently of foreground prompting.** On a physical iPhone 14 running iOS 26.6, every callback reached this method with `scenePhase == .inactive`, so this guard skipped `refreshBiometricAvailability()` and no later active transition was delivered. The signup toggle therefore remains disabled even though LocalAuthentication reports biometrics as available. Please refresh the non-interactive availability state independently of the scene-gated automatic prompt and add fresh-signup regression coverage.

szijpeter (changes requested):
The previous Android, auto-unlock, PIN-verifier and biometric-contract findings are addressed, but native iOS tests do not compile, fresh signup cannot enable biometrics, and lock/reset retain one stale presentation-continuation token.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
JakeFernandes98: Done in `dc5421791`. `lock()` and `resetWallet()` now call `clearPendingPresentationContinuation()`, so the unpublished success-message guard is cleared with the URL/HTML. `testLockClearsPendingContinuationAndIgnoresLateCallbacks` and `testResetClearsPendingContinuationAndIgnoresLateCallbacks` complete/fail after lock or reset and assert the late callbacks stay inert.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosAppTests/DemoPinStoreTests.swift
JakeFernandes98: Done in `dc5421791`. Every `verifyPin` / `bootstrapCalls` value is awaited into a local first, then asserted. Same pattern in `DemoPinStoreTests`, `WalletViewModelReceiveTests`, and `WalletViewModelPinTests`.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/PinView.swift
JakeFernandes98: Done in `dc5421791`. `PinView` now refreshes availability on appear and every scene-phase change, and only the automatic biometric prompt stays gated on `.active`. `WalletViewModel` also refreshes availability during init so first-run signup does not depend on a later active transition. Covered by `testFreshSignupReportsBiometricAvailabilityWithoutSceneActivation`.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
szijpeter: **[P2] Reset still invalidates the continuation too late.** Lock is fixed, but `resetWallet()` calls `clearPendingPresentationContinuation()` only after awaiting `walletClient.deleteLocalData()`. A queued callback can still complete or fail the old presentation while deletion is suspended, and the token remains if deletion throws. Please clear it synchronously before starting the reset task and test delayed/failing deletion with an immediate late callback.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/PinView.swift
```diff
@@ -0,0 +1,115 @@
...
+            promptBiometricsIfSceneActive()
+        }
+        .onChange(of: scenePhase) { _ in
+            viewModel.refreshBiometricAvailability()
+            promptBiometricsIfSceneActive()
+        }
+    }
+
+    private func promptBiometricsIfSceneActive() {
+        guard scenePhase == .active else { return }
```
szijpeter: **[P2] Preserve cold-launch biometric auto-unlock.** A force-quit/relaunch creates a fresh `WalletViewModel` with `biometricPromptConsumed == false`, but on the physical iPhone `PinView` can appear while `scenePhase == .inactive` without receiving a later `.active` callback. This guard therefore prevents `promptBiometricUnlockIfNeeded()` from ever running, leaving native on the PIN screen while Compose auto-prompts. Please trigger the one-shot launch prompt from a reliable foreground lifecycle signal while preserving `lock()`’s consumed flag so Settings → Lock remains manual, and add cold-relaunch regression coverage.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosAppTests/WalletViewModelReceiveTests.swift
```diff
@@ -46,6 +47,7 @@ final class WalletViewModelReceiveTests: XCTestCase {
     func testChangingOfferClearsTransactionCodeState() async throws {
         let client = TransactionCodeWalletClient()
         let viewModel = WalletViewModel(walletClient: client)
+        viewModel.unlockForTests()
```
szijpeter, line 50: **[P3] Give the new PIN-gated tests a realistic deadline.** The exact-head `iosAppTests` run compiled, but `testChangingOfferClearsTransactionCodeState` and `testAuthorizationCodeOfferOpensIssuerSignInContinuation` timed out against this class’s two-second helper; both passed when rerun alone, and the current native CI job passed. The added unlock/bootstrap step still makes the local full suite flaky under load. Please use the elapsed 20-second deadline used by the other native suites, or avoid waiting through the gate in unrelated receive tests.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/PinScreen.kt
```diff
@@ -39,10 +51,22 @@ internal fun PinScreen(
         is WalletAuthState.Setup -> auth.error
         is WalletAuthState.Login -> auth.error
     }
+    var biometricAvailable by remember { mutableStateOf(false) }
+    val biometricUnlockEnabled = controller.isBiometricUnlockEnabled()
+
+    LaunchedEffect(Unit) {
```
szijpeter: **[P3] Refresh biometric availability when the app resumes.** This `LaunchedEffect(Unit)` snapshots availability only once for the lifetime of `PinScreen`. If biometric enrollment or permission changes in system Settings while the app remains alive, Compose continues showing the old toggle/button state; when it is stale-true, tapping the button silently does nothing because the controller rechecks availability. Please refresh on foreground/resume so both Compose platforms reflect the current state, and cover the availability transition.

szijpeter (approved):
Nice improvements. Some concerns remain, but nothing blocking imo; pls see below.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/WalletViewModel.swift
JakeFernandes98: Done in `e2855bda2`. `resetWallet()` now calls `clearPendingPresentationContinuation()` synchronously before starting the delete task, so a late complete/fail is inert even while `deleteLocalData()` is delayed or throws. `testResetClearsPendingContinuationBeforeDeleteLocalDataCompletes` and `testResetClearsPendingContinuationWhenDeleteLocalDataFails` cover those two races.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/PinView.swift
JakeFernandes98: Done in `e2855bda2`. The one-shot launch prompt now comes from the app becoming active (`WalletDemoApp` scene phase plus `UIApplication.didBecomeActiveNotification`), not from `PinView`'s scene-phase guard. `lock()` still sets `biometricPromptConsumed`, so Settings → Lock stays manual. `testColdLaunchAutoPromptsBiometricsWithoutActiveScenePhase` covers the fresh-VM path.

waltid-applications/waltid-wallet-demo-ios/iosApp/iosAppTests/WalletViewModelReceiveTests.swift
JakeFernandes98: Done in `e2855bda2`. `WalletViewModelReceiveTests` now uses the same 20-second `waitUntil` / `waitUntilAsync` deadline as the other native suites.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/PinScreen.kt
JakeFernandes98: Done in `e2855bda2`. `PinScreen` now reads availability from controller state and refreshes it when the window is focused, so both Compose platforms pick up enrollment/permission changes after resume. `refreshBiometricUnlockAvailabilityPicksUpEnrollmentChange` and `pinScreenRefreshesBiometricAvailabilityWhenItBecomesAvailable` cover the unavailable → available transition.

PR #2144, opened 2026-08-21
feat(wallet-demo): OpenID4VCI credential cards and compact DC API GET tray

JakeFernandes98:
## Summary

Persist and preview OpenID4VCI credential `display` (background, logo, text color) through the wallet SDK and Swift surface, then parse it into demo card-art models. Replace list-row credentials with ID-1 card art and an Apple Wallet-style stack on Compose and native iOS, including compact art on credential details and offered-credential review.

Host Digital Credentials GET review in a translucent compact tray after the system picker (Android) and keep Annex C review compact (iOS), without changing Cancel vs back outcomes. In-app Receive and Present use the same review chrome: hide the URL after preview, pin Accept/Share actions, restore verifier metadata, and collapse claim selection until a credential is tapped. Successful issuance goes to Credentials. Cancel, reject, and a finished presentation return to the default URL-entry page; there is no leftover New receive / New presentation lock.

Closes WAL-1327, Closes WAL-1291 as the GET-tray subtask, and Closes WAL-1317 for receive/share consistency.

This PR targets `main`. Settings, branding, credential copy/delete, dismissable status, PIN lock, and biometric unlock from [walt-id/waltid-identity#2142](https://github.com/walt-id/waltid-identity/pull/2142) and [walt-id/waltid-identity#2143](https://github.com/walt-id/waltid-identity/pull/2143) are already on `main`.

## What reviewers should inspect

- Display metadata is stored per credential from the offer, not re-fetched later from the issuer. Sidecar JSON already stored `issuerDisplay`; this PR also writes `credentialDisplay` (name, logo, colors, background image). Existing stored credentials without display keep default art until they are received again.
- wallet-api2 and enterprise wallet2 list endpoints still return `StoredCredentialMetadata` (id, format, issuer, subject, label, addedAt) and do not include the sidecar. GET-by-id still returns `StoredCredential.metadata` as an open `JsonObject`.
- Credentials-tab stack: tap a card to expand it in place and reveal details below; tap it again, the close control, or back to collapse. Test tags stay `wallet.credentialCard.<id>`. Copy and Delete stay on the expanded card so the main-line credential actions still work without a pushed details destination.
- Android GET still uses `Theme.WalletDemoCompose.Translucent` and `WalletDemoSharingReviewSheet`. Cancel still finishes with a provider cancellation; back/dismiss still returns to the system selector.
- The GMS / system picker still advertises the app icon. Android CredMan rows map title/subtitle from known credential claims, or a humanized type, and Android `replace()` may attach a per-record thumbnail. Card art itself is wallet UI only.
- In-app Receive/Present drop the URL editor after preview. Issuer/verifier identity is the tappable metadata box; technical requester, reader-trust, response-protection, and request details open from that box, not nested dropdowns.
- In-app Present lists credentials as checkbox cards and uses stored sidecar metadata for card art. Claims stay collapsed until the credential is opened. Compact DC GET uses the same claims dialog, not a generic credential-details screen.
- Settings include a DC API presentation-preview toggle (default on) that survives wallet reset.
- Successful stored issuance clears the offer and switches to Credentials. Decline, cancel, reject, and Present success all land on the default empty Receive or Present entry. Errors stay on the current tab.

## Architecture notes

- SDK sidecar JSON now includes `issuerDisplay` and `credentialDisplay` arrays with the OpenID4VCI visual fields. Mobile issuance previously stored a label plus issuer display; it now persists credential configuration display on immediate and deferred paths as well.
- Receive and Present share `ReviewScaffold` (Compose) and a matching safe-area inset (native iOS): details scroll, actions stay pinned.
- Compact GET still calls `SharingReviewSection` / `SharingReviewView` with `compact = true` for the card stack. Tapping a compact card opens the same claims dialog as in-app Present. Default credential selection is unchanged, so expanding a compact card does not change the Share payload.
- In-app Present uses the non-compact checkbox cards. Disclosure controls are the claims popup, not a separate details screen.
- Android Credential Manager provider and the compact sheet wrap the review in `WalletDemoTheme`.
- Settings, branding, PIN setup/unlock, and dismissable status banners already landed on `main` via 2142/2143 and stay in place. Native tabs keep the Settings toolbar; Compose Settings remains the header overlay.

## Caveats

- Default art uses the bundled walt.id logo on a constructed blue card.
- No device gallery or Play-services GET E2E evidence is attached. Do not treat this as merge-ready until current-head checks are green.
- Share still relies on existing key authorization (`BiometricTimedReuse`). There is no extra demo-PIN dialog when biometrics are unavailable. Demo unlock PIN/biometrics from 2143 is a separate gate.
- Image loads are HTTPS-only and fail closed. Background images outside about 1.2–2.0 aspect fall through to color/logo construction.
- Success status banners from 2142 still auto-hide after 4s. Compose iOS E2E waits for wallet-ready state rather than that banner text.
- Android `replace()` registers a local color/portrait thumbnail immediately, then refreshes remote HTTPS art under one shared 5 s budget. Presentation preview does not fetch artwork.

## Breaking changes

Yes — published mobile wallet types, no deprecated shims.

- `MobileWalletPresentationCredentialOption` adds `metadataJson` to the primary constructor and `copy`.
- `MobileWalletCredentialRegistryRecord` replaces `iconMetadataJson` / `iconCredentialDataJson` with typed `cardArtImageUris`, `cardArtBackgroundColor`, and `cardArtFallbackPng`.

List APIs and `StoredCredentialMetadata` are unchanged. Newly stored credentials may include extra keys on the existing opaque `metadata` `JsonObject` (`issuerDisplay` was already written; `credentialDisplay` is additive). Older stored credentials remain readable without those keys.


waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/api/waltid-openid4vc-wallet-mobile.klib.api
```diff
@@ -1597,7 +1605,8 @@ final class id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption { //
...
     final fun component8(): kotlin/String // id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption.component8|component8(){}[0]
     final fun component9(): kotlin.collections/List<id.walt.wallet2.mobile/MobileWalletPresentationDisclosure> // id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption.component9|component9(){}[0]
-    final fun copy(kotlin/String = ..., kotlin/String = ..., kotlin/Boolean = ..., kotlin/String = ..., kotlin/String? = ..., kotlin/String? = ..., kotlin/String? = ..., kotlin/String = ..., kotlin.collections/List<id.walt.wallet2.mobile/MobileWalletPresentationDisclosure> = ...): id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption // id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption.copy|copy(kotlin.String;kotlin.String;kotlin.Boolean;kotlin.String;kotlin.String?;kotlin.String?;kotlin.String?;kotlin.String;kotlin.collections.List<id.walt.wallet2.mobile.MobileWalletPresentationDisclosure>){}[0]
+    final fun component10(): kotlin/String? // id.walt.wallet2.mobile/MobileWalletPresentationCredentialOption.component10|component10(){}[0]
```
szijpeter: [P1] Regenerate the committed ABI dump. `:waltid-libraries:protocols:waltid-openid4vc-wallet-mobile:checkKotlinAbi` still fails at this head because the generated declaration sorts `component10()` before `component2()`, while this file puts it after `component9()`.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
+
+@Composable
+internal actual fun DefaultWaltLogo(modifier: Modifier) {
+    val context = LocalContext.current
+    val resId = remember(context) {
+        context.resources.getIdentifier("waltid_logo", "drawable", context.packageName)
```
szijpeter: [P1] Use the common Compose resource instead of resolving an Android drawable by name. The exact-head APK contains no `waltid_logo` drawable, so `resId` is always zero and Android silently omits the bottom-right logo; the common and Android PNGs are identical.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
```diff
@@ -0,0 +1,109 @@
...
+    issuerFallback: String,
+) {
+    var flipped by rememberSaveable(credential.configurationId) { mutableStateOf(false) }
+    val rotation by animateFloatAsState(
+        targetValue = if (flipped) 180f else 0f,
+        animationSpec = tween(durationMillis = 420),
+        label = "offerCardFlip",
+    )
+    val art = (credential.display ?: WalletDemoMetadataDisplay(
+        name = credential.vct ?: credential.doctype ?: credential.configurationId,
```
szijpeter: [P1] Do not fall back to the configuration ID for the card title. The issuance preview currently drops `vct`/`doctype`, so both Compose and native iOS show raw identifiers such as `org.iso...` whenever issuer display metadata omits a name. Preserve the standard credential type in the shared preview and resolve its friendly title through one shared SDK path.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/RegistryImageFetch.android.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.ktor.client.request.get
+import io.ktor.client.statement.bodyAsBytes
+import io.ktor.http.isSuccess
+
+internal actual suspend fun fetchRegistryIconBytes(url: String): ByteArray? {
+    if (!isHttpsUrl(url)) return null
+    return runCatching {
+        RegistryImageHttp.client.get(url).let { response ->
+            if (!response.status.isSuccess()) return@runCatching null
+            response.bodyAsBytes()
```
szijpeter: [P1] Bound remote artwork before buffering it and keep it off the consent path. `bodyAsBytes()` accepts an unbounded issuer-controlled response before the 2 MB check, and `registryRecords()` invokes this for every credential even during presentation preview; iOS then discards the produced icon. Move thumbnail enrichment to the Android adapter and enforce byte, pixel, and timeout limits.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/CredentialDisplayNormalizer.kt
```diff
@@ -24,16 +24,23 @@ object CredentialDisplayNormalizer {
 
     fun toDetails(summary: CredentialSummary): CredentialDetails {
         val issuerDisplay = StoredCredentialMetadataParser.issuerDisplay(summary.metadataJson)
+        val credentialDisplay = StoredCredentialMetadataParser.credentialDisplay(summary.metadataJson)
```
szijpeter: [P2] Pass the platform preferred locales here. Both new credential-display parsers accept locale preferences, but production Compose and Swift calls use the empty default, so the first metadata entry wins regardless of device language. This selection and normalization should also be shared rather than duplicated in Kotlin and Swift.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/SettingsScreen.kt
```diff
@@ -106,6 +109,43 @@ internal fun SettingsScreen(
                 copyTag = WalletUiTestTags.SettingsPublicJwkCopy,
                 onCopy = { text -> clipboard.setText(AnnotatedString(text)) },
             )
+            Column(
```
szijpeter, line 129: [P2] Update `lockDoesNotAutoPromptBiometrics` for the added settings height. Lock is now below the viewport; the test still clicks it without `performScrollTo()` and deterministically fails because the PIN screen never opens.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
```diff
@@ -2,104 +2,276 @@ import SwiftUI
...
+                    .allowsHitTesting(isSelected || !othersHidden)
+                }
             }
+            .frame(width: width, height: stackHeight, alignment: .top)
+            .clipped()
+            .animation(.spring(response: 0.42, dampingFraction: 0.86), value: selectedAtTop)
+            .animation(.easeInOut(duration: 0.22), value: othersHidden)
+        }
+        .frame(maxWidth: .infinity)
+        .frame(height: displayedHeight(forWidth: UIScreen.main.bounds.width - 40))
```
szijpeter: [P2] Derive the outer stack height from the `GeometryReader` width used above instead of `UIScreen.main.bounds`. In split view, landscape, or an embedded sheet these widths differ, producing clipping or excess whitespace.

szijpeter (changes requested):
Nice improvements overall—there are still several issues to address; please see the inline comments.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/api/waltid-openid4vc-wallet-mobile.klib.api
JakeFernandes98: Regenerated the wallet-mobile ABI dump with `updateKotlinAbi` (iOS klib). `component10()` now sorts before `component2()`, and `checkKotlinAbi` passes at `c4b809864`.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
JakeFernandes98: Android `DefaultWaltLogo` now uses the common Compose resource (`Res.drawable.waltid_logo`), same as iOS and wasm. It no longer looks up an APK drawable by name.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
JakeFernandes98: Issuance preview now keeps `vct`/`doctype` from the offered configuration. Compose and native resolve the card title through `CredentialTitles` (display name, then friendly type, then format) and no longer fall back to the configuration ID.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/RegistryImageFetch.android.kt
JakeFernandes98: `registryRecords()` no longer fetches or builds thumbnails, so presentation preview stays off the network. Android `replace()` resolves icons with HTTPS-only fetches that cap bytes, pixels, and timeout; iOS still ignores the icon.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/CredentialDisplayNormalizer.kt
JakeFernandes98: Production Compose and Swift now pass platform preferred locales into the credential-display parsers. Selection lives in shared `DisplayLocales` (Kotlin + WalletSDK Swift); the Compose web target keeps an aligned copy because digital-credentials has no wasm artifact.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/SettingsScreen.kt
JakeFernandes98: `lockDoesNotAutoPromptBiometrics` now `performScrollTo()` before clicking Lock, matching the other settings-height tests.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
JakeFernandes98: The stack now measures its `GeometryReader` width and derives height from that width. It no longer uses `UIScreen.main.bounds`.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
```diff
@@ -839,7 +836,11 @@ object WalletIssuanceHandler {
                 val entry = wallet.parseAndStore(
                     issuedCredential,
                     label = offeredCredential.configuration.credentialMetadata?.display?.firstOrNull()?.name,
-                    metadata = metadata,
+                    metadata = storedCredentialDisplayMetadata(
```
JakeFernandes98, line 839: To see if this is a breaking change for stored credentials

JakeFernandes98:
Crashes when receiving a credential. Localised issue for this PR only

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialRegistry.kt
```diff
@@ -152,7 +155,8 @@ public class AndroidDigitalCredentialRegistry(
             creationRegistrationAvailable = false
             return MobileWalletCredentialRegistrationResult(false, 0, "Credential Manager requires API 23")
         }
-        val entries = records.map { it.toAndroidEntry() }
+        val enriched = records.map { it.withResolvedThumbnail() }
```
szijpeter: [P2] `replace()` resolves every record’s thumbnail sequentially before registration. Each record may try background art, credential logo, and issuer logo with a 5 s timeout each, while wallet bootstrap, successful issuance, and deletion await this synchronization; unreachable issuer art can therefore delay the critical path by up to 15 s × credential count and is fetched again on later syncs. Register a local fallback synchronously and cache/refresh remote thumbnails outside that path, or use one bounded concurrent budget for the entire refresh.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
```diff
@@ -2,104 +2,284 @@ import SwiftUI
...
+        Set(uris.compactMap { $0 }).forEach { uri in
+            group.addTask { _ = await loadMetadataArt(from: uri) }
+        }
+    }
+}
+
+private func loadMetadataArt(from value: String?) async -> UIImage? {
+    guard let url = httpsURL(value) else { return nil }
+    do {
+        let (data, _) = try await URLSession.shared.data(from: url)
```
szijpeter: [P2] This loads issuer-controlled metadata with `URLSession.shared.data(from:)` and decodes it using `UIImage(data:)` without checking HTTP status or enforcing response-byte, decoded-pixel, or explicit overall limits. `WalletViewModel` also awaits `prefetchCredentialCardArt` before publishing `offerPreview`, so slow or oversized artwork can hold the consent screen and exhaust memory. Publish constructed art immediately, load remote art asynchronously, and apply the same bounded-image policy already used on Android.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletPresentationModels.kt
```diff
@@ -194,6 +196,7 @@ public data class MobileWalletPresentationCredentialOption(
     val label: String?,
     val credentialDataJson: String,
     val disclosures: List<MobileWalletPresentationDisclosure> = emptyList(),
+    val metadataJson: String? = null,
```
szijpeter, line 199: [P2] Adding `metadataJson` to this public primary constructor replaces the published JVM constructor and `copy` signature; Kotlin defaults do not retain the old binary entry points. `MobileWalletCredentialRegistryRecord` changes the same way, while the PR currently says “Breaking changes: None.” If mobile binary compatibility is intentionally out of scope, declare the break rather than adding compatibility overloads; otherwise reshape the additive API.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletDigitalCredentials.kt
```diff
@@ -105,7 +114,43 @@ public data class MobileWalletCredentialRegistryRecord(
     public val type: String,
     public val fields: List<MobileWalletCredentialRegistryField>,
     public val displayName: String,
-)
+    public val subtitle: String = "",
+    public val iconPng: ByteArray? = null,
```
szijpeter, line 120: [P2] `iconPng`, `iconMetadataJson`, and `iconCredentialDataJson` are public common-model fields consumed only by the Android registry adapter; the last field also duplicates the full credential JSON alongside the matcher fields. That grows the KMP API and couples common registry records to one platform’s thumbnail plumbing. Expose typed, platform-neutral display metadata if it is an SDK contract, or keep the Android enrichment internal to the Android adapter—without adding compatibility shims merely to preserve this shape.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/CredentialCardArt.kt
```diff
@@ -0,0 +1,191 @@
...
+        -> true
+        CredentialCardMetadataArtState.Pending,
+        CredentialCardMetadataArtState.Ready,
+        -> false
+    }
+
+@Composable
+internal expect fun DefaultWaltLogo(modifier: Modifier)
+
+internal fun parseCssColor(value: String?): Color? {
```
szijpeter, line 179: [P2] OID4VCI defines `background_color` and `text_color` as numerical CSS Color Level 3 values, which includes `rgb()`, `rgba()`, `hsl()`, and `hsla()` as well as hex. This hex-only parser silently falls back for those valid forms, while near-identical parsing is copied into Android registry thumbnails and Swift; the extra eight-digit hex branch is not part of the referenced CSS3 grammar. Normalize the protocol color value once at the display-metadata boundary and feed typed channels to each renderer. ([OpenID4VCI 1.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html), [CSS Color Level 3](https://www.w3.org/TR/2022/REC-css-color-3-20220118/))

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
```diff
@@ -133,6 +147,106 @@ fun WalletDemoSharingReviewScreen(
...
+
+    SystemBackHandler(enabled = true) {
+        when {
+            openDetails != null -> openCredentialDetailsId = null
+            !enabled -> Unit
+            else -> onBackAtRoot()
+        }
+    }
+
+    MaterialTheme {
```
szijpeter: [P2] The standalone DC API provider activity wraps the review screen in a bare `MaterialTheme`, so it gets Material’s default purple primary color instead of the blue wallet theme used by the same review UI in-app. Use `WalletDemoTheme` (or inject the same color scheme) at this entry point.

Android DC API provider review with an inconsistent purple Share button

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/CredentialDetailsScreen.kt
```diff
@@ -43,55 +38,39 @@ internal fun CredentialDetailsScreen(
     val rawCredential = details.summary.credentialDataJson?.takeIf { it.isNotBlank() }
         ?: "No raw credential available"
 
-    Column(
+    Box(
```
szijpeter: [P2] Removing Delete from the presentation consent flow is a sensible boundary, but the old path was only partially removed. Compose’s `WalletRoute` / `WalletTabNavDisplay` and native iOS’s `CredentialDetailsDestination` / `CredentialDetailsScreen` now only reference each other and have no external callers (the Swift files remain compiled). `CredentialDetailsChromeOverlay` still exposes copy/delete callbacks that its sole caller never supplies. The UI test named `presentationDetailsDeleteUsesStoreCredentialId` no longer opens presentation details either; it switches to Credentials and deletes there. Remove the dead callbacks/navigation files/project entries and rename that still-useful test to describe deleting from Credentials while a review is active.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
```diff
@@ -133,6 +147,106 @@ fun WalletDemoSharingReviewScreen(
...
+        ) {
+            ModalBottomSheet(
+                onDismissRequest = {
+                    if (enabled) onBackAtRoot()
+                },
+                sheetState = sheetState,
+            ) {
+                Box(modifier = Modifier.exportTestTagsForPlatformAutomation()) {
+                    if (openDetails != null) {
+                        CredentialDetailsScreen(
```
szijpeter: [P2] These two presentation paths now open different concepts for the same credential. Normal in-app sharing uses `SharingClaimsDialog`, showing the issuer and requested disclosures relevant to consent; compact DC API sharing opens the generic `CredentialDetailsScreen`, showing the card plus all stored credential groups and different chrome. Swift mirrors the same split between its claims sheet and generic `CredentialDetailsView`. The host container can remain platform-specific, but the presentation-details content should be shared: resolve the same title and render the same issuer/requested-disclosure model in both, while injecting dialog/sheet navigation and selection controls as needed. The platform transport does not require this semantic divergence.

| Normal in-app presentation | Compact DC API presentation |
|---|---|
| Normal presentation claims dialog showing the requested disclosures | DC API presentation opening the generic full credential details view |

szijpeter (approved):
Great work, @JakeFernandes98 ! 👏🏼 

I still see a few inconsistencies and issues, but I wouldn’t block on any of them. Please consider them, and if you agree, feel free to address them here or in a follow-up—I’ll leave that up to you.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
szijpeter: The missing-logo symptom is fixed, but the implementation still ships three byte-identical 252 KB PNGs in the Compose app (`androidApp` drawable, `sharedUI` Android drawable, and common Compose resources) and probes drawable, asset, then classloader paths at runtime. Please keep one common packaged Compose resource and make the Android renderer consume that source; the native Swift package can retain its own language/package resource.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
szijpeter: The current head still does not use one resolved title throughout a flow. In Compose issuance, the same live mDL is `org.iso.18013.5.1.mDL` on review and `Mobile Driving Licence` after receipt because the non-card-first branch in `OfferReviewSection` builds `CredentialCardArt` directly instead of using `resolvedCardTitle()`. In an ordinary (non-DC API) presentation, the card says `Mobile Driving Licence`, but opening it shows `mso_mdoc` because the claims dialog renders raw `option.label`; Swift has the same raw-label fallback in its claims view. Resolve the title once in each demo’s adapter/view-model boundary and pass it to offer, card, and detail views; protocol-level issuer-display selection can remain shared without moving demo UI policy into the SDK.

| Live flow | First view | Following view |
|---|---|---|
| Compose issuance | Compose mDL issuance review titled org.iso.18013.5.1.mDL | The received Compose mDL card titled Mobile Driving Licence |
| Ordinary Compose presentation | Presentation card titled Mobile Driving Licence | Opened presentation claims dialog titled mso_mdoc |

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
szijpeter: The width-based height calculation did not fix the live multi-card layout. Two cards already lose the lower part of the second card; with three, only the title strip of the third remains. The offset cards extend below the effective container while the final `.clipped()` removes the remainder. Please size the stack for the complete visible offsets in the actual selected/expanded state, or remove the clipping boundary once the parent owns the full height.

| Two credentials | Three credentials |
|---|---|
| SwiftUI credential stack clipping the lower part of the second card | SwiftUI credential stack showing only the title strip of the third card |

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
JakeFernandes98: Landed on 969a9fc87: Compose now ships one `waltid_logo.png` in common resources. Android decodes that Compose resource (assets, then classloader) and no longer probes a drawable or keeps the two extra copies.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
JakeFernandes98: Landed on 969a9fc87: Receive review and presentation claims now use the same `resolveCardTitle` path as stored cards. A nameless mDL offer and a presentation option labeled `mso_mdoc` both resolve to Mobile Driving Licence.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
JakeFernandes98: Landed on 969a9fc87: the stack frame is now `offsets.last + cardHeight` for the current selected/expanded state, and the extra `.clipped()` is gone so the 2nd/3rd peek cards are not cropped.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialRegistry.kt
JakeFernandes98: Landed on 969a9fc87: `replace()` registers a local color/portrait thumbnail immediately, then refreshes remote HTTPS art concurrently under one shared 5s budget. Bootstrap/issue/delete no longer wait on sequential 5s fetches per URI.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
JakeFernandes98: Landed on 969a9fc87: offer preview is shown first; art prefetch runs in the background. The fetch now requires HTTPS, an HTTP success status, a 2 MB byte cap, a decoded-pixel cap, a 5s timeout, and still fail-closes on aspect outside 1.2–2.0.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletPresentationModels.kt
JakeFernandes98: Agreed this is a published constructor/`copy` break. No deprecated overloads. The PR body now lists `MobileWalletPresentationCredentialOption.metadataJson` as a binary break.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletDigitalCredentials.kt
JakeFernandes98: Landed on 969a9fc87: the Android-only JSON blobs are gone from the common record. Registry records now carry typed `cardArtImageUris`, `cardArtBackgroundColor`, and optional `cardArtFallbackPng`. ABI dump updated; PR body declares the break, no shims.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/CredentialCardArt.kt
JakeFernandes98: Landed on 969a9fc87: one CSS Color Level 3 parser (`CssColors`) now feeds Compose, registry swatches, and Swift. It accepts `#rgb`, `#rrggbb`, `rgb()`, `rgba()`, `hsl()`, `hsla()`, and `transparent`. 8-digit hex is rejected.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
JakeFernandes98: Landed on 969a9fc87: both `WalletDemoSharingReviewScreen` and `WalletDemoSharingReviewSheet` wrap in `WalletDemoTheme`, so the DC tray uses the same blue primary as in-app Share.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/CredentialDetailsScreen.kt
JakeFernandes98: Landed on 969a9fc87: unused `WalletRoute` / `WalletTabNavDisplay` and native `CredentialDetailsDestination` / `CredentialDetailsScreen` are deleted. The leftover test is renamed to `deleteFromCredentialsWhileAReviewIsActive`.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
JakeFernandes98: Landed on 969a9fc87: compact DC GET now opens the same claims dialog as in-app Present (Compose and native iOS). It no longer pushes generic credential details.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
szijpeter: Still reproducible on exact head `969a9fc8`. The Android loader asks assets and then the classloader for `composeResources/.../waltid_logo.png`, but that path is absent from the exact-head APK, so `loadWaltLogoBitmap()` returns `null` and the logo is omitted. Please consume the generated common resource through `painterResource(Res.drawable.waltid_logo)` instead of resolving its packaging path manually.

| Compose received card | DC API provider |
|---|---|
| Compose Android received Mobile Driving Licence card without the bottom-right logo | Android DC API provider review with the bottom-right card logo absent |

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
szijpeter: The public mDL path is now consistent: issuance, stored card, and presentation all show `Mobile Driving Licence`. One code path is still incomplete, though. `MobileWalletPresentationCredentialOption.label` is the optional display label stored with the credential, but both presentation adapters now discard it and call `resolveCardTitle(... fallback = option.format)`; the common `resolvedCardTitle()` repeats the same projection. When metadata and payload do not yield a friendlier title, this falls through to raw `mso_mdoc`/`vc+sd-jwt` instead of the SDK-provided label. Please use the non-blank option label as the presentation fallback before the raw format.

| Issuance | Received | Presentation |
|---|---|---|
| Compose Android issuance review titled Mobile Driving Licence | Compose Android received card titled Mobile Driving Licence | Compose Android presentation review titled Mobile Driving Licence |

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/CredentialCardArt.kt
szijpeter: The KMP parser is a good consolidation for Compose and Android registry swatches, but Swift still has an independent `CssColorChannels` / `parseCssColor` implementation in `CredentialCardView.swift`; there is no Swift call into `CssColors` or typed KMP color channels. The two copies also share the same permissive deviations from the declared CSS Color 3 grammar: bare hex without `#`, prefix-based function matching, and ignored trailing text after `)`.

~~Please finish the original boundary by exposing typed channels from the shared display model to both renderers.~~

**Update after rechecking the Swift boundary:** `CssColors` is not exported through `WalletCore`, and exposing it here would create a new public SDK surface solely for these demos. Keeping the small Swift parser is reasonable in this PR. Please just keep both implementations aligned with the documented grammar, or explicitly document and test the accepted superset. A future SDK-level normalization could expose typed colors consistently for issuance, presentation, and stored metadata together.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/CredentialDetailsScreen.kt
szijpeter: The main dead destinations are gone, but the cleanup still leaves one coherent obsolete detail/presentation slice:

- Compose `CredentialDetailsScreen` and Swift `CredentialDetailsChromeOverlay` have no callers.
- Every production caller passes `{}` to `SharingReviewSection.onCredentialClick`, while Swift stores/documents `SharingReviewView.onCredentialSelected` but never reads it.
- After removing the standalone screen, every live Compose and native details caller disables the embedded card, so the `showCard` and card-click branches are unreachable; the expanded Credentials stack already owns the card.
- `CredentialSummary.credentialId` and Swift `CredentialDetails.credentialId` have no production reads; presentation selections already retain their store IDs.
- Swift `WalletAccessibilityID.credentialDetails(_:)` has no caller, and the stored `CredentialDetails.metadataJSON` is not read after its display projections are parsed.

The live Credentials-tab details UI, shared claims content, copy/delete controller APIs, close/menu components, root details tag, and presentation claims dialogs should remain.

waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/CredentialCardView.swift
szijpeter: Confirmed fixed on exact head `969a9fc8`: a fresh three-credential run shows the complete third card without clipping.

Native iOS presentation review showing three complete credential cards without clipping

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
szijpeter: Confirmed fixed on exact head `969a9fc8`: the standalone provider review now uses the wallet blue Share action.

Android DC API provider review showing the wallet blue Share button

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/WalletDemoSharingReviewScreen.kt
szijpeter: Confirmed fixed on exact head `969a9fc8`. Normal presentation and the DC API provider now open the same `Mobile Driving Licence` claims content with the same issuer and requested disclosures.

| In-app presentation | DC API provider |
|---|---|
| Compose Android in-app Mobile Driving Licence claims with requested disclosures | Android DC API provider Mobile Driving Licence claims with matching requested disclosures |

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialRegistry.kt
```diff
@@ -152,6 +157,29 @@ public class AndroidDigitalCredentialRegistry(
...
+        val refreshed = withTimeoutOrNull(RegistryIconFetchTimeoutMs) {
+            coroutineScope {
+                records.map { record -> async { record.withResolvedThumbnail() } }.awaitAll()
+            }
+        } ?: return result
+        val changed = refreshed.indices.any { index ->
+            !refreshed[index].iconPng.contentEquals(immediate[index].iconPng)
+        }
+        if (changed) {
+            registerRecords(registryId, refreshed)
```
szijpeter: [P2] Keep the returned result and registry capability state consistent after this refresh. If the initial local-art registration succeeds but this refreshed-art registration fails, `registerRecords` sets `registrationAvailable = false`, while `replace` discards that failure and returns the original success. A best-effort art refresh should either preserve the successful initial registration/capability or return the refreshed failure; it should not report success while `capabilities()` reports registration unavailable.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletRegistryIcons.kt
```diff
@@ -0,0 +1,267 @@
...
+            bytes
+        }
+    }
+    is JsonPrimitive -> {
+        val raw = contentOrNull?.trim().orEmpty()
+        if (raw.isEmpty()) {
+            null
+        } else {
+            val encoded = raw.substringAfter("base64,", raw)
+            runCatching { Base64.decode(encoded) }.getOrNull()
```
szijpeter: [P2] Enforce `MaxRegistryIconBytes` before decoding this issuer-controlled Base64 string. The JSON-array branch bounds allocation first, but this branch decodes the entire value and only later rejects oversized bytes in `isImageBytes`; a credential can therefore force a decoded allocation far beyond the intended 2 MB icon limit. Reject an encoded length above the largest Base64 representation of the limit before either decoder runs.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/androidMain/kotlin/id/walt/walletdemo/compose/ui/components/DefaultWaltLogo.android.kt
JakeFernandes98: Landed on `9e9742040`. The missing mark was packaging, not just the loader: Android KMP libraries leave asset processing off, so `copyAndroidMainComposeResourcesToAndroidAssets` had no output and `waltid_logo` never reached the APK. `sharedUI` now enables Android resources; `DefaultWaltLogo` uses `painterResource(Res.drawable.waltid_logo)` like iOS/wasm.

Proof at this head:
- debug APK lists `assets/composeResources/id.walt.walletdemo.compose.ui.resources/drawable/waltid_logo.png`
- host UI test `CredentialCardLogoAndroidTest.constructedCardShowsBundledWaltLogo`
- Pixel 9 Play Store emulator: constructed-art receive preview, stored Credentials card, and DC GET provider tray (`Share digital credential?`) all show the bottom-right walt.id mark

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/FlippableOfferCard.kt
JakeFernandes98: Landed on `9e9742040`. `presentationDisplayName(..., storedLabel)` was already on HEAD; this adds the missing case that would have failed on `969a9fc8`: empty metadata, untyped payload, `label = "Personal ID"`, `format = "mso_mdoc"` → title is `Personal ID`.

- Compose: `MobileWalletIssuanceMappingTest.presentationOptionUsesStoredLabelWhenMetadataAndPayloadHaveNoTitle`
- Swift: `CredentialDisplayNormalizerTests.testPresentationOptionUsesStoredLabelWhenMetadataAndPayloadHaveNoTitle`
- Claims keep `details.cardSummary.title` / `resolvedCardTitle()`

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/CredentialCardArt.kt
JakeFernandes98: Landed on `9e9742040`. Kotlin `CssColors` and a demo-local Swift `CssColorParser` now share the CSS Color Level 3 grammar: `#rgb` / `#rrggbb`, exact `rgb()`/`rgba()`/`hsl()`/`hsla()`, `transparent`. Bare hex, prefix function names, trailing text after `)`, and 8-digit hex are rejected. `CssColors` is not exported through WalletCore.

- `CssColorsTest` (accept + reject vectors)
- `CssColorParserTests` (same vectors)

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/CredentialDetailsScreen.kt
JakeFernandes98: Landed on `9e9742040`. Deleted unused `CredentialDetailsScreen`, Swift `CredentialDetailsChromeOverlay`, no-op `SharingReviewSection.onCredentialClick` / `SharingReviewView.onCredentialSelected`, and `WalletAccessibilityID.credentialDetails(_:)`.

Kept `credentialId`, Swift `metadataJSON`, `showCard = false` callers, and the root `wallet.credentialDetailsScreen` tag used by the live Credentials-tab details.

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/androidMain/kotlin/id/walt/wallet2/mobile/AndroidDigitalCredentialRegistry.kt
JakeFernandes98: Landed on `9e9742040`. A best-effort art refresh no longer clears a successful initial registration: `replace()` sets `registrationAvailable = initial || refresh`, so capabilities stay aligned with the returned success.

- `AndroidDigitalCredentialRegistryTest.bestEffortRefreshDoesNotClearSuccessfulInitialRegistration`

waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletRegistryIcons.kt
JakeFernandes98: Landed on `9e9742040`. The issuer Base64 icon branch now rejects an encoded length above the Base64 size of `MaxRegistryIconBytes` before `Base64.decode`.

- `MobileWalletRegistryIconsTest.rejectsOversizedBase64PortraitBeforeDecoding`

szijpeter (approved):

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/screens/CredentialDetailsScreen.kt
szijpeter: Thanks—rechecking the items deliberately kept, the root `wallet.credentialDetailsScreen` tag is live and should remain. The others still have no production read or caller after the presentation-details delete flow was removed: the details-only store IDs and stored Swift `metadataJSON` are write-only, while every live caller disables the embedded card, making its tap callback unreachable. [`87f18289de`](https://github.com/walt-id/waltid-identity/commit/87f18289dedcffe4d641f755a2cd3ae535d1ea4e) contains a focused cleanup while retaining the live tag and details UI.

waltid-applications/waltid-wallet-demo-compose/sharedUI/src/commonMain/kotlin/id/walt/walletdemo/compose/ui/components/CredentialCardArt.kt
szijpeter: One small correctness edge case remains: both parsers accept `NaN` and infinities as numbers. That can produce invalid Swift color channels, and Kotlin throws when `NaN` reaches `roundToInt()`. [`a0276dd496`](https://github.com/walt-id/waltid-identity/commit/a0276dd4960a6c161440a308f71d629e97c366f6) adds finite-value checks and matching tests if helpful.

PR #2152, opened 2026-08-26
feat: wal-1338 X509CertificateUtil: binary copy subject DN from CSR to certificate

hokorn:
### X509CertificateUtil:
* Extended certificate builder, so it is possible to copy subject DN raw value from CSR to certificate
* Extended certificate builder, so subject public key info of CSR can be used to create SPKI of certificate
* Extendend X509CertificateUtil to be able to create CSR with crypto2 keys

## Type of Change

## Checklist

## Breaking

- \-

## Summary by CodeRabbit

- **New Features**
  - Added CSR and certificate signing support using modern cryptographic keys with explicitly selected signature algorithms.
  - Added support for raw subject distinguished names and subject public-key information when creating certificates.
  - Certificate creation can now reuse subject details from a CSR.
- **Bug Fixes**
  - Added validation to prevent incompatible keys and signature algorithms from being used together.
  - Improved handling of certificate subjects and optional extension requests.
- **Deprecations**
  - Existing legacy-key signing methods remain available but are now deprecated.
- **Tests**
  - Added coverage for creating and validating CSR-based certificate chains.


philpotisk:
## Code review findings

Automated review (medium effort) of this diff. 4 findings, most severe first.

### 1. `createCertificate(issuerKey: Crypt1Key, ...)` never resets `builder.issuerDnRaw` after `block.invoke(builder)` — unlike its sibling overload
**File:** `waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/X509CertificateUtil.kt` (around the `createCertificate(issuerKey: Crypt1Key, issuerCert, block)` overload)

The sibling `createCertificate(issuerKey: Key, issuerCert, signatureAlgorithm, block)` defensively re-sets `builder.issuerDnRaw = issuerCert.data.subjectDnRaw` right after `block.invoke(builder)`, so a caller-supplied block cannot corrupt the issuer DN. The `Crypt1Key` overload — touched by this same diff, with new `require()`s added a few lines below — has no such reset. If a caller's block sets `builder.issuerDnRaw` (a public `var`, e.g. by mistake or copy-pasted from another cert-building block), the resulting certificate is signed with an issuer DN that doesn't match the actual issuing certificate's subject DN, silently violating chain-building assumptions.

### 2. Bouncy vs Signum diverge on empty-extensions CSR encoding
**File:** `waltid-libraries/crypto/waltid-x509/src/jvmBouncyMain/kotlin/id/walt/certificate/x509/bouncycastle/BouncyPkcs10CertificateSigningRequestSigner.kt`

`buildBouncyCsr` now wraps the extensionRequest attribute generation in `if (csrBuilder.requestedCertificate.extensions.isNotEmpty())`, so a CSR built with zero extensions on the JVM/Bouncy backend carries no extensionRequest PKCS#9 attribute at all. `SignumCertificateSigner.buildTbsCsr` (used on other targets) unconditionally sets `extensions = extensions` on `TbsCertificationRequest` even when the list is empty. For a multiplatform library this means the DER encoding of a CSR with no extensions now differs by platform/signer backend. No test exercises the empty-extensions case for both signers to confirm equivalent behavior.

### 3. New exclusivity `require()` messages cite the wrong condition
**File:** `waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/X509CertificateUtil.kt`

`require(subjectPublicKeyInfo.crypto1key == null) { "For subject public key info key or SPKI must be set" }` and its mirror `require(subjectPublicKeyInfo.key == null) { "For subject public key info Crypto1Key or SPKI must be set" }` both reuse the sibling require's message text, which describes a "must be set" condition rather than the actual violated invariant ("crypto1key/key must NOT be set together with the other key type"). If either check ever fires, the message would misdirect debugging.

### 4. New shared `profileDocumentSignerCertificate()` helper only wired into one of two twin overloads
**File:** `waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/profile/IsoDocumentSignerX509CertificateProfile.kt`

The diff extracts validity/`extensionSubjectKeyIdentifier`/`extensionKeyUsage`/`extensionExtendedKeyUsage` setup out of `profileDocumentSignerCertificate(..., subjectKey: Crypto1Key, subjectDn: String)` into a new no-arg `profileDocumentSignerCertificate()` helper, but the sibling `profileDocumentSignerCertificate(..., subjectKey: Key, subjectDn: String)` (crypto2, untouched by this PR) still duplicates the identical logic inline. A future change to the DS certificate's validity period or key-usage extensions applied only to the shared helper will silently leave the crypto2 `Key`-based issuance path on stale settings.

---
🤖 Posted by an automated review pass.

philpotisk (approved):

PR #641, opened 2024-07-23
Feat/e2e auth api

mikeplotean:
## Description

Adds keycloak auth-api tests.

Fixes #642 

## Type of Change

- [ ] ~~Bug fix (non-breaking change which fixes an issue)~~
- [ ] ~~New feature (non-breaking change which adds functionality)~~
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
- [ ] ~~This change requires a documentation update~~

## Checklist

- [x] I have performed a self-review of my code
- [ ] ~~If it is a core feature, I have added thorough tests~~
- [ ] ~~Do we need to implement analytics?~~
- [ ] ~~Will this be part of a product update? If yes, please write one phrase about this update.~~
- [ ] My code follows the style guidelines of this project
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have made corresponding changes to the documentation~~
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #2154, opened 2026-08-26
[WAL-1343] feat(wallet): bind mdocs to holder keys

szijpeter:
## Summary

- Persist a typed, versioned holder-key binding with an opaque provider-qualified reference and RFC 7638 SHA-256 public-key identity.
- Carry the exact Crypto2 proof key through immediate, deferred, batch, and isolated issuance storage paths; imported credentials discard incoming bindings and bind only on one unique verified local match.
- Keep the durable binding operation-neutral: it identifies the MSO DeviceKey, while each presentation route explicitly requests the operation it performs. Existing OpenID4VP and Annex C routes require `SIGN`; proximity can resolve the same binding for `SIGN` or `KEY_AGREEMENT`.
- Discover candidate identities through immutable public JWK material, then perform operational lookup only after a binding and explicit usage have been selected.
- Resolve and revalidate the credential-bound key for OpenID4VP and Annex C presentation, with no wallet-default, legacy-key, or software-key fallback.
- Provision the OpenID4VP conformance mdoc fixture through the same Crypto2 wallet key store and assert that the imported credential is bound, so conformance exercises the production holder-key path.

Enterprise persistence and KMS counterpart: [walt-id/waltid-identity-enterprise#641](https://github.com/walt-id/waltid-identity-enterprise/pull/641).

## Persistence and failure semantics

- Mobile uses the new SQLDelight schema directly. The unpublished SDK has no migration, compatibility model, or legacy-key path.
- Published OSS server persistence evolves additively through a nullable binding column; existing rows remain unbound.
- Dedicated issuance and import binding paths make the persisted origin structural rather than caller-selected.
- Holder-key eligibility is evaluated after DCQL alternatives, optional sets, and transaction-data requirements; only a required credential set actually blocked by an unusable binding fails.
- A credential can be correctly bound yet ineligible for a route whose requested operation it cannot perform; for example, a MAC-capable `KEY_AGREEMENT` key is intentionally not eligible for the existing signature-only OpenID4VP and Annex C routes.
- Missing, corrupt, stale, ambiguous, mismatched, unsupported, unauthorized, or unavailable binding/key failures remain local and typed before mdoc authentication.
- Public-key identity discovery cannot expose or authorize an operational private-key handle. Provider availability, usage, and authorization are evaluated only for the selected route operation.
- Configured key-provider order is part of the resolver namespace; the persisted thumbprint makes configuration drift fail closed.

## Related

- [WAL-1343](https://linear.app/walt-new/issue/WAL-1343/bind-stored-mdocs-to-their-holder-keys)
- [Architecture contract](https://github.com/walt-id/waltid-architecture/pull/60)

## Follow-up boundary

Proximity session establishment, device-MAC construction, radio transports, SDK session APIs, and demo UI remain in the later WAL-1335 children. This PR retains the operation-neutral binding and `KEY_AGREEMENT` resolution foundation those children use.

## Breaking changes

- The mobile persistence schema and shared experimental wallet model change intentionally without a mobile migration or compatibility layer because the SDK is not published. Existing published server rows remain readable and unbound.
- Isolated issuance callers that store an mdoc must pass the exact original proof/device `keyId` in `FetchCredentialRequest` or `PollDeferredRequest`; omitting it fails closed instead of guessing a wallet key.
- Low-level mdoc presentation through `WalletPresentFunctionality2` must provide `mdocHolderKeyResolver`; mdocs no longer fall back to a wallet-default key.


waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/data/HolderKeyBinding.kt
```diff
@@ -0,0 +1,606 @@
...
+            val keyId = value.removePrefix("static:").decodeFromBase64Url().decodeToString().takeIf { it.isNotBlank() }
+                ?: return@runCatching null
+            WalletKeyLocation.Static(keyId)
+        }
+
+        else -> null
+    }
+}.getOrNull()
+
+private val MDOC_PRESENTATION_USAGE_ALTERNATIVES = listOf(
```
JakeFernandes98, line 600: **Major:** Import now treats `SIGN` or `KEY_AGREEMENT` as a presentable holder key, but every shipped eligibility and presentation path still calls `resolveHolderKey` with the default `SIGN` usage (`WalletPresentationHandler` matching / `mdocHolderKeyResolver`, Annex C preview and submit).

A MAC-only import therefore persists a binding and then fails remote matching with `KEY_USAGE_UNSUPPORTED`. [Architecture #60](https://github.com/walt-id/waltid-architecture/pull/60) still says import binds exactly one local Crypto2 **signing** key and leaves zero or multiple matches unbound.

Either keep import signing-only until WAL-1344 / WAL-1335 request `KEY_AGREEMENT`, or make eligibility use these same alternatives and update the contract. Do not leave “bound but ineligible” as the OpenID4VP outcome for a successful import.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
```diff
@@ -1317,27 +1395,187 @@ object WalletPresentationHandler {
...
+                val raw = result.credential as RawDcqlCredential
+                val stored = storedById[raw.id]
+                    ?: error("Credential '${raw.id}' disappeared while checking holder-key eligibility")
+                if (allowedCredentialIds != null && stored.id !in allowedCredentialIds) {
+                    false
+                } else if (stored.credential !is id.walt.credentials.formats.MdocsCredential) {
+                    true
+                } else {
+                    try {
+                        resolveHolderKey(stored)
```
JakeFernandes98: **Major:** Eligibility always authorizes `SIGN` here (`resolveHolderKey(stored)` uses the default usages). That is the other half of the MAC-import mismatch: `withRequiredImportedHolderKeyBinding` can persist a `KEY_AGREEMENT` binding, then this filter drops it (or fails the required set) with `KEY_USAGE_UNSUPPORTED`.

Same for `mdocHolderKeyResolver`, which also calls `resolveHolderKey` without an explicit usage set. Please keep import, eligibility, and presentation resolution on one rule.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/WalletPresentFunctionality2.kt
```diff
@@ -123,7 +125,10 @@ object WalletPresentFunctionality2 {
                             )
 
                         resolvedFormat == WalletPresentationFormatRegistry.SupportedFormat.MSO_MDOC -> {
-                            holderCrypto2Key?.let {
+                            val mdocHolderKey = mdocHolderKeyResolver
```
JakeFernandes98: **Major:** This still falls back to the wallet default when `mdocHolderKeyResolver` is omitted. Wallet2 / mobile now pass a resolver, and `MdocPresenter.requireHolderKeyMatchesMso` will reject a *wrong* default, but a single-key wallet can still sign without validating the persisted binding, provider slot, or thumbprint.

Architecture #60: no mdoc presentation path falls back to a wallet default key. Require the resolver (or fail) on the mdoc branch instead of coalescing.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/data/WalletKeyStore.kt
```diff
@@ -7,15 +7,30 @@ import kotlinx.coroutines.flow.Flow
...
+     * distinct exceptions.
+     */
     suspend fun getCrypto2Key(keyId: String, usages: Set<KeyUsage> = emptySet()): Crypto2Key? = null
 
+    /**
+     * Resolves public Crypto2 material for identity matching without authorizing a private-key
+     * operation. Providers whose operational lookup requires a non-empty usage set should override
+     * this method with a public-only lookup.
+     */
+    suspend fun getPublicCrypto2Key(keyId: String): Crypto2Key? = getCrypto2Key(keyId)
```
JakeFernandes98: **Major:** The default public lookup delegates to `getCrypto2Key(keyId)` with an empty usage set. `emptySet().all { … }` is vacuously true, so `SqlDelightKeyStore`, `ExposedKeyStore`, and `InMemoryKeyStore` restore the full operational key (including signer) during import discovery.

Enterprise #641 added a real verify-only `getPublicCrypto2Key` / `resolvePublic` so sidecar validation stays fail-closed and discovery does not authorize a private operation. Please override public-only lookup on the SQLDelight and Exposed stores (and in-memory if it remains a mobile stand-in) to match that contract.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonTest/kotlin/id/waltid/openid4vp/wallet/WalletCrypto2PresentationSigningTest.kt
```diff
@@ -205,70 +200,4 @@ class WalletCrypto2PresentationSigningTest {
         )
     }
```
JakeFernandes98, line 208: **Minor:** The old `mdoc device authentication signs through crypto2` test was deleted here (the MSO-less fixture would fail the new DeviceKey equality check). Holder-key signing coverage moved to `HolderKeyBindingTest`, but the `coseKeyToCrypto2Key` algorithm / empty `key_ops` guards were not replaced. Please restore those in an mdoc-crypto test if they are still product invariants.

waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-mobile/src/androidHostTest/kotlin/id/walt/wallet2/persistence/stores/SqlDelightCredentialStoreBindingTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.time.Instant
+
+class SqlDelightCredentialStoreBindingTest {
+
+    @Test
+    fun `fresh schema round trips holder key binding`() = runTest {
```
JakeFernandes98, line 23: **Minor:** This is Android-host only. The SQLDelight schema is shared, so a JVM driver round-trip is probably enough, but WAL-1343’s acceptance criteria name both Android and iOS persistence. Please add an iOS-driver binding round-trip or state why the shared schema makes that redundant.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
```diff
@@ -1451,27 +1462,41 @@ object WalletIssuanceHandler {
...
             label = label,
             addedAt = Clock.System.now(),
             metadata = metadata,
-        ).also { addCredential(it) }
+        )
+        val bound = if (stored.credential is MdocsCredential) {
+            withVerifiedIssuanceHolderKeyBinding(
+                credential = stored,
+                keyMaterial = requireNotNull(keyMaterial) {
+                    "Exact issuance holder-key material is required when storing an mdoc"
```
JakeFernandes98, line 1499: **Minor:** Isolated `FetchCredentialRequest.keyId` / `PollDeferredRequest.keyId` are optional on the type, but this fail-closed check makes them required in practice for any isolated path that stores an mdoc. The PR breaking-changes section covers the mobile schema and experimental model, not this caller contract. `Wallet2ExtendedIntegrationTest` isolated `storeInWallet = true` still omits `keyId` (safe only because that fixture is not an mdoc).

Please call the isolated-mdoc `keyId` requirement out in the PR body and the OpenAPI / KDoc for those request types so Wallet API2 and Enterprise clients do not discover it in production.

waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-mobile/src/commonMain/kotlin/id/walt/wallet2/persistence/stores/SqlDelightCredentialStore.kt
```diff
@@ -74,6 +78,9 @@ public class SqlDelightCredentialStore(
             label = label,
             addedAt = Instant.fromEpochMilliseconds(added_at),
             metadata = metadata?.let { json.decodeFromString(JsonObject.serializer(), it) },
+            holderKeyBinding = holder_key_binding?.let {
+                json.decodeFromString(HolderKeyBinding.serializer(), it)
```
JakeFernandes98: **Minor:** Corrupt persisted binding JSON throws a generic serialization exception here (same pattern in `ExposedCredentialStore`). Presentation then fails without a typed `HolderKeyBindingException` (`UNSUPPORTED_BINDING_VERSION` / `KEY_REFERENCE_INVALID`). Please map decode failures to a stable holder-key code before they escape the store.

JakeFernandes98 (changes requested):
Please check the inline comments regarding requested changes

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/data/HolderKeyBinding.kt
szijpeter: Thanks, this was a valid mismatch. I kept the binding operation-neutral rather than reverting import to signing-only: the ISO DeviceKey is used for device signature or MAC, and the downstream proximity flow already resolves the same binding with `SIGN` or `KEY_AGREEMENT`. `resolveHolderKey` no longer has a default; OpenID4VP and Annex C now request `SIGN` explicitly, and a MAC-only key is tested as correctly bound but intentionally ineligible for those signature-only routes. The architecture contract, local plan, Linear ticket, and PR body are aligned with that rule.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletPresentationHandler.kt
szijpeter: Addressed as the call-site half of the same invariant. Eligibility and presentation can no longer inherit `SIGN`: every resolver call supplies a non-empty usage set explicitly. Current remote and Annex C paths request `SIGN`; the later proximity processor selects `SIGN` or `KEY_AGREEMENT` from its device-authentication method.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonMain/kotlin/id/waltid/openid4vp/wallet/WalletPresentFunctionality2.kt
szijpeter: Agreed. The mdoc branch now requires `mdocHolderKeyResolver` and fails if it is absent; it no longer coalesces to either wallet-default key. This keeps provider slot, persisted thumbprint, MSO equality, usage, and authorization on the single credential-bound path.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/data/WalletKeyStore.kt
szijpeter: Agreed. Public discovery now returns a dedicated immutable `WalletPublicKeyMaterial` containing a public JWK and cannot return an operational key handle. SQLDelight and Exposed read public material from their persisted key descriptors, in-memory/test stores export public-only material, and unavailable public-only lookup fails closed. Operational resolution happens only after the binding and explicit usage are known.

waltid-libraries/protocols/waltid-openid4vp-wallet/src/commonTest/kotlin/id/waltid/openid4vp/wallet/WalletCrypto2PresentationSigningTest.kt
szijpeter: Agreed. I restored focused mdoc-crypto tests for the algorithm-mismatch and empty-`key_ops` guards without bringing back the obsolete MSO-less signing fixture.

waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-mobile/src/androidHostTest/kotlin/id/walt/wallet2/persistence/stores/SqlDelightCredentialStoreBindingTest.kt
szijpeter: Agreed. I added an encrypted iOS native-driver holder-binding round-trip alongside the Android-host coverage, so both named platform persistence paths are exercised through the shared SQLDelight schema.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
szijpeter: Agreed. `keyId` remains optional for requests that do not store an mdoc, but KDoc and OpenAPI now state that isolated fetch/poll callers storing an mdoc must supply the exact original proof/device key ID. The PR breaking-change section calls out the same conditional requirement.

waltid-libraries/protocols/waltid-openid4vc-wallet-persistence-mobile/src/commonMain/kotlin/id/walt/wallet2/persistence/stores/SqlDelightCredentialStore.kt
szijpeter: Agreed. Malformed persisted JSON now maps to stable `BINDING_INVALID` in both SQLDelight and Exposed stores, with cancellation preserved. Once JSON decodes, unsupported schema/extractor versions and invalid references still use their more specific existing codes. Both store paths have corruption coverage.

JakeFernandes98 (approved):

PR #2159, opened 2026-08-28
[WAL-1346] feat(wallet): expose proximity presentation through the Wallet SDK

szijpeter:
## Summary

This adds the public holder-side proximity-presentation capability to the mobile Wallet SDK for [WAL-1346](https://linear.app/walt-new/issue/WAL-1346). Android/Kotlin consumers get one sealed KMP session contract, while native iOS gets a Swift-idiomatic facade over the same source of truth.

This PR is stacked on [walt-id/waltid-identity#2157](https://github.com/walt-id/waltid-identity/pull/2157), which owns the mobile BLE transport. The governing architecture is documented in [walt-id/waltid-architecture#60](https://github.com/walt-id/waltid-architecture/pull/60).

### Public Wallet SDK contract

- Uses the `Proximity` prefix consistently across Kotlin and Swift public types. The README covers every session state and cancellation-safe session cleanup.
- Adds immutable configuration and side-effect-free capability results.
- Exposes a single session with sealed states, typed legal actions, stable errors, idempotent close, and explicit remediation results.
- Exposes `NoData` in Kotlin and `noData(exchange:)` in Swift when the final request sends no credential data; late observations cannot replace that terminal result.
- Computes the shared consent-time reader-authentication summary from document and whole-request coverage. Malformed, invalid, and revoked authentication is rejected before consent and rendered through the terminal outcome.
- Models multi-document review, credential alternatives, exact disclosure choices, reader-authentication scope and evidence, asserted purpose/retention data, application-profile authorization, and repeated requests.
- Requires a session-bound opaque review ID on approve and decline. Invalid submissions leave the review open; one accepted decision consumes it exactly once, and cancellation makes it stale.
- Distinguishes prerequisite retry in an eligible live session from starting a fresh session after terminal failure.

### Shared coordination

- Bridges wallet credential inventory, holder-key authorization, reader trust, application profiles, the holder protocol engine, and mobile transport without leaking lower-layer types into the public API. Existing mdoc validation uses the released models; proximity parses the original signed bytes into the strict edition-2 models after validation.
- Snapshots configuration, review choices, selected disclosure, application values and continuation before validation or suspension. Holder authorization reads the retained accepted choice by identity; fresh credential/key/trust/profile checks remain in force.
- Orders decisions and public state through one session owner with exchange/review guards, preventing delayed engine observations from regressing a newer review or terminal state.
- Matches issuer constraints against applicable AKIs throughout the supplied issuerAuth certificate chain while keeping certificate validation separate.
[...]

PR #2157, opened 2026-08-27
[WAL-1345] feat(mdoc): add mobile BLE proximity transport

szijpeter:
## Summary

This adds the Kotlin Multiplatform BLE transport for ISO mdoc proximity presentation defined by [WAL-1345](https://linear.app/walt-new/issue/WAL-1345/implement-ble-proximity-transport-for-android-and-ios).

This targets its direct parent, [walt-id/waltid-identity#2155](https://github.com/walt-id/waltid-identity/pull/2155), which owns the protocol and transport contracts.

Architecture contract: [walt-id/waltid-architecture#60](https://github.com/walt-id/waltid-architecture/pull/60).

### Shared BLE transport

- Documents provider setup, capability checks, BLE role/bearer selection, lifecycle ownership, and the wallet integration boundary in the module README.
- Adds validated transaction UUID, central-client/peripheral-server/dual-role, and GATT-only/L2CAP-preferred models that prevent illegal role and bearer combinations.
- Implements BLEIdent HKDF-SHA-256 verification from the complete tagged `EDeviceKeyBytes`, normalizes supported public-key export encodings before COSE conversion, and provides ISO GATT chunking/reassembly, L2CAP length framing, message limits, truncation detection, and dynamic LE PSM validation.
- Provides deterministic role and bearer races, serialized sends, one receive consumer, bounded preparation setup and post-connection inactivity, normalized errors, structured cancellation, and idempotent cleanup; a prepared listener remains available for the session-owned engagement lifetime, and the selected connection is completed and closed exactly once.
- Gives successful L2CAP responses a one-second drain interval before native socket/stream teardown on Android and iOS, so immediate cleanup does not discard queued bytes. Error and cancellation cleanup remain immediate; the interval is not a reader acknowledgement.
- Moves opaque session messages only; engagement construction, session cryptography, requests, consent, trust, and wallet policy remain in the shared proximity engine and wallet layers.

### Android and Apple adapters

- Binds Android GATT completion to the selected peer and matching attribute/operation; wrong or already-delivered callbacks cannot acknowledge a later send.
- Preserves CoreBluetooth packet order under backpressure and releases blocked sends on disconnect/cancellation. Native callback fault tests use test-only OS doubles and internal access, without adding published framework exports.

- Implements Android scanning/advertising, both GATT roles, negotiated MTU handling, notification backpressure, and insecure LE L2CAP client/listener channels from API 30.
[...]

Issue #61, opened 2024-01-04
Erroneous did document for did:web not accepted by MS Entra


severinstampler:
did:web documents as generated by DidService are erroneous, as the verification method #0 id does not correspond to the referenced ids of the other methods.
The document as created by the DidService is therefore not accepted by MS Entra, whereas if the method id is corrected it will be accepted:

**Faulty document:**

```
{
  "id": "did:web:entra.walt.id:holder",
  "verificationMethod": [
    {
      "id": "did:web:entra.walt.id:holder",
      "type": "JsonWebKey2020",
      "controller": "did:web:entra.walt.id:holder",
      "publicKeyJwk": {
        "kty": "EC",
        "use": "sig",
        "crv": "secp256k1",
        "kid": "48d8a34263cf492aa7ff61b6183e8bcf",
        "x": "TKaQ6sCocTDsmuj9tTR996tFXpEcS2EJN-1gOadaBvk",
        "y": "0TrIYHcfC93VpEuvj-HXTnyKt0snayOMwGSJA1XiDX8"
      }
    }
  ],
  "assertionMethod": [
    "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf"
  ],
  "authentication": [
    "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf"
  ],
  "capabilityInvocation": [
    "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf"
  ],
  "capabilityDelegation": [
    "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf"
  ],
  "keyAgreement": [
    "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf"
  ]
}
```

**Correct document:**
```
{
  "id": "did:web:entra.walt.id:holder",
  "verificationMethod": [
    {
      "id": "did:web:entra.walt.id:holder#48d8a34263cf492aa7ff61b6183e8bcf",
      "type": "JsonWebKey2020",
      "controller": "did:web:entra.walt.id:holder",
      "publicKeyJwk": {
        "kty": "EC",
        "use": "sig",
        "crv": "secp256k1",
        "kid": "48d8a34263cf492aa7ff61b6183e8bcf",
        "x": "TKaQ6sCocTDsmuj9tTR996tFXpEcS2EJN-1gOadaBvk",
[...]

PR #2160, opened 2026-08-28
[WAL-1346] feat(wallet): add proximity presentation to wallet demos

szijpeter:
## Summary

This integrates holder-side proximity presentation into all three mobile wallet demos for [WAL-1346](https://linear.app/walt-new/issue/WAL-1346): Compose Android, Compose iOS, and native SwiftUI. Each app gets a dedicated in-person entry below the existing online presentation action and keeps the complete journey inside the familiar Present-tab wallet shell.

Shared reader trust also gains optional, explicit CRL verification for reader certificates and their issuing authorities. Applications supply transport; the SDK validates the signed status before disclosure.

This PR is stacked on the Wallet SDK contract in [walt-id/waltid-identity#2159](https://github.com/walt-id/waltid-identity/pull/2159). The lower mobile BLE transport is provided by [walt-id/waltid-identity#2157](https://github.com/walt-id/waltid-identity/pull/2157), the governing architecture is documented in [walt-id/waltid-architecture#60](https://github.com/walt-id/waltid-architecture/pull/60), and exact-build physical evidence and the public ISO holder-test plan are tracked in [WAL-1349 architecture PR #61](https://github.com/walt-id/waltid-architecture/pull/61).

### Shared Compose journey

- Adds one shared Android/iOS controller and screen set for prerequisites, Device Engagement QR, connection progress, request review, holder choices, repeated requests, terminal states, cancellation, and recovery.
- Preflights the exact selected SDK configuration before session creation, executes runtime-permission remediation, rechecks after every host result, and starts only when the selected route is viable. On Android, permissions that the OS will no longer prompt for become an explicit Open app settings action; returning from Settings triggers a fresh capability check before session creation.
- Embeds QR, connection, review, and result states under the normal wallet toolbar and status message instead of replacing the host screen.
- Reuses the existing credential art, human-readable credential and claim labels, metadata sections, and review action layout so online and in-person presentation remain visually aligned.
- Renders the exact Wallet SDK review, including reader-authentication scope, signature validity, certificate path, revocation, optional RICAL evidence, product trust, asserted purpose, and retention intent.
- Renders reader information as an honest collapsed summary with expandable authentication/trust evidence when the request provides it; unsigned requests do not invent a reader identity or logo.
- Uses the SDK coverage summary in Compose and SwiftUI, so valid whole-request authentication does not appear partially missing merely because per-document statements are absent.
- Shows “No data shared” for the no-data outcome, with wording scoped to the final request in repeated sessions.
- Supports per-document credential alternatives and exact element disclosure choices. Every review captures its opaque identity for approve/decline. The first decision reserves that review, blocks duplicate/conflicting actions and selection changes, and keeps cancellation available; stale results cannot affect a later review.

### Reader-trust configuration ownership

- Owns the cross-platform Settings surface for Reader CA anchors and optional RICAL trust sources, with add, inspect, remove, reset, and explicit session-snapshot behavior.
- Imports only public DER/PEM certificate material and a versioned walt.id trust-bundle format after validation and fingerprint preview; PKCS#12/PFX and private reader keys are rejected.
- Keeps trust material app-private and avoids logs, bundled production roots, or implicit network trust. Decode validates stored structure; import validates material, and session evaluation establishes current trust. Applying settings and constructing a native session configuration fail recoverably.
- Validates raw reader-certificate structure before platform projections can hide signature-algorithm mismatches or duplicate extensions, keeping Android and iOS rejection behavior consistent.
- Adds shared verification of complete direct CRLs, including reader-authority revocation, with issuer/key linkage, signing authority, signature and freshness checks. Verified revocation prevents disclosure before consent; unsupported, stale or unavailable status remains indeterminate. Good status is rechecked after all required fetches finish.
- Exposes reader-only, validated-path (excluding the selected anchor), and explicit terminal-authority CRL scopes with an application-owned fetcher. Direct and RICAL trust evaluation supply the actual validated path. The SDK performs no implicit lookup or persistent caching, and demo trust imports do not configure a CRL client.
- Selects RICAL constraints from the first matching authority along the validated path, independent of array order when both intermediate and root are anchors; incompatible maximal paths fail closed.
- Supports an explicit required IACA direct issuer in Kotlin and Swift, matching its exact certificate and enforcing the conditional non-critical issuerAlternativeName email/URI rule. Generic CA imports do not establish IACA role.
- Runs bounded file reads, import parsing and persistence off the UI thread; canceled imports cannot publish late previews. Applying stored settings replaces the evaluator, so applications install custom CRL/IACA policy afterward.
- Exposes the same evaluator through the native Swift facade with Foundation certificate/CRL bytes, an async application fetcher, recoverable input errors and Swift task cancellation.

### Platform integration

- Aligns Compose, native SwiftUI, reader-trust adapters, and their fixtures with the public `Proximity` API names.
- Keeps runtime permission requests, Bluetooth settings actions, lifecycle interruption, keep-awake, temporary QR brightness, and platform accessibility behind Android and iOS host boundaries.
- Keeps the complete journey inside each host's Present-tab safe area and makes the native SwiftUI parent observe the child session directly, with one start owner and deterministic dismissal/restart behavior.
- Reuses the shared native iOS review components while keeping the SwiftUI journey a thin renderer over the Swift Wallet SDK session rather than reconstructing protocol state.
- Ends the journey on in-app Settings navigation or leaving Present. Wallet reset awaits startup/session cleanup before deleting material, and the SDK closes admission to new sessions. Screen policy is restored on exit/background and reapplied on iOS activation; transient native inactive phases do not cancel the journey.

### QR, accessibility, and app wiring

- Pins the minimum proximity-only renderer dependencies directly in this PR: ZXing Java 3.5.4 on Android and ZXing-C++ 3.1.1 on Compose iOS and native SwiftUI.
- Accepts only bounded ASCII `mdoc:` Device Engagement text without whitespace or control bytes, uses explicit low error correction with no ECI and non-interpolated output, fails closed instead of truncating oversized input, pins the Compose iOS/native SwiftUI module fingerprint, and renders an exact four-module quiet zone using whole physical-pixel modules.
- Uses a package-private Objective-C++ adapter in the native SwiftUI demo with a narrow proximity-only entry point because the pinned Swift wrapper does not expose no-ECI ASCII encoding; the adapter adds no second QR implementation or public module.
- Adds aligned cross-platform accessibility identifiers, privacy-safe QR labels, resource-backed Compose settings and shared Swift online/proximity action titles, and focused controller/UI/view-model fixtures, including native presentation lifecycle and cross-tab layout regressions.
- Preflights capabilities in both hosts at this PR boundary. Compose retains automatic permission prompting and native retains explicit prerequisite actions; both iOS hosts wait for return from Settings before rechecking.

## Architecture Notes

- Protocol state, trust, request meaning, legal actions, disclosure validation, and capability/remediation reporting remain owned by the Wallet SDK. Demo code owns holder UI choices and execution of platform actions such as permission prompts.
- Compose Android and Compose iOS share the journey semantics and rendering. Native SwiftUI shares the SDK contract, review primitives, and test identifiers while remaining platform-idiomatic.
- The normal UI does not expose BLE roles, UUIDs, transport internals, or protocol-version controls.
- QR encoding remains a demo-host concern outside the public Wallet SDK; the shared behavior is explicit without introducing a new published KMP module.
- CRL DER and signature verification live in the shared X.509 layer; the wallet owns reader-path selection and trust policy. Android certificate/CSR verification uses an explicit provider and does not depend on earlier global registration.

## Caveats and Follow-Ups

- The [WAL-1349 qualification index](https://github.com/walt-id/waltid-architecture/pull/61) records current case outcomes, exact tested builds, repaired regressions and remaining gaps, with pinned reader forks and step-by-step trust setup. Physical results stay bound to their recorded artifacts.
- This owns app integration and public reader-trust configuration; it does not introduce a production payment profile or universal trust list.
- CRL support is limited to direct complete v2 lists with AKI, CRL number and nextUpdate, using ECDSA or RSA PKCS#1 with SHA-256/384/512. Delta, indirect and partitioned CRLs and OCSP remain separate work. The shared verifier overlaps existing [WAL-1148](https://linear.app/walt-new/issue/WAL-1148/certificate-revocation-list-validation); independent physical status-server qualification remains in WAL-1349.

## Breaking

- No released API or persisted wallet data is migrated. Native consumers handle throwing proximity-model construction and `ProximityReaderTrustSettings.applying(to:)` directly in the unreleased SDK contract.


JakeFernandes98 (commented):
## Review — Major findings (WAL-1346 demos + reader-trust + QR)

Focused review of this PR's own delta (base `feature/wal-1346-proximity-wallet-sdk`, head `6146ec9`). **0 Blocking, 2 Major.** The security-critical surfaces verified sound — trust import validates fully before atomic persist (malformed/partial input never replaces prior config), PKCS#12/PFX + private keys rejected, preview fingerprint and persisted DER come from the same parsed cert, app-private storage, QR is ASCII-only + fail-closed size + EC-level L + no ECI + exact 4-module quiet zone on all three platforms, the ObjC++ CoreGraphics ownership is correct, and session trust config is snapshotted once and frozen. The items below are cross-platform parity concerns.

### Major
1. **Prerequisite/permission decision semantics diverge across platforms.** Compose preflights capabilities and auto-requests BLE permission before creating the session; native SwiftUI creates the session immediately with manual-only remediation. The README documents only the Compose behavior. Align the semantics (and docs) across the three surfaces.
   - `WalletDemoProximityController.kt:1310-1333` + `WalletDemoProximityScreen.kt:3634-3638` vs `ProximityPresentationViewModel.swift:5662-5696` + `ProximityPresentationView.swift:6438-6482`; `waltid-wallet-demo-compose/README.md:44-45`

2. **Native SwiftUI `preconditionFailure` (fatal) on malformed persisted trust bytes at session start**, whereas the KMP path throws recoverably. Gated by decode validation today, but a fatalError-on-app-group-data divergence from the KMP contract. Prefer a recoverable/throwing path.
   - `ProximityReaderTrustSettings.swift:15216-15221` vs `MobileWalletProximityReaderTrustSettings.kt:12026-12061`

---
_AI-assisted review; Major severity (non-blocking). Verified against code at head `6146ec9`._

szijpeter:
Thanks for the review. Both points are addressed with the stack boundary made explicit:

- **Permissions:** this PR's Compose/native difference is now documented. The existing alignment is in [#2166](https://github.com/walt-id/waltid-identity/pull/2166): both demos use explicit permission actions, and native preflight runs before session creation. The full-stack docs now describe that behavior; we have not duplicated the later implementation here.
- **Stored trust:** malformed persisted data was already rejected by the throwing decoder. As defensive hardening, applying settings now throws instead of using the fatal Base64URL fallback, and native consumers surface configuration errors without starting a session. Applying settings also cannot weaken a profile that requires trusted readers.

The PR description and integration docs now describe the recoverable errors and permission ownership.

waltkb (commented):
# CRL stack hidden ins this wallet demos PR, and appears to be dead code

ProximityCrlRevocationEvaluator (MobileWalletProximityCrlRevocation.kt), CertificateRevocationListVerifier + CrlDer.kt and ProximityReaderTrustSettings.swift add ~1300 lines, yet:
- ProximityReaderTrustSettings.applyTo (MobileWalletProximityReaderTrustSettings.kt) never sets revocationPolicy, so it defaults to NotChecked. There is no Kotlin-side path to enable CRL checking at all, and no demo configures it (the PR admits "demo trust imports do not configure a CRL client").
- The only consumer is KMPWalletCoreBridge.swift, i.e. iOS-native SDK users only. Compose Android and Compose iOS cannot use the feature
- overlaps existing WAL-1148

# RicalReaderPathResult.Revoked is unreachable, RICAL revocation does not exist
Rical.kt gained a Revoked variant, MobileWalletProximityReaderTrust.kt maps it to ProximityReaderTrustState.Revoked, and the review UI renders "revocation". But the only implementation, X509RicalReaderPathValidator.validate (RicalX509Validation.kt), returns only Invalid, NoMatch or Valid, and RicalCertificateInfo (Rical.kt:) carries no revocation or status field to derive it from. This is a dead branch

# ISO mdoc Reader CA
ProximityReaderTrustAnchorPreview.profile (MobileWalletProximityReaderTrustSettings.kt) is a hardcoded default shown to the holder during import review. The actual check is validateReaderCa -> validateCertificateAuthorityUsage (OrderedCertificateChain.kt), which only verifies validity dates, cA=true and keyCertSign. Any CA certificate at all is labelled "ISO mdoc Reader CA".

ISO 18013-5 Table B.6 NOTE 1 explicitly says the reader CA certificate "and the manner in which it is trusted by an mDL, is outside the scope of this document", there is no ISO reader CA profile to conform to. So this appears to be an invented standard claim?

# Serial-number length is checked on DER encoding instead of value
validateUnsignedInteger (CrlDer.kt) requires bytes.size <= 20 on the raw INTEGER content, before the leading-zero strip. A conforming 20-octet positive serial with the high bit set encodes as 21 content octets and is rejected. CertificateRevocationListVerifierTest.kt asserts exactly this (byteArrayOf(0) + ByteArray(20) { 0x80 } -> INVALID_DER).
RFC 5280 §4.1.2.2 bounds the value, not the encoding. Roughly half of all 20-octet random serials hit this. Same defect in validateIsoCertificateSerialNumber (IsoMdocReaderAuthenticationCertificate.kt, serialNumber.size in 8..20), which rejects reader certificates outright. Fix by normalise first, then bound the normalised length to 20.

# In-flight startProximityPresentation is cancelled without joining
WalletDemoProximityController.kt does sessionJob?.cancel() with no join, and retryPrerequisites() re-enters checkPrerequisitesAndStart(configuration, generation) without bumping the generation. Meanwhile MobileWalletProximity.kt does active = session; session.start(). If the cancelled coroutine dies at a suspension point inside start(), active stays set and the handle is unreachable, so every subsequent start fails permanently with "A proximity presentation session is already active" and the BLE advertiser leaks for the process lifetime.

The active-leak is PR #2159's bug, but this PR makes it reachable. Minimum fix here would be to bump generation in retryPrerequisites, and cancelAndJoin before relaunching.

# Compose keeps the session live outside the QR screen
WalletDemoProximityScreen.kt derives qrVisible purely from selectedTab == Present, but WalletScreen.kt early-returns to SettingsScreen/credential details while selectedTab is still Present. Opening Settings mid-engagement keeps FLAG_KEEP_SCREEN_ON, full brightness and BLE advertising with no visible cancel affordance.
Related: LaunchedEffect(walletState.auth, ...) is the only guard on wallet reset, but WalletDemoController.resetWallet() awaits deleteWallet() before clearing auth, so the BLE session runs across wallet deletion. The native app orders this correctly (WalletViewModel.swift).

# Reader-trust import runs certificate/COSE validation and file I/O on the main thread
DemoReaderTrustSettings.kt launches prepareImport on Dispatchers.Main with no withContext(Dispatchers.Default); persist() does SharedPreferences.commit() on main (AndroidDemoReaderTrustSettingsStore.kt); ReaderTrustImportPicker.android.kt reads up to 1 MiB in the activity-result callback. Swift is the same (DemoReaderTrustSettings.swift is @MainActor). The importInProgress spinner cannot render while this happens.

# Checking the self-signed trust anchor against its own CRL
resolvePath (MobileWalletProximityCrlRevocation.kt) walks up until certificate == issuer, appending (root, root), and then requires every entry - including the root - to have a fresh Good CRL. A trust anchor cannot meaningfully revoke itself, RFC 5280 §6 does not check anchors for revocation, and most roots publish no CRL. ReaderCertificateAndIssuingAuthorities will therefore almost always return Indeterminate.

# Duplication
Five copies of the same Base64URL codec, two serial validators, two identical bundle DTOs
This PR adds:
- decodeTrustBase64Url/encodeTrustBase64Url in MobileWalletProximityReaderTrust.kt, an identical pair in MobileWalletProximityReaderTrustSettings.kt, crlBase64 in MobileWalletProximityCrlRevocation.kt, plus pre-existing copies in MobileWalletProximityRequestProcessor.kt and MobileWalletProximityModels.kt.
- validateIsoCertificateSerialNumber (IsoMdocReaderAuthenticationCertificate.kt) is a third implementation of the ISO serial rule alongside ValidationCommons.kt validateSerialNo and IsoProfileX509CertificateValidationUtil.validateSerialNumber. The new one is the best of the three
- PersistedReaderAuthority and BundleReaderAuthority are field-identical, while PersistedRicalProvider is shared between both formats.
- The two demo journey controllers (WalletDemoProximityController.kt 392 lines / ProximityPresentationViewModel.swift 448 lines) duplicate generation counters, default selections, canApprove, engagement extraction and failure text - and have already diverged (Compose preflights capabilities and auto-requests permission; Swift does not). This belongs in waltid-openid4vc-wallet-mobile?
- Terminal-state classification is re-enumerated a third and fourth time in WalletDemoProximityController.kt and ProximityPresentationViewModel.swift because the SDK keeps it private. Adding a terminal state will silently hang both demos.

# RicalReaderTrustEvaluator changed authority-selection semantics
Old code iterated all matching certificateInfos and accepted the first whose constraints were satisfied. New code picks the single bottom-most authority (RicalX509Validation.kt, maxByOrNull, non-deterministic on ties) and returns noMatchingAuthority() if that authority's constraints fail. Also rical.certificateInfos.singleOrNull { it == path.authority } and .single { it.subjectKeyIdentifier == ... } mean a RICAL containing duplicate infos, or two infos sharing an SKI, degrades to Invalid/throws. And the candidate loop runs a full chain validation (with signature verification) per certificateInfo, inside a live BLE exchange with timing constraints.

# Kotlin idiom / style
- CertificateRevocationListVerifier.kt: requireCrl(verifier != null, ...) then verifier!!. Should be ?: throw CrlValidationException(CrlFailure.UNAVAILABLE).
- CertificateRevocationListVerifier.kt and MobileWalletProximityCrlRevocation.kt: catch (_: Throwable) swallows Errors and turns any implementation bug into a silent Indeterminate. catch (_: Exception) at minimum.
- CrlCertificateStatus.Revoked(revocationTime, reasonCode: Int?) exposes a raw Int in a public API where an enum belongs.
- CertificateRevocationListVerifier.kt: a bad certificate signature is reported as CrlFailure.INVALID_ISSUER. Misleading category.
- MobileWalletProximityReaderTrustSettings.kt: require(existingFingerprints.add(f) && importedFingerprints.add(f)) relies on short-circuited side effects; importedFingerprints is redundant because existingFingerprints is already mutated in the loop.
- MobileWalletProximityReaderTrustSettings.kt: every bundle certificate is parsed twice and the two lists are correlated positionally by index.
- MdocReaderAuthentication23220EkuOid (IsoMdocReaderAuthenticationCertificate.kt) is a public const with a KDoc calling it "recommended" that is never used anywhere.
- DemoReaderTrustSettings.kt appears to use en dashes

# Import safety checks guarantee nothing
MobileWalletProximityReaderTrustSettings.kt rejects .p12/.pfx by filename and looks for the literal text "PRIVATE KEY" in a decodeToString() of up to 1 MiB of binary. Renaming a file defeats the first; a DER PKCS#8 key defeats the second. The real protection is that non-certificates fail to parse. The 1 MiB string allocation per DER import is a cost. Similarly, decode()'s validateStoredShape only re-parses structure - it does not re-check CA usage or the RICAL signature, while the description states "invalid stored settings are rejected during decode".

# Misc
- WalletDemoQRCodeCore.mm: CGImageCreate(..., /*shouldInterpolate*/ true, ...) on a QR bitmap. Masked by .interpolation(.none) in SwiftUI, but wrong at the image level and contradicts whole physical-pixel modules claim, should be false.
- WalletDemoQRCodeCore.mm: std::string ascii(utf8) truncates at an embedded NUL, which validateProximityQrCodePayload (QrCodeImage.kt:21, 0..0x7F) and WalletQRCodeRenderer.swift both permit. Unreachable for base64url in practice, but it's not "fails closed instead of truncating".
- WalletDemoQRCodeCore.mm uses ZXing::MultiFormatWriter, which is [[deprecated("use CreateBarcode()")]] in the pinned v3.1.1 (core/src/MultiFormatWriter.h). The non-deprecated CreateBarcodeFromText + WriteBarcodeToImage would work, and is actually what the Kotlin/Native side uses.
- QrCodeImage.ios.kt: throw OutOfMemoryError() for a null data pointer - wrong exception type. Also readBytes(width * height) assumes rowStride == width with no check.
- CreatorOptions.options = "ecLevel=L" (QrCodeImage.ios.kt) is a no-op in the non-Zint build: CreateBarcode.cpp only forwards single-digit "0"-"8". The outcome is still L because that is the writer default, and Zint builds do parse "L" - but the option is not doing what it looks like it does.
- ColdCertificateSignatureValidationTest.kt mutates process-global java.security.Security state; it will flake other tests running concurrently in the same JVM. The fix it covers (BouncyX509CertificateSigner.kt, instance provider instead of setProvider("BC")) is good.
- ISO 18013-5 Table B.6 requires issuerAlternativeName (rfc822Name or URI) when the reader certificate is issued by an IACA. Not validated, although IssuerAlternativeName.kt already exists in the module.
- Cross-platform accessibility identifiers diverge for the new settings surface: settingsReaderPolicyAllowUntrusted/settingsReaderTrustImport/... (WalletUiTestTags.kt) vs readerTrustAllowUntrusted/readerTrustImport/... (WalletAccessibilityID.swift). The proximity* ones do match.
- SharingReviewView.swift now passes String to Button(_:action:), selecting the StringProtocol overload and de-localising Share/Cancel/Reject in the existing online flow too. submitTitle returns the same value on both branches.
- Hardcoded unlocalised strings in Compose despite the new 95-entry strings.xml: "Verifier" (WalletDemoProximityScreen.kt), "Device engagement QR code"- note proximity_qr_accessibility is the only unused entry in strings.xml - and the whole reader-trust settings screen (DemoReaderTrustSettings.kt).
- ProximityPresentationView.swift uses MainActor.assumeIsolated in deinit; it traps rather than restoring brightness if the last reference is released off-main.
- PresentView.swift cancels the session for any scenePhase != .active (Control Center, app switcher, system alerts), whereas Compose iOS correctly observes UIApplicationDidEnterBackgroundNotification (ProximityPlatformHooks.ios.kt:61). Compose iOS in turn never re-applies keep-awake/brightness on foreground (no WillEnterForeground observer), unlike Android's ON_START.
- ProximityPlatformHooks.ios.kt and ProximityPresentationViewModel.swift report remediation .completed as soon as Settings is opened, so prerequisites are re-checked while the user is still in Settings. Android correctly awaits the activity result.
- WalletDemoProximityScreen.kt hardcodes enabled = true on the approve button, so a double tap dispatches Approve twice and shows "Action could not be completed" after a successful disclosure.
- ProximityReaderTrustConfiguration and ProximityRicalConfiguration are data classes holding fun interface instances, so equals/hashCode are identity-based and meaningless; snapshot() is a shallow copy and the name oversells what it detaches.

szijpeter:
Following up on [this review](https://github.com/walt-id/waltid-identity/pull/2160#pullrequestreview-5236343448). The fixes are in the existing stack, rebased onto `main`; general host repairs are included in this PR and the NFC/prepared layer retains its extensions. Each concern is quoted verbatim below.

**1.**

> # CRL stack hidden ins this wallet demos PR, and appears to be dead code
> 
> ProximityCrlRevocationEvaluator (MobileWalletProximityCrlRevocation.kt), CertificateRevocationListVerifier + CrlDer.kt and ProximityReaderTrustSettings.swift add ~1300 lines, yet:
> - ProximityReaderTrustSettings.applyTo (MobileWalletProximityReaderTrustSettings.kt) never sets revocationPolicy, so it defaults to NotChecked. There is no Kotlin-side path to enable CRL checking at all, and no demo configures it (the PR admits "demo trust imports do not configure a CRL client").
> - The only consumer is KMPWalletCoreBridge.swift, i.e. iOS-native SDK users only. Compose Android and Compose iOS cannot use the feature
> - overlaps existing WAL-1148

CRL checking already has a common Kotlin entry point: `ProximityReaderTrustConfiguration(revocationPolicy = Check(...))`, used by the shared request processor. Android and Compose iOS can use it too. I updated the Kotlin/Swift examples and documented that `applyTo` replaces the evaluator; install runtime CRL/IACA policy afterward. Demo imports still do not start CRL networking: fetching, timeouts and caching belong to the application. The generic verifier remains in X.509 and the wallet adapter in the SDK. WAL-1148 overlaps this domain, but this bounded implementation does not complete that broader ticket; the architecture records that distinction. The existing stack is retained, with the shared scope made explicit in this PR. [README.md](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/README.md#L376).

**2.**

> # RicalReaderPathResult.Revoked is unreachable, RICAL revocation does not exist
> Rical.kt gained a Revoked variant, MobileWalletProximityReaderTrust.kt maps it to ProximityReaderTrustState.Revoked, and the review UI renders "revocation". But the only implementation, X509RicalReaderPathValidator.validate (RicalX509Validation.kt), returns only Invalid, NoMatch or Valid, and RicalCertificateInfo (Rical.kt:) carries no revocation or status field to derive it from. This is a dead branch

The built-in X.509 path validator does not emit that variant. It remains a supported result for an injected path validator; it is not evidence of built-in RICAL status checking. Separately, the configured wallet evaluator applies reader revocation after a RICAL match and has an explicit signer-revocation policy. No status field was added to the RICAL format, and the UI still reports only the resulting evidence. [Rical.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/credentials/waltid-mdoc-proximity/src/commonMain/kotlin/id/walt/mdoc/proximity/Rical.kt#L200), [MobileWalletProximityReaderTrust.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrust.kt#L88).

**3.**

> # ISO mdoc Reader CA
> ProximityReaderTrustAnchorPreview.profile (MobileWalletProximityReaderTrustSettings.kt) is a hardcoded default shown to the holder during import review. The actual check is validateReaderCa -> validateCertificateAuthorityUsage (OrderedCertificateChain.kt), which only verifies validity dates, cA=true and keyCertSign. Any CA certificate at all is labelled "ISO mdoc Reader CA".
> 
> ISO 18013-5 Table B.6 NOTE 1 explicitly says the reader CA certificate "and the manner in which it is trusted by an mDL, is outside the scope of this document", there is no ISO reader CA profile to conform to. So this appears to be an invented standard claim?

Changed the import label to “X.509 CA certificate.” It describes the validity/CA/key-usage checks actually performed and no longer claims an ISO Reader CA profile. Reader-certificate profile validation remains a separate use-time operation. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L148).

**4.**

> # Serial-number length is checked on DER encoding instead of value
> validateUnsignedInteger (CrlDer.kt) requires bytes.size <= 20 on the raw INTEGER content, before the leading-zero strip. A conforming 20-octet positive serial with the high bit set encodes as 21 content octets and is rejected. CertificateRevocationListVerifierTest.kt asserts exactly this (byteArrayOf(0) + ByteArray(20) { 0x80 } -> INVALID_DER).
> RFC 5280 §4.1.2.2 bounds the value, not the encoding. Roughly half of all 20-octet random serials hit this. Same defect in validateIsoCertificateSerialNumber (IsoMdocReaderAuthenticationCertificate.kt, serialNumber.size in 8..20), which rejects reader certificates outright. Fix by normalise first, then bound the normalised length to 20.

Kept the strict canonical DER limit. The sign-padding octet counts toward the 20 content-octet limit; stripping it first would admit the 21-octet case explicitly rejected by [ZLint’s RFC 5280 serial lint](https://github.com/zmap/zlint/blob/master/v3/lints/rfc/lint_serial_number_longer_than_20_octets.go). Normalization is still used for serial comparison. The reader profile also retains its lower bound; certificate parsing cannot prove how randomly a serial was generated. [CrlDer.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/revocation/CrlDer.kt#L88).

**5.**

> # In-flight startProximityPresentation is cancelled without joining
> WalletDemoProximityController.kt does sessionJob?.cancel() with no join, and retryPrerequisites() re-enters checkPrerequisitesAndStart(configuration, generation) without bumping the generation. Meanwhile MobileWalletProximity.kt does active = session; session.start(). If the cancelled coroutine dies at a suspension point inside start(), active stays set and the handle is unreachable, so every subsequent start fails permanently with "A proximity presentation session is already active" and the BLE advertiser leaks for the process lifetime.
> 
> The active-leak is PR #2159's bug, but this PR makes it reachable. Minimum fix here would be to bump generation in retryPrerequisites, and cancelAndJoin before relaunching.

The claimed permanent SDK admission leak was not reproduced: admission is followed by a non-suspending atomic worker launch with owned cleanup. I did not add indiscriminate joins inside that worker. The real host cleanup race is fixed: dismissal/reset now retain and await cancelled startup work, close late-returned handles, and drain previous cleanup before deletion or replacement. SDK deletion also closes admission to new sessions. [MobileWalletProximity.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximity.kt#L83), [WalletDemoProximityController.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/mobileMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoProximityController.kt#L401).

**6.**

> # Compose keeps the session live outside the QR screen
> WalletDemoProximityScreen.kt derives qrVisible purely from selectedTab == Present, but WalletScreen.kt early-returns to SettingsScreen/credential details while selectedTab is still Present. Opening Settings mid-engagement keeps FLAG_KEEP_SCREEN_ON, full brightness and BLE advertising with no visible cancel affordance.
> Related: LaunchedEffect(walletState.auth, ...) is the only guard on wallet reset, but WalletDemoController.resetWallet() awaits deleteWallet() before clearing auth, so the BLE session runs across wallet deletion. The native app orders this correctly (WalletViewModel.swift).

Opening in-app Settings now dismisses the Compose proximity journey through the actual route callback. Reset awaits startup/session cleanup before deleting wallet material in both hosts; native previously scheduled close without waiting. The upper stack also drains prepared-sharing revocation. Tests cover a cancelled start that returns a handle late and a close that finishes later still, plus the wallet Settings route. [WalletDemoProximityControllerTest.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/androidHostTest/kotlin/id/walt/walletdemo/compose/logic/WalletDemoProximityControllerTest.kt#L65), [WalletDemoProximityTestScenarios.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/mobileUiTest/kotlin/id/walt/walletdemo/compose/ui/WalletDemoProximityTestScenarios.kt#L67).

**7.**

> # Reader-trust import runs certificate/COSE validation and file I/O on the main thread
> DemoReaderTrustSettings.kt launches prepareImport on Dispatchers.Main with no withContext(Dispatchers.Default); persist() does SharedPreferences.commit() on main (AndroidDemoReaderTrustSettingsStore.kt); ReaderTrustImportPicker.android.kt reads up to 1 MiB in the activity-result callback. Swift is the same (DemoReaderTrustSettings.swift is @MainActor). The importInProgress spinner cannot render while this happens.

Moved Compose certificate/bundle validation, settings load/save, and Android file reads off Main. Both iOS file loaders use bounded reads off the UI thread and retain security-scoped access until reading finishes. Native settings load/encode/save are off Main too; the existing asynchronous SDK validation bridge is retained. Pending loads block new session configuration, cancelled imports cannot publish a late preview, and saves are serialized and published only after persistence succeeds. [DemoReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/mobileMain/kotlin/id/walt/walletdemo/compose/logic/DemoReaderTrustSettings.kt#L19), [DemoReaderTrustSettings.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/DemoReaderTrustSettings.swift#L63).

**8.**

> # Checking the self-signed trust anchor against its own CRL
> resolvePath (MobileWalletProximityCrlRevocation.kt) walks up until certificate == issuer, appending (root, root), and then requires every entry - including the root - to have a fresh Good CRL. A trust anchor cannot meaningfully revoke itself, RFC 5280 §6 does not check anchors for revocation, and most roots publish no CRL. ReaderCertificateAndIssuingAuthorities will therefore almost always return Indeterminate.

Added `ValidatedPath` / `.validatedPath`: it checks the reader and non-anchor intermediates on the actual authenticated path, stopping at the selected configured anchor. That path is preserved through the real Swift bridge. Leaf-only and the existing explicit terminal-authority policy remain separate. Raw evidence cannot use the new scope without a validated path. Tests cover a root without CDP, an intermediate anchor, revoked intermediate status, and the stricter terminal policy. This matches the anchor boundary in [RFC 5280 §6.1](https://www.rfc-editor.org/rfc/rfc5280.html#section-6.1) without silently weakening an existing caller’s stricter choice. [MobileWalletProximityCrlRevocation.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityCrlRevocation.kt#L18), [ProximityValidatedTrustPathTest.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonTest/kotlin/id/walt/wallet2/mobile/ProximityValidatedTrustPathTest.kt#L11).

**9.**

> decodeTrustBase64Url/encodeTrustBase64Url in MobileWalletProximityReaderTrust.kt, an identical pair in MobileWalletProximityReaderTrustSettings.kt, crlBase64 in MobileWalletProximityCrlRevocation.kt, plus pre-existing copies in MobileWalletProximityRequestProcessor.kt and MobileWalletProximityModels.kt.

Kept the small domain helpers for now: they do not all have the same acceptance contract (padding, empty input, field size and fingerprint validation differ). Consolidating them is optional cleanup, not a trust fix; a mechanical replacement would risk changing those boundaries. No new utility API/module was introduced. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L269).

**10.**

> validateIsoCertificateSerialNumber (IsoMdocReaderAuthenticationCertificate.kt) is a third implementation of the ISO serial rule alongside ValidationCommons.kt validateSerialNo and IsoProfileX509CertificateValidationUtil.validateSerialNumber. The new one is the best of the three

Kept the strict reader serial validator and the CRL comparison normalization separate from the existing issuance/profile helpers. They operate at different boundaries and have different failure contracts. This change does not broaden into a migration of other certificate profiles, and it does not relax canonical serial acceptance. [IsoMdocReaderAuthenticationCertificate.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/x509/IsoMdocReaderAuthenticationCertificate.kt#L48).

**11.**

> PersistedReaderAuthority and BundleReaderAuthority are field-identical, while PersistedRicalProvider is shared between both formats.

Kept the two private DTOs. Stored settings and import bundles are separate versioned input formats; their currently identical authority fields do not require coupling their future migrations. The actual parse-twice/index-correlation duplication is removed below. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L472).

**12.**

> The two demo journey controllers (WalletDemoProximityController.kt 392 lines / ProximityPresentationViewModel.swift 448 lines) duplicate generation counters, default selections, canApprove, engagement extraction and failure text - and have already diverged (Compose preflights capabilities and auto-requests permission; Swift does not). This belongs in waltid-openid4vc-wallet-mobile?

Kept host lifecycle/UI orchestration in the two demos; protocol state, trust evaluation and disclosure remain shared in the SDK. Reused the existing native capability-preflight repair and placed its general behavior in #2160. The pending-review and awaited-cleanup behavior now match across hosts. Native retains its explicit permission affordance; matching behavior does not require moving both controllers wholesale into the SDK. [ProximityPresentationViewModel.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/ProximityPresentationViewModel.swift#L198).

**13.**

> Terminal-state classification is re-enumerated a third and fourth time in WalletDemoProximityController.kt and ProximityPresentationViewModel.swift because the SDK keeps it private. Adding a terminal state will silently hang both demos.

Kept the small terminal projections in the hosts for this change. They currently cover every terminal SDK state, and the lifecycle tests exercise those exits. A shared public terminal-state API would be a separate API change; no missing current state or hanging terminal transition was established. [WalletDemoProximityController.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/mobileMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoProximityController.kt#L124).

**14.**

> # RicalReaderTrustEvaluator changed authority-selection semantics
> Old code iterated all matching certificateInfos and accepted the first whose constraints were satisfied. New code picks the single bottom-most authority (RicalX509Validation.kt, maxByOrNull, non-deterministic on ties) and returns noMatchingAuthority() if that authority's constraints fail. Also rical.certificateInfos.singleOrNull { it == path.authority } and .single { it.subjectKeyIdentifier == ... } mean a RICAL containing duplicate infos, or two infos sharing an SKI, degrades to Invalid/throws. And the candidate loop runs a full chain validation (with signature verification) per certificateInfo, inside a live BLE exchange with timing constraints.

Fixed a real ordering defect: when both intermediate and root were anchors, the depth heuristic could tie and choose the root’s constraints. Selection now uses the actual validated path, keeps the highest applicable anchor, and applies only the first matching authority from the reader upward. Incompatible maximal paths are rejected. A real signed-certificate-chain regression checks both entry orders, both intermediate-anchor flags and the returned highest-anchor path; core evaluator tests separately cover constraint rejection. Duplicate SKIs are already rejected by the RICAL model; the redundant SKI lookup was removed. Path building is now per explicit anchor rather than per every CertificateInfo, but no unmeasured BLE latency improvement is claimed. [RicalX509Validation.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/credentials/waltid-mdoc-proximity/src/commonMain/kotlin/id/walt/mdoc/proximity/RicalX509Validation.kt#L77), [MobileWalletProximityReaderTrustTest.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonTest/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustTest.kt#L265).

**15.**

> CertificateRevocationListVerifier.kt: requireCrl(verifier != null, ...) then verifier!!. Should be ?: throw CrlValidationException(CrlFailure.UNAVAILABLE).

Replaced the nullable check plus `!!` with an Elvis expression that throws the existing `UNAVAILABLE` validation failure. [CertificateRevocationListVerifier.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/revocation/CertificateRevocationListVerifier.kt#L78).

**16.**

> CertificateRevocationListVerifier.kt and MobileWalletProximityCrlRevocation.kt: catch (_: Throwable) swallows Errors and turns any implementation bug into a silent Indeterminate. catch (_: Exception) at minimum.

Narrowed the CRL recovery catches, preserving cancellation and unexpected `Error` propagation. Signum’s `Asn1Exception` extends `Throwable` directly, so malformed certificate/extension input is normalized at the X.509 provider boundary and CRL DER failures are handled explicitly. Unsupported platform operations reported as `NotImplementedError` remain explicit recoverable cases. A fetch-provider `AssertionError` regression and the malformed-input Swift bridge regression cover both sides of that boundary. [SignumCertificateParser.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/signumMain/kotlin/id/walt/certificate/x509/signum/SignumCertificateParser.kt#L21), [MobileWalletProximityCrlRevocationTest.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonTest/kotlin/id/walt/wallet2/mobile/MobileWalletProximityCrlRevocationTest.kt#L115).

**17.**

> CrlCertificateStatus.Revoked(revocationTime, reasonCode: Int?) exposes a raw Int in a public API where an enum belongs.

Kept the existing numeric reason API in this bounded fix. The DER parser already validates the supported reason values; the wallet projects revoked status without exposing a reason integer to the holder. A typed reason is a reasonable follow-up API cleanup, but replacing the public type is not needed to correct trust or revocation behavior here. [CertificateRevocationListVerifier.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/revocation/CertificateRevocationListVerifier.kt#L23).

**18.**

> CertificateRevocationListVerifier.kt: a bad certificate signature is reported as CrlFailure.INVALID_ISSUER. Misleading category.

Kept the distinction: `INVALID_ISSUER` means the supplied issuer did not authenticate the target certificate; `INVALID_SIGNATURE` means the CRL’s own signature failed. Both remain indeterminate and neither can produce a good status. No trust behavior change was needed. [CertificateRevocationListVerifier.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/revocation/CertificateRevocationListVerifier.kt#L72).

**19.**

> MobileWalletProximityReaderTrustSettings.kt: require(existingFingerprints.add(f) && importedFingerprints.add(f)) relies on short-circuited side effects; importedFingerprints is redundant because existingFingerprints is already mutated in the loop.

Removed the redundant imported-fingerprint set. One seen set now detects both previously configured certificates and duplicates within the same import. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L265).

**20.**

> MobileWalletProximityReaderTrustSettings.kt: every bundle certificate is parsed twice and the two lists are correlated positionally by index.

Each bundle anchor now produces one validated record containing its preview and stored representation. The result lists come from those records, avoiding a second parse and positional/index correlation. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L505).

**21.**

> MdocReaderAuthentication23220EkuOid (IsoMdocReaderAuthenticationCertificate.kt) is a public const with a KDoc calling it "recommended" that is never used anywhere.

Kept the public OID as an optional producer-facing profile constant. It is not a mandatory acceptance check, and I did not manufacture one merely to make the constant internally referenced. [IsoMdocReaderAuthenticationCertificate.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/x509/IsoMdocReaderAuthenticationCertificate.kt#L28).

**22.**

> DemoReaderTrustSettings.kt appears to use en dashes

No separate typography rewrite was needed. The settings copy now uses localization resources consistently; Unicode punctuation is not a certificate-validation issue. [DemoReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/mobileMain/kotlin/id/walt/walletdemo/compose/ui/DemoReaderTrustSettings.kt#L30).

**23.**

> MobileWalletProximityReaderTrustSettings.kt rejects .p12/.pfx by filename and looks for the literal text "PRIVATE KEY" in a decodeToString() of up to 1 MiB of binary. Renaming a file defeats the first; a DER PKCS#8 key defeats the second. The real protection is that non-certificates fail to parse. The 1 MiB string allocation per DER import is a cost.

The filename/text checks remain early user-facing diagnostics, not the acceptance boundary. Successful import still requires the strict certificate/PEM/bundle grammar, so a renamed container or DER private key does not become an accepted CA. The documentation now describes the bounded import and its positive parsing checks without claiming those preliminary hints alone provide security. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L239).

**24.**

> Similarly, decode()'s validateStoredShape only re-parses structure - it does not re-check CA usage or the RICAL signature, while the description states "invalid stored settings are rejected during decode".

Clarified the three stages: decode validates persisted structure; import validates the supplied material; session evaluation performs current certificate/path/RICAL trust checks. Loading a stored record is not a trusted result. Decode remains deterministic and does not start network revocation checks or silently discard unrelated settings. [MobileWalletProximityReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrustSettings.kt#L214).

**25.**

> WalletDemoQRCodeCore.mm: CGImageCreate(..., /*shouldInterpolate*/ true, ...) on a QR bitmap. Masked by .interpolation(.none) in SwiftUI, but wrong at the image level and contradicts whole physical-pixel modules claim, should be false.

Set `shouldInterpolate` to `false` in the CGImage itself. SwiftUI also retains nearest-neighbor rendering and whole-module scaling. [WalletDemoQRCodeCore.mm](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoQRCodeCore/WalletDemoQRCodeCore.mm#L33).

**26.**

> WalletDemoQRCodeCore.mm: std::string ascii(utf8) truncates at an embedded NUL, which validateProximityQrCodePayload (QrCodeImage.kt:21, 0..0x7F) and WalletQRCodeRenderer.swift both permit. Unreachable for base64url in practice, but it's not "fails closed instead of truncating".

Both validators now reject NUL, whitespace and other control bytes. The native writer also receives an explicit byte length. Android/iOS/native regression cases cover those inputs, and the QR fingerprint/round-trip checks preserve the intended payload. [QrCodeImage.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/mobileMain/kotlin/id/walt/walletdemo/compose/ui/components/QrCodeImage.kt#L17), [WalletQRCodeRenderer.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/WalletQRCodeRenderer.swift#L9).

**27.**

> WalletDemoQRCodeCore.mm uses ZXing::MultiFormatWriter, which is [[deprecated("use CreateBarcode()")]] in the pinned v3.1.1 (core/src/MultiFormatWriter.h). The non-deprecated CreateBarcodeFromText + WriteBarcodeToImage would work, and is actually what the Kotlin/Native side uses.

Replaced the deprecated writer with `CreateBarcodeFromText` plus `WriteBarcodeToImage`, using the pinned dependency. The private adapter now explicitly requests C++17, as required by those headers; no dependency upgrade was introduced. [WalletDemoQRCodeCore.mm](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoQRCodeCore/WalletDemoQRCodeCore.mm#L19).

**28.**

> QrCodeImage.ios.kt: throw OutOfMemoryError() for a null data pointer - wrong exception type.

Replaced the inaccurate `OutOfMemoryError` with a descriptive missing-image-data failure. Rendering still fails closed through the existing UI fallback. [QrCodeImage.ios.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/iosMain/kotlin/id/walt/walletdemo/compose/ui/components/QrCodeImage.ios.kt#L70).

**29.**

> Also readBytes(width * height) assumes rowStride == width with no check.

Kept the owned-image manual copy/delete path: the pinned API produces tightly packed luminance images here. The code now states that format/ownership assumption; switching to the wrapper’s image-data accessor would introduce its ownership problem. The Objective-C++ adapter accesses pixels through the image view, which respects its stride. [QrCodeImage.ios.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/iosMain/kotlin/id/walt/walletdemo/compose/ui/components/QrCodeImage.ios.kt#L66).

**30.**

> CreatorOptions.options = "ecLevel=L" (QrCodeImage.ios.kt) is a no-op in the non-Zint build: CreateBarcode.cpp only forwards single-digit "0"-"8". The outcome is still L because that is the writer default, and Zint builds do parse "L" - but the option is not doing what it looks like it does.

Changed the option to `ecLevel=1`, the numeric form recognized by the pinned non-Zint writer, on both native paths. Existing exact QR fingerprints and decoder tests verify the resulting matrix/payload. [QrCodeImage.ios.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/iosMain/kotlin/id/walt/walletdemo/compose/ui/components/QrCodeImage.ios.kt#L54).

**31.**

> ColdCertificateSignatureValidationTest.kt mutates process-global java.security.Security state; it will flake other tests running concurrently in the same JVM. The fix it covers (BouncyX509CertificateSigner.kt, instance provider instead of setProvider("BC")) is good.

Moved the provider-removal check into an isolated JVM subprocess. The parent test process’s security-provider registry is untouched, while the child still verifies cold certificate/CSR signing and validation without a registered BC provider. [ColdCertificateSignatureValidationTest.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/crypto/waltid-x509/src/androidHostTest/kotlin/id/walt/certificate/x509/ColdCertificateSignatureValidationTest.kt#L22).

**32.**

> ISO 18013-5 Table B.6 requires issuerAlternativeName (rfc822Name or URI) when the reader certificate is issued by an IACA. Not validated, although IssuerAlternativeName.kt already exists in the module.

Added explicit application-supplied IACA issuer context in Kotlin and Swift. When required, the validated direct issuer must match that exact certificate and the reader must contain a non-critical issuerAlternativeName with a nonblank email or URI. Existing X.509 extension parsing is reused. Tests cover missing, critical, wrong-form and valid contacts, a generic CA, wrong/unknown required issuer and the real Swift bridge. No IACA role is inferred from a self-signature/name, and default validation does not claim the conditional check without that context. [MobileWalletProximityReaderTrust.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/src/commonMain/kotlin/id/walt/wallet2/mobile/MobileWalletProximityReaderTrust.kt#L260), [ProximityCRLBridgeTests.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosAppTests/ProximityCRLBridgeTests.swift#L6).

**33.**

> Cross-platform accessibility identifiers diverge for the new settings surface: settingsReaderPolicyAllowUntrusted/settingsReaderTrustImport/... (WalletUiTestTags.kt) vs readerTrustAllowUntrusted/readerTrustImport/... (WalletAccessibilityID.swift). The proximity* ones do match.

Aligned the shared settings identifiers for policy choices, import/review/confirm/cancel, reset and errors with Compose. Platform-only controls retain their own identifiers. [WalletAccessibilityID.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/WalletAccessibilityID.swift#L22).

**34.**

> SharingReviewView.swift now passes String to Button(_:action:), selecting the StringProtocol overload and de-localising Share/Cancel/Reject in the existing online flow too. submitTitle returns the same value on both branches.

The shared Swift button titles now resolve localization keys from the package resource bundle. This covers the existing online Share/Reject/Cancel flow as well as proximity Approve/Decline/Cancel. Removed the identical submit-title branches. [SharingReviewView.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-shared-ios/Sources/WalletDemoSharingUI/Components/SharingReviewView.swift#L383).

**35.**

> Hardcoded unlocalised strings in Compose despite the new 95-entry strings.xml: "Verifier" (WalletDemoProximityScreen.kt), "Device engagement QR code"- note proximity_qr_accessibility is the only unused entry in strings.xml - and the whole reader-trust settings screen (DemoReaderTrustSettings.kt).

Moved reader-trust settings copy and the remaining Verifier/QR accessibility labels into Compose resources. The QR now uses `proximity_qr_accessibility`. Technical diagnostic details can still come from the SDK; they are not presented as translated strings. [DemoReaderTrustSettings.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/mobileMain/kotlin/id/walt/walletdemo/compose/ui/DemoReaderTrustSettings.kt#L30).

**36.**

> ProximityPresentationView.swift uses MainActor.assumeIsolated in deinit; it traps rather than restoring brightness if the last reference is released off-main.

Removed `MainActor.assumeIsolated` from deinit. The view owner explicitly restores screen policy when the presentation view disappears; restoration stays on its UI lifecycle boundary. [ProximityPresentationView.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/ProximityPresentationView.swift#L969).

**37.**

> PresentView.swift cancels the session for any scenePhase != .active (Control Center, app switcher, system alerts), whereas Compose iOS correctly observes UIApplicationDidEnterBackgroundNotification (ProximityPlatformHooks.ios.kt:61).

Reused the existing background-only interruption policy and placed it in #2160. Transient inactive phases, such as system permission alerts, do not by themselves cancel the native session. [PresentView.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/Views/PresentView.swift#L8).

**38.**

> Compose iOS in turn never re-applies keep-awake/brightness on foreground (no WillEnterForeground observer), unlike Android's ON_START.

Compose iOS now restores brightness/keep-awake on background/disposal and reapplies the current active policy on activation. Observers are disposed with the effect and original screen values are restored when the journey ends. [ProximityPlatformHooks.ios.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedUI/src/iosMain/kotlin/id/walt/walletdemo/compose/ui/ProximityPlatformHooks.ios.kt#L24).

**39.**

> ProximityPlatformHooks.ios.kt and ProximityPresentationViewModel.swift report remediation .completed as soon as Settings is opened, so prerequisites are re-checked while the user is still in Settings. Android correctly awaits the activity result.

Reused the return-from-Settings implementation and placed the general repair in #2160 for both iOS hosts. Opening Settings no longer means remediation completed: rechecking waits for app activation, handles an unsuccessful open, and removes the observer on completion/cancellation. [ProximityPresentationViewModel.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/ProximityPresentationViewModel.swift#L690).

**40.**

> WalletDemoProximityScreen.kt hardcodes enabled = true on the approve button, so a double tap dispatches Approve twice and shows "Action could not be completed" after a successful disclosure.

Added a pending review ID in both hosts. The first approve/decline reserves that review synchronously, disables further decisions/selection changes, and keeps Cancel available. A rejected action re-enables retry; a reply for an old review cannot overwrite a newer review. Tests cover double/conflicting dispatch, late rejection, and a following review. [WalletDemoProximityController.kt](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/mobileMain/kotlin/id/walt/walletdemo/compose/logic/WalletDemoProximityController.kt#L71), [ProximityPresentationViewModel.swift](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-applications/waltid-wallet-demo-ios/iosApp/iosApp/ViewModels/ProximityPresentationViewModel.swift#L52).

**41.**

> ProximityReaderTrustConfiguration and ProximityRicalConfiguration are data classes holding fun interface instances, so equals/hashCode are identity-based and meaningless; snapshot() is a shallow copy and the name oversells what it detaches.

Clarified the snapshot contract: certificate/policy collections are copied, while application-provided evaluators/providers/fetchers remain live service references. Data-class equality is not a semantic comparison of those services. Deep-copying callbacks would not be meaningful, so the behavior and API names are retained with narrower documentation. [README.md](https://github.com/walt-id/waltid-identity/blob/e7043f2d9260e20ad0a28e43b6d524706a1c07d9/waltid-libraries/protocols/waltid-openid4vc-wallet-mobile/README.md#L414).

waltkb (approved):

JakeFernandes98 (approved):
Approving as part of compressing the proximity stack into https://github.com/walt-id/waltid-identity/pull/2168 against main. Not merging into main until CI is fully green.

JakeFernandes98:
Closing this stacked PR to stop duplicate CI. The full proximity stack now lands in a single PR against main: https://github.com/walt-id/waltid-identity/pull/2168

PR #2169, opened 2026-08-31
feat(issuer2): configure mdoc MSO validity via msoData

JakeFernandes98:
## Summary

This adds Issuer2 `msoData` so mDoc issuance can set ISO 18013-5 MSO `ValidityInfo` (`validFrom`, `validUntil`, optional `expectedUpdate`) at claim time. Operators can now choose a validity window other than the hardcoded ~365-day default, and timestamp data functions work for MSO dates the same way they already do for W3C `mapping`.

W3C `mapping.validFrom` / `validUntil` and mDL namespace `issue_date` / `expiry_date` are still not MSO fields. That mismatch is what [WAL-1365](https://linear.app/walt-new/issue/WAL-1365/support-unable-to-configure-validfrom-and-validuntil-for-mdoc) and [WAL-1015](https://linear.app/walt-new/issue/WAL-1015/mdocs-validuntilvalidfrom) reported.

Related PRs:
- [walt-id/waltid-identity-enterprise#650](https://github.com/walt-id/waltid-identity-enterprise/pull/650) (depends on this `MsoData` / `MsoValidityResolver` surface)
- [walt-id/waltid-docs#374](https://github.com/walt-id/waltid-docs/pull/374)

### Public API
- New optional `msoData` on Issuer2 profiles, `runtimeOverrides`, and issuance sessions. Accepted only for `mso_mdoc`; rejected on other formats.
- Each of `validFrom`, `validUntil`, and `expectedUpdate` is a static ISO-8601 timestamp or a timestamp data function, resolved at claim time. `signed` is not configurable.
- Offer overrides field-merge onto the profile (`validFrom`, `validUntil`, and `expectedUpdate` independently).
- Authorization-code requests without `issuer_state` (including offerless) copy profile `msoData` onto the session.
- Defaults are unchanged: `validFrom` equals the MSO signed time, `validUntil` is 365 days after claim, `expectedUpdate` is omitted.
- Window rules are checked after ISO tdate second-normalization: `validUntil` must be after `validFrom`; `validFrom` cannot be before signed time; if set, `expectedUpdate` must fall between them.
- Issuer/session policy is authoritative. Holder `requestForm["validUntil"]` does not change the signed MSO window.

### Shared issuance
- `MsoData` and claim-time `MsoValidityResolver` live in openid4vci so OSS and Enterprise share one resolution path.
- Resolver order for `validUntil`: explicit `msoData.validUntil` → optional `fallbackValidUntil` → 365-day default. Enterprise uses the fallback for persisted `IssuanceRequest.validUntil`.
- `expectedUpdate` is threaded through `OAuth2Provider.createCredentialResponse`, `MdocCredentialSigner.generateMdocCredential`, `MdocIssuer`, and both credential-handler `sign` overloads. W3C and SD-JWT handlers accept the argument and ignore it.
- `MsoMdocCredentialHandler.issueMdoc` now receives the resolved instants instead of hardcoding `validityDays = 365`.
- `ValidityInfo.precheck()` requires `validUntil > validFrom` and checks the `expectedUpdate` window. If a provided `validFrom` falls a moment behind the sign timestamp (claim-time resolve vs. signing), it is raised to `signed` so ISO `validFrom >= signed` still holds; a future `validFrom` is kept.

### Issuer2 wiring
- Profile load rejects `msoData` unless metadata format is `mso_mdoc`.
- Offer creation stores the merged `msoData` on the session.
- Offerless / no-`issuer_state` authorization sessions copy profile `msoData`.
- The mDoc claim path resolves session `msoData` and passes the three instants into `createCredentialResponse`. Non-mDoc sessions that still carry `msoData` fail closed.

### Tests
- Resolver coverage for merge, defaults, static ISO-8601, timestamp functions, invalid windows, sub-second tdate collapse, and `fallbackValidUntil`.
- Issued-MSO `expectedUpdate` on the crypto2 mDoc issuer path.
- Handler coverage that holder `requestForm["validUntil"]` cannot extend a configured window, and that `MsoMdocCredentialHandler` forwards the resolved instants.
- Profile mapping/rejection plus Issuer2 wallet flows for pre-authorized issuance, profile-level `msoData`, offerless scope and authorization_details sessions, and authorization-code with and without `issuer_state`.

## Architecture Notes

- `msoData` is nested, not top-level `validFrom` / `validUntil`, so it cannot be confused with W3C mapping or namespace claims.
- Resolution runs at claim, not offer creation, matching existing data-function timing.
- The library change is the contract Enterprise #650 consumes; land or pin this PR first.

## Caveats and Follow-Ups

- Legacy Issuer (`waltid-issuer-api`) is unchanged.
- `signed` remains issuer-generated.
- Static `validFrom` must still be at or after claim/sign time; a timestamp captured at offer creation and claimed later is rejected.

## Breaking

- `CredentialEndpointHandler` / `Crypto2CredentialEndpointHandler` `sign` now require `expectedUpdate: Instant?`. Fun-interface implementors outside this repo must add the parameter.
- `OAuth2Provider.createCredentialResponse` inserts `expectedUpdate: Instant? = null` before `proofValidationContext`. Named-argument callers are source-compatible; positional callers that passed `proofValidationContext` after `validUntil` must be updated.
- `MdocCredentialSigner.generateMdocCredential` and `MdocIssuer` overloads add `expectedUpdate: Instant?`.
- `MsoMdocCredentialHandler.issueMdoc` now takes `validFrom`, `validUntil`, and `expectedUpdate` (with `validityDays` as a fallback default). External subclasses must update their override.
- Existing mDoc issuance without `msoData` keeps the previous defaults. No issued-credential format change when `msoData` is omitted.


waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
```diff
@@ -1100,6 +1109,9 @@ class OpenId4VciProtocolService @JvmOverloads constructor(
                     mDocNameSpacesDataMappingConfig = session.mDocNameSpacesDataMappingConfig,
                     authorizedTransactionDataTypes = session.authorizedTransactionDataTypes,
                     credentialStatus = mDocStatus,
+                    validFrom = resolvedMsoValidity?.validFrom,
+                    validUntil = resolvedMsoValidity?.validUntil,
```
szijpeter, line 1113: **[P1] Keep the issuer-configured expiry authoritative**

This configured value is subsequently passed to `MdocCredentialHandler.resolveValidUntil()`, which prefers the holder-controlled `request.requestForm["validUntil"]`. I reproduced a credential request replacing a configured one-year expiry with a roughly ten-year expiry in the signed MSO. Please make the server/session policy authoritative—or explicitly bound the legacy request extension—and add regression coverage.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
```diff
@@ -1077,6 +1078,14 @@ class OpenId4VciProtocolService @JvmOverloads constructor(
                     else -> null
                 }
             }
+            val resolvedMsoValidity = if (configuration.format == CredentialFormat.MSO_MDOC) {
+                MsoValidityResolver.resolve(session.msoData)
```
szijpeter, line 1082: **[P1] Preserve `msoData` in offerless authorization sessions**

For Authorization Code requests without `issuer_state`, `toAuthorizationCodeSession()` copies the profile fields but omits `msoData`. I reproduced an offerless mdoc session with `session.msoData == null`, so this resolver uses the 365-day default instead of the profile configuration. Please copy `msoData` into that session and cover the offerless flow.

waltid-libraries/credentials/waltid-mdoc-credentials2/src/commonMain/kotlin/id/walt/mdoc/issuance/MdocIssuer.kt
```diff
@@ -151,7 +159,8 @@ object MdocIssuer {
                 signed = signedTimestamp,
                 validFrom = effectiveValidFrom,
                 // ISO 18013-5 MSO tdate fields must not include fractional seconds
-                validUntil = Instant.fromEpochSeconds(validUntil.epochSeconds)
+                validUntil = Instant.fromEpochSeconds(validUntil.epochSeconds),
```
szijpeter, line 162: **[P2] Validate the normalized TDate window**

The resolver validates full-precision instants before these values are truncated to seconds. I reproduced `validFrom=.100Z` and `validUntil=.900Z` passing validation and then serializing as equal timestamps, while `ValidityInfo.precheck()` accepts equality. Please validate the normalized values and retain the stated `validUntil > validFrom` invariant.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MsoMdocCredentialHandler.kt
```diff
@@ -52,6 +52,7 @@ abstract class MsoMdocCredentialHandler : CredentialEndpointHandler {
         credentialStatus: Status?,
         validFrom: Instant?,
         validUntil: Instant?,
+        expectedUpdate: Instant?,
```
szijpeter, line 55: **[P2] Do not accept validity values that this handler drops**

This public integrator handler now accepts `validFrom`, `validUntil`, and `expectedUpdate`, but `sign()` still calls `issueMdoc()` with only `validityDays = 365`. Please extend the signing contract to receive the resolved values, or intentionally deprecate/remove this path; otherwise custom handlers silently ignore the new configuration.

szijpeter (changes requested):
The core `MsoData`/resolver design is sensible, but two supported issuance paths bypass or lose the configured policy and two public validity contracts remain incomplete; please also confirm the release decision for the binary-breaking `expectedUpdate` insertions in `OAuth2Provider`, `MdocCredentialSigner`, and `MdocIssuer`, which are only partially covered by the PR's Breaking section.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Fixed. `MdocCredentialHandler` now uses only the issuer/session `validUntil` (else the 365-day default). Holder `requestForm["validUntil"]` is ignored.

Regression: `MdocCredentialHandlerValidUntilAuthorityTest` issues with a ~10-year form expiry and asserts the signed MSO keeps the configured one-year window.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Fixed. `toAuthorizationCodeSession()` now copies profile `msoData`, so offerless / no-`issuer_state` auth-code sessions no longer fall back to the 365-day default.

Coverage in `Issuer2MsoDataWalletFlowTest`: offerless scope and authorization_details session copy, auth-code with `issuer_state` (offer merge), and auth-code without `issuer_state` (profile copy).

waltid-libraries/credentials/waltid-mdoc-credentials2/src/commonMain/kotlin/id/walt/mdoc/issuance/MdocIssuer.kt
JakeFernandes98: Fixed. `MsoValidityResolver` now truncates to tdate seconds before window checks, so `validFrom=.100Z` / `validUntil=.900Z` is rejected. `ValidityInfo.precheck()` now requires `validUntil > validFrom` (and the inverted error message is corrected). `MdocIssuer` re-checks that invariant after second-normalization.

Resolver and `MdocTDateFormatTest` cover the equal-second case.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MsoMdocCredentialHandler.kt
JakeFernandes98: Fixed by extending the integrator contract rather than removing the path. `MsoMdocCredentialHandler.issueMdoc` now receives `validFrom`, `validUntil`, and `expectedUpdate`; `validityDays` remains only as a fallback default. `sign()` forwards the resolved instants.

`MsoMdocCredentialHandlerValidityTest` records those values. This is an additional binary break for external subclasses; it is now listed in the Breaking section.

JakeFernandes98:
@szijpeter Thanks — all four inline items are addressed on `3616f1114`.

On the binary-breaking `expectedUpdate` insertions: they stay. Named-argument callers remain source-compatible; positional callers and fun-interface / abstract `issueMdoc` implementors need the new parameter. The Breaking section now lists `OAuth2Provider.createCredentialResponse`, `MdocCredentialSigner.generateMdocCredential`, `MdocIssuer`, and `MsoMdocCredentialHandler.issueMdoc` in addition to the handler `sign` overloads.

szijpeter (approved):

PR #2175, opened 2026-09-01
Feat/wal 1321 fixes

philpotisk:
## Summary

`verifier2`'s `configureMonitoring()` installs Ktor's `CallId` plugin with a header retriever and a
verifier, but no generator. A request that arrives without `X-Request-ID` therefore ends up with no
call id at all.

For verifier2's own routes that is invisible: no handler reads `call.callId`, it only feeds the
`callIdMdc` log field. It becomes a hard failure in services that reuse this function while also
serving `issuer-api2`'s OpenID4VCI controllers, because those call
`requireNotNull(call.callId) { "Missing call ID" }` in every handler. Public OAuth clients do not
send `X-Request-ID`, so every token request is rejected with a `400` whose body carries no `error`
field — the client then reports an opaque HTTP 400 with no OAuth error code.

This PR adds the missing generator so the shared function is safe for both consumers.

Related PRs (fill in numbers when opened):
- `walt-id/waltid-identity-enterprise` — `feat/wal-1321-fixes`, unique WALLET_CREATED usage ids
- `walt-id/waltid-identity-enterprise-license` — `feat/wal-1321-fixes`, regression test for the token endpoint

### waltid-verifier-api2
- `install(CallId)` in `configureMonitoring()` now has `generate { Uuid.random().toString() }`, so an
  absent `X-Request-ID` yields a generated id instead of none. A supplied header still wins, and the
  existing `verify` block is unchanged.

## Architecture Notes

- This aligns verifier2 with the two other copies of the same block:
  `issuer-api2`'s `Monitoring.kt` and `waltid-enterprise-api`'s `Main.kt` both already generate.
  verifier2 was the only one that did not, which is what made the behaviour depend on which of the
  three a composing service happened to import.
- Note for context: `main`'s license server no longer trips over this because it was switched to
  issuer2's monitoring plugin in `44e6ac1`. That commit is not on `feat/wal-1321`, which is why the
  branch build still had the failing behaviour. This PR fixes the shared function itself so the trap
  is gone regardless of which monitoring setup a service picks.

## Caveats and Follow-Ups

- Three near-identical `CallId` + `CallLogging` blocks still exist across issuer-api2, verifier2 and
  waltid-enterprise-api. Consolidating them into one shared function in `waltid-commons` would remove
  the drift that caused this; deliberately out of scope here.

## Breaking

- None. The only observable change for verifier2 itself is that log lines and the `X-Request-ID`
  response header now carry a value where they previously carried none.


waltkb (approved):
Yes, this makes sense when combining both Verifier and Issuer OpenID4VCI routes in a service

philpotisk:
resolved by https://github.com/walt-id/waltid-identity/pull/2158

PR #2177, opened 2026-09-01
Fixes for ios Distribution

JakeFernandes98:
some fixes from our recent testflight releases


szijpeter (approved):
Looks good for private TestFlight use — approved. Two non-blocking notes:

- The native 1024 px icon is still RGBA; consider re-exporting it without an alpha channel if App Store Connect rejects the upload ([Apple guidance](https://developer.apple.com/forums/thread/795411)).
- Before public distribution, remove the standalone Apple logo unless it is covered by an express written license ([Apple trademark guidelines](https://www.apple.com/legal/intellectual-property/guidelinesfor3rdparties.html)).

Issue #657, opened 2024-07-31
Java usage during migration to crypto-jvm


vanhoanHoang:
Hi team, 

Using waltid-crypto-jvm version 0.3.1 in Java application, we are able to resolve serialized json key via: KeyManager.INSTANCE.resolveSerializedKey(jwk). Since the latest relases, this function could not be found in Java. I could not find any other function which could do the same in the library. 

Most of them for example importJwkSync and those in KeySerialization only return Result<Key> which is not accessible in Java. Could you please have a look to bring this feature back, which is very appeciated. 

Cheer, 
Hoan

PR #2189, opened 2026-09-04
Feat/issuer batch support

chsavvaidis:
## Summary

This adds OpenID4VCI 1.0 batch issuance to Community Stack issuer-api2 so a wallet can receive several credentials from one offer, and can request a homogeneous cryptographic batch (multiple holder proofs) on a single Credential Endpoint call.

It implements the OSS issuer slice of [WAL-1290](https://linear.app/walt-new/issue/WAL-1290/add-support-for-batch-issuance-epic). Related PRs: [walt-id/waltid-identity-enterprise#657](https://github.com/walt-id/waltid-identity-enterprise/pull/657). Wallet-api2, wallet-sdk, and Compose consume are not in this change.

### Management API
- `POST /issuer2/credential-offers` now takes `credentials: [{ profileId, runtimeOverrides? }]` instead of a top-level `profileId`.
- The offer response lists those credentials with their `credentialConfigurationId`s.
- Runtime overrides, including `expectedCredentialProofKeyJwk`, are per credential.

### OpenID4VCI protocol
- Credential issuer metadata can advertise `batch_credential_issuance.batch_size` via `issuer-service.conf` `batchCredentialIssuance`.
- `POST /openid4vci/credential` accepts `proofs.jwt` arrays. When batch is enabled, the issuer signs one credential per verified proof, in order, and returns `{ "credentials": [...] }`.
- Phase one issues the same configuration and dataset once per proof. JWT proofs only. Sessions with a preconfigured `credentialStatus` stay single-issuance, because each batched credential would need its own status entry.

### Session model
- An issuance session stores `issuanceRequests[]` and `issuanceResults` keyed by `credentialIdentifier`.
- Multi-credential offers stay `ACTIVE` until every requested credential is issued. The session is not closed on first success, so a wallet can request remaining configurations (and, today, retry an already-issued identifier until expiry).

## Architecture Notes

- Proofs are verified as a set, then `CredentialIssuanceInputProvider` supplies one input per proof, then the format handler signs the batch in one call.
- Distinct `credential_configuration_id`s in one offer are separate Credential Endpoint requests. A homogeneous batch is multiple proofs for **one** of those targets.
- Two datasets that share a configuration ID are ambiguous unless the wallet uses `authorization_details` / `credential_identifier`.

## Caveats and Follow-Ups

- WAL-1290 wallet-api2, Enterprise wallet, wallet-sdk, and Compose support are out of scope.
- Default `issuer-service.conf` currently sets `batchCredentialIssuance.batchSize = 10`, so sample deployments advertise batch unless that block is removed.
- License-server still compiles against the previous offer/session shape; current-head unified gradle is red until that caller is updated with the Enterprise counterpart.
- Product docs are being updated separately in waltid-docs.

## Breaking

- Offer create JSON must use `credentials[]`. Top-level `profileId` / `runtimeOverrides` no longer deserialize.
- Session JSON and stored session documents use `issuanceRequests` / `issuanceResults` instead of a single flat credential.
- `OAuth2Provider.createCredentialResponse` now takes `CredentialIssuanceInputProvider` instead of a single `credentialData` object.
- Draft 13 singular `proof` is not accepted; wallets must send `proofs`.


waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/models/CredentialOfferModels.kt
```diff
@@ -32,19 +32,50 @@ data class CredentialOfferRuntimeOverrides(
...
+    val runtimeOverrides: CredentialOfferRuntimeOverrides? = null,
+) {
+    init {
+        require(profileId.isNotBlank()) { "profileId must not be blank" }
+    }
+}
+
+@Serializable
+data class CredentialOfferCreateRequest(
+    val credentials: List<CredentialOfferCredential>,
```
JakeFernandes98, line 46: **Blocking — keep offer create additive.** `credentials` cannot be the only accepted shape.

Please restore deprecated top-level `profileId` and `runtimeOverrides`, keep `credentials[]` optional, and require at least one of `profileId` or `credentials`. Existing non-batch clients must keep working; batch/multi-credential can be the new array path.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/models/CredentialOfferModels.kt
```diff
@@ -60,18 +91,24 @@ data class CredentialOfferCreateRequest(
...
+data class CredentialOfferCredentialResponse(
+    val profileId: String,
+    val credentialConfigurationId: String,
+)
+
 @Serializable
 data class CredentialOfferCreateResponse(
     val offerId: String,
-    val profileId: String,
+    val credentials: List<CredentialOfferCredentialResponse>,
```
JakeFernandes98, line 111: **Blocking — do not change the create-offer response shape.** Keep the existing response (`offerId`, `profileId`, `authMethod`, …). Do not return `credentials[]` or `credentialConfigurationId`.

`credentialConfigurationId` is not needed on this management response; the wallet learns it from the OID4VCI offer/metadata.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/domain/IssuanceSession.kt
```diff
@@ -37,18 +37,41 @@ data class IssuanceSession(
...
+data class IssuanceResult(
+    val issuedAt: Instant,
+    val issuedCredentialFormat: String,
+)
+
+@Serializable
+data class IssuanceSession(
+    val sessionId: String,
+    val authenticationMethod: AuthenticationMethod,
+    val issuanceRequests: List<IssuanceRequest>,
```
JakeFernandes98, line 55: Even with an additive offer create API, renaming `issuanceRequest` → `issuanceRequests` / `issuanceResults` still breaks stored sessions and any client that reads session JSON or webhooks.

Please add a dual decoder / migration, or document that in-flight sessions are discarded on deploy (including `expiresInSeconds = -1`).

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
```diff
@@ -1195,24 +1298,38 @@ class OpenId4VciProtocolService @JvmOverloads constructor(
...
-                            statusReason = "Credential issued successfully",
-                            issuedCredentialFormat = configuration.format.value,
+                            issuanceResults = issuanceResults,
+                            status = if (issuanceComplete) {
+                                IssuanceSessionStatus.SUCCESSFUL
+                            } else {
+                                IssuanceSessionStatus.ACTIVE
+                            },
+                            statusReason = if (issuanceComplete) "Credentials issued successfully" else null,
                             isClosed = false,
```
JakeFernandes98, line 1325: After a successful issue, `status` becomes `SUCCESSFUL` but `isClosed` stays `false`. Tests show the same access token can mint the same credential again until `expiresAt`.

If retries are intentional, bound them. Otherwise close the session (or that `credentialIdentifier`) after first success.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/CredentialOfferService.kt
```diff
@@ -34,24 +36,55 @@ class CredentialOfferService(
...
             AuthenticationMethod.PRE_AUTHORIZED -> null
             AuthenticationMethod.AUTHORIZED -> request.issuerStateMode ?: IssuerStateMode.INCLUDE
         }
-        require(issuerKey.isNotEmpty()) { "issuerKey must not be empty" }
-        require(issuerKey["type"] != null) { "issuerKey must contain a key type" }
-
         var resolvedTxCodeValue: String? = null
+        val credentialConfigurationIds = resolvedCredentials
+            .map { it.second.credentialConfigurationId }
+            .distinct()
```
JakeFernandes98, line 87: Two datasets that share a `credential_configuration_id` collapse to one ID on the offer. A later Credential Request that uses `credential_configuration_id` then fails as ambiguous (`use credential_identifier`), and identifiers are only returned if the token request included `authorization_details`.

Please document this, reject the offer at create time, or always emit `authorization_details` for those offers.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/OAuth2Provider.kt
```diff
@@ -133,21 +133,19 @@ interface OAuth2Provider {
...
     ): CredentialRequestResult
 
-    @Deprecated("Use the Crypto2CredentialSigningKey overload")
     suspend fun createCredentialResponse(
         request: CredentialRequest,
         configuration: CredentialConfiguration,
         issuerKey: Key,
         issuerId: String,
-        credentialData: JsonObject,
+        issuanceInputData: CredentialIssuanceInputProvider,
```
JakeFernandes98, line 141: This signature change (`CredentialIssuanceInputProvider` instead of a single `credentialData`) is a library compile break independent of the additive offer JSON.

Please keep a source-compatible overload, or call out the library break and update in-repo callers in the same change.

waltid-services/waltid-issuer-api2/config/issuer-service.conf
```diff
@@ -4,6 +4,10 @@ ciTokenKey = """{"type":"jwk","jwk":{"kty":"EC","d":"KJ4k3Vcl5Sj9Mfq4rrNXBm2MoPo
 
 credentialEncryptionKey = """{"type":"jwk","jwk":{"kty":"EC","d":"ZSHgIcRvbwV9s224kHUaFqkEPShCAdwXocGl_w3M42Q","crv":"P-256","kid":"issuer2-credential-encryption-key","x":"GWKpdL3jPoPJ5wKgSA-jxS2jgp-ZUDE6sIQbeB86vF0","y":"F3xAwH96_xVciV7mFQslU_eRQgP-5pSZiNf8bjMoGfo"}}"""
 
+batchCredentialIssuance {
```
JakeFernandes98, line 7: README still describes batch as opt-in, but the sample conf advertises `batchSize = 10`. Copying this file enables protocol batch immediately.

Please either leave `batchCredentialIssuance` unset in the sample, or state in the README that the shipped conf enables it.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
```diff
@@ -1089,36 +1173,55 @@ class OpenId4VciProtocolService @JvmOverloads constructor(
...
-                    credentialData = credentialDataWithStatus,
-                    dataMapping = session.mapping,
-                    selectiveDisclosure = session.selectiveDisclosure,
-                    x5Chain = x5Chain,
-                    mDocNameSpacesDataMappingConfig = session.mDocNameSpacesDataMappingConfig,
-                    authorizedTransactionDataTypes = session.authorizedTransactionDataTypes,
-                    credentialStatus = mDocStatus,
-                    proofValidationContext = proofValidationContext,
+            val issuanceInputData = CredentialIssuanceInputProvider { issuanceCount ->
+                if (issuanceCount > 1 && issuanceRequest.credentialStatus != null) {
```
JakeFernandes98, line 1179: Protocol batch clones the same dataset and the same status object. Sessions with a preconfigured `credentialStatus` reject `issuanceCount > 1`.

That is a real constraint vs Enterprise (which allocates a status entry per proof). Please keep it explicit in issuer-api2 docs/OpenAPI so operators do not expect status-enabled batch here.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/DefaultOAuth2Provider.kt
```diff
@@ -813,6 +800,53 @@ class DefaultOAuth2Provider(
...
 
+    private fun validateCredentialProofBatch(
+        request: CredentialRequest,
+        context: CredentialProofValidationContext?,
+    ): CredentialError? {
+        val proofCount = request.credentialProofCount()
+        if (proofCount <= 1) return null
+        val batch = context?.batchCredentialIssuance
+            ?: return CredentialError(
+                CredentialErrorCodes.INVALID_PROOF,
```
JakeFernandes98, line 811: Over-size / disabled batch currently returns `invalid_proof`. OpenID4VCI 1.0 treats `batch_size` overflow as `invalid_credential_request`. Please align the error code if that is the intended spec reading.

JakeFernandes98 (changes requested):
## Requested changes

The offer management API must stay backward compatible for existing non-batch clients. Protocol batch (`proofs.jwt` / `batch_credential_issuance`) can remain opt-in.

**Offer create (blocking):** keep deprecated top-level `profileId` and `runtimeOverrides`, add optional `credentials[]`, and reject only when neither shape is present. The create-offer **response** must keep the current non-array shape (`profileId`, …). Do not return `credentials[]` or `credentialConfigurationId`.

That makes the offer JSON additive. Session persistence (`issuanceRequests` / `issuanceResults`) and the `OAuth2Provider.createCredentialResponse` signature remain separate breaks and still need an explicit plan.

Counterpart: https://github.com/walt-id/waltid-identity-enterprise/pull/657

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Still present in the latest review (head `8edbe4d`). `OpenId4VciProtocolService.kt:1327-1334` still sets `isClosed = false` on `SUCCESSFUL`, and there is no guard rejecting an already-issued `credentialIdentifier` (completion is only `establishedSelection.all { it in issuanceResults }`). The same access token can re-mint an identifier until `expiresAt`. Enterprise #657 rejects already-issued identifiers — please confirm the intended OSS/Enterprise split, or bound the retries.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/CredentialOfferService.kt
JakeFernandes98: Still present in the latest review. Two datasets sharing a `credential_configuration_id` collapse to one id on the offer; a wallet that then sends `credential_configuration_id` gets `invalid_credential_request` unless the token request used `authorization_details`. Please document this, or always emit `authorization_details` for such offers.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/OAuth2Provider.kt
JakeFernandes98: Re-review: we are **not** concerned about breaking external SDK consumers of `createCredentialResponse`, so the signature change itself is fine.

The real question was credential status. Dropping the `credentialStatus: Status?` parameter here is OK because status now flows per-instance through `CredentialIssuanceInputProvider` → `CredentialIssuanceInput.credentialStatus` (`OpenId4VciProtocolService.kt:1183-1191`), and preconfigured-status + batch is explicitly rejected via `UnsupportedBatchCredentialStatusException`. So status is relocated, not lost. Please just confirm this is the intended design (per-instance status via the provider) and that no non-mdoc status path regressed.

waltid-services/waltid-issuer-api2/config/issuer-service.conf
JakeFernandes98: Still present in the latest review — the sample `issuer-service.conf` still ships `batchCredentialIssuance { batchSize = 10 }`, so copying it enables protocol batch by default. Ship it unset or note this in the README.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Still present — sessions with a preconfigured `credentialStatus` reject `issuanceCount > 1` (`UnsupportedBatchCredentialStatusException`). Fine as a constraint, but keep it explicit in issuer-api2 docs/OpenAPI (status-enabled batch is Enterprise-only).

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/DefaultOAuth2Provider.kt
JakeFernandes98: Still present — over-size / disabled batch returns `invalid_proof`; OpenID4VCI 1.0 treats `batch_size` overflow as `invalid_credential_request`. Align if that is the intended reading.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
chsavvaidis: Intentional, see https://github.com/walt-id/waltid-identity-enterprise/pull/657#discussion_r4069021853

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/CredentialOfferService.kt
chsavvaidis: Documented in the Credential Endpoint OpenAPI description. When multiple authorized datasets share a configuration, clients must request authorization_details to obtain dataset identifiers, then use credential_identifier in a separate Credential Request for each dataset.

waltid-services/waltid-issuer-api2/config/issuer-service.conf
chsavvaidis: The README now explicitly states that the shipped configuration enables batch issuance with a maximum of 10 proofs per Credential Request. It also explains how to adjust the limit or remove batchCredentialIssuance to disable batching while retaining single-credential issuance.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/DefaultOAuth2Provider.kt
chsavvaidis: Fixed: disabled batching and requests exceeding batch_size now return invalid_credential_request in validateCredentialProofBatch. Invalid proof signatures still return invalid_proof. Regression tests cover both rejection cases and verify that issuance inputs are not requested

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
chsavvaidis: Fixed and documented in the docs/openapi

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/DefaultOAuth2Provider.kt
JakeFernandes98: Verified on `320892d`: `validateCredentialProofBatch` now returns `invalid_credential_request` for both disabled batch and over-`batch_size`; `invalid_proof` is reserved for bad signatures. Resolving.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Verified on `320892d`: `UnsupportedBatchCredentialStatusException` is gone; shared `credentialStatus` is now injected per format (W3C into `credentialData`, mdoc via `Status`) and covered by the new shared-status tests. Resolving.

waltid-services/waltid-issuer-api2/config/issuer-service.conf
JakeFernandes98: Confirmed — the README now documents that the shipped config enables batch (max 10 proofs) and how to lower or remove it. Resolving.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/CredentialOfferService.kt
JakeFernandes98: Confirmed — documented in the Credential Endpoint OpenAPI (request `authorization_details`, then use `credential_identifier` per dataset). Resolving as a docs clarification.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/core/OAuth2Provider.kt
JakeFernandes98: Confirmed intended: `credentialStatus` is relocated to the per-instance `CredentialIssuanceInputProvider`, and the new shared-status commits cover status across formats in batch. No SDK-external concern on our side. Resolving.

waltid-services/waltid-issuer-api2/src/main/kotlin/id/walt/issuer2/service/openid4vci/OpenId4VciProtocolService.kt
JakeFernandes98: Acknowledged as intentional (OSS re-issues until `expiresAt`; Enterprise rejects re-issue by design, cross-stack alignment tracked in WAL-1438). Leaving this open for explicit human sign-off since it is a token-reuse behavior (same access token can re-mint until expiry).

JakeFernandes98 (approved):
PR implementation is approved. Please revise the license PR for remaining change required.

Rebase family and have CI pass before merging.

PR #2190, opened 2026-09-04
feat(wallet): add Compose web demo against wallet-api2 auth

JakeFernandes98:
## Summary

This adds a no-install Compose Multiplatform web wallet demo against **wallet-api2 with auth enabled**, so a reviewer can run the docker identity stack (issuer2, verifier2, wallet-api2) and complete issuance and presentation in the browser.

The Wasm host is a custodial wallet-api2 client (email/password JWT, isolated `/wallet` HTTP). It does not run the mobile wallet SDK and it does not replace `waltid-web-wallet` (wallet-api v1). Android and iOS continue to use the SDK, PIN, and platform keys.

Ticket: [WAL-1198](https://linear.app/walt-new/issue/WAL-1198/initial-web-wallet-ui)

### wallet-api2
- Register the `ktor-authnz` Authentication provider **before** `WebService` installs the plugin. Without that, enabling `auth` fails at startup with a missing `ktor-authnz` configuration.
- CORS now allows GET, POST, PUT, DELETE, PATCH, HEAD, and OPTIONS so a browser client can call `.../keys/{id}/set-default` and `.../dids/{did}/set-default`.

### Docker identity profile
- Enable the `auth` feature on wallet-api2 and add a demo `auth.conf` JWT signing key (local compose only; same pattern as the other identity demo keys).
- Document that published `stable` (0.23.0) cannot load this `auth.conf`. Use `latest` or a locally built `1.0.0-SNAPSHOT` image for wallet-api2 when auth is on.

### Compose web demo
- Email/password register and login (`POST /auth/register`, `POST /auth/emailpass`). JWT in `localStorage`; wallet id in `localStorage` and a cookie.
- After login the wallet opens unlocked (no PIN/biometrics) and bootstraps immediately: reuse the default key/DID, or generate a secp256r1 JWK and `did:jwk`.
- Receive/present via pasted OpenID4VCI offer and OpenID4VP request URLs. Pre-authorized issuance uses `.../receive`; authorization-code issuance uses `authorization-url` plus `.../receive/authorized` and the current page origin as `redirect_uri`. Presentation is `/preview` → `/build-vp-token` → `/send-response`.
- Wasm-safe JSON serializers for `JsonObject`/`JsonElement`, and `index.html` now loads `webApp.js` so the webpack host actually boots.
- Settings hides lock, signing protection, and DC API when PIN is disabled, and Sign out is wired.

`enableWalletDemoComposeWeb` stays **false** by default.

## Architecture Notes

- Web is `wasmJs`. Protocol work stays on wallet-api2 HTTP rather than embedding the mobile SDK in the browser.
- Shared `WalletDemoController` gains `skipPin` and `issuanceRedirectUri` so Android/iOS keep the existing PIN default.
- Authorization-code issuance persists the pending session across the issuer redirect/reload.

## Screenshots

Compose web against docker identity issuer2, verifier2, and wallet-api2 with auth. Pre-authorized OpenID4VCI (`jwt_vc_json` / OpenBadgeCredential) and OpenID4VP. Wasm card titles currently show the credential format (`jwt_vc_json`) rather than the OpenBadge display name.

| Sign in | Empty wallet | Offer review |
| --- | --- | --- |
| Sign in | Empty wallet | Offer review |

| Stored credential | Presentation review | Presentation sent |
| --- | --- | --- |
| Stored credential | Presentation review | Presentation sent |

## Caveats and Follow-Ups

- QR scanning, Digital Credentials API, BLE/NFC, and hardware-backed keys are not available on web; paste URLs instead.
- Authorization-code issuance is implemented but was not exercised in this gallery (pre-authorized only).
- Android and iOS demo builds were not re-verified in this change.
- wallet-api2 auth accounts are in-memory in the current OSS auth module.

## Breaking

- wallet-api2 CORS now allows PUT and DELETE, which browser clients need for default key/DID setup.


waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
```diff
@@ -0,0 +1,354 @@
...
+    ): WalletDemoOperationResult {
+        val session = presentationSessions[previewHandle.value]
+        val built = client.buildVpToken(
+            walletId,
+            BuildVpTokenRequestDto(
+                requestUrl = previewHandle.value,
+                selectedCredentialOptions = selectedCredentialOptions.map {
+                    CredentialSelectionDto(queryId = it.queryId, credentialId = it.credentialId)
+                },
+                selectedDisclosureOptions = selectedDisclosureOptions.map {
```
szijpeter, line 192: **[P1] Preserve explicit empty disclosure selections**

When the user deselects every optional disclosure, `ifEmpty { null }` restores the backend’s default disclosure selection. This can share claims the user explicitly declined. Preserve `[]`; the backend distinguishes it from `null`. The focused adapter probe reproduces the lost empty selection.

waltid-applications/waltid-wallet-demo-compose/webApp/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+WORKDIR /src
+COPY . .
+
+RUN chmod +x ./gradlew \
+    && sed -i 's/-Xmx6[gG]/-Xmx3g/g' gradle.properties \
+    && ./gradlew --no-daemon \
+        :waltid-applications:waltid-wallet-demo-compose:webApp:wasmJsBrowserDistribution \
+        -PenableWalletDemoComposeWeb=true \
+    && DIST="$(find waltid-applications/waltid-wallet-demo-compose/webApp/build -type d -name productionExecutable | head -n 1)" \
```
szijpeter, line 17: **[P1] Copy the complete web distribution into the image**

Gradle creates several directories named `productionExecutable`, so selecting the first match can omit `composeResources`. The built image returned HTML for the logo request, causing an image-decoding exception and freezing the wallet UI. Copy `webApp/build/dist/wasmJs/productionExecutable` explicitly, including its resources.

docker-compose/wallet-api2/config/_features.conf
```diff
@@ -1,6 +1,6 @@
 enabledFeatures = [
     # dev-mode # Enabling developer config from file `dev-mode.conf`
-    # auth # Enable user authentication
+    auth # Enable user authentication (email/password JWT; see auth.conf)
```
szijpeter, line 3: **[P1] Pair enabled authentication with a compatible default image**

This enables authentication while the committed Compose default still selects `wallet-api2:stable`. The published application tested with this configuration failed because `ktor-authnz` was not registered. The startup fix in this PR does not reach that image automatically. Make the default select or build a compatible backend.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
```diff
@@ -0,0 +1,354 @@
...
+        issuanceSessions.remove(sessionId)
+        WalletApi2BrowserSessionStore.clearPendingIssuance()
+        return result.toOutcome()
+    }
+
+    override suspend fun continueAuthorizationIssuance(
+        sessionId: String,
+        callbackUri: String,
+    ): WalletDemoIssuanceOutcome {
+        val session = requireIssuance(sessionId)
```
szijpeter, line 130: **[P2] Validate the pending authorization callback before exchange**

`authorizationState` is persisted but never compared with the callback; only `code` is extracted. A callback containing `state=WRONG` still reaches `/receive/authorized`. Validate the callback against the pending session before exchange, following the existing issuance engine’s state and redirect validation.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
```diff
@@ -0,0 +1,354 @@
...
+
+    override suspend fun discardPresentationPreview(previewHandle: WalletDemoPresentationPreviewHandle) {
+        presentationSessions.remove(previewHandle.value)
+    }
+
+    override suspend fun deleteCredential(credentialId: String): Boolean =
+        client.deleteCredential(walletId, credentialId)
+
+    override suspend fun deleteWallet() {
+        runCatching { client.deleteWallet(walletId) }
```
szijpeter, line 233: **[P2] Stop reset when wallet deletion fails**

Swallowing the deletion exception lets reset create and select another wallet while the original wallet’s keys and credentials remain. The HTTP-500 probe reproduces this. Propagate the failure and keep the current wallet selected until deletion succeeds.

szijpeter (changes requested):
The browser issuance/presentation journey completed successfully, including verifier `SUCCESSFUL`, using these local workarounds:

- Built all three backend images from this PR’s exact head instead of using the published `stable` images.
- Mounted the missing `composeResources` into the Docker-built web frontend to prevent the image-decoding failure and frozen UI.
- Configured a local verifier certificate with `digitalSignature` key usage and explicitly pinned its test CA in the wallet. Signature and certificate validation remained enabled.

These results do not establish that the default Docker setup works unchanged. The certificate/trust configuration problems also exist in the base branch and are not counted as new PR regressions.

Please address the five inline findings covering disclosure consent, Docker packaging/defaults, authorization callback validation, and reset error handling.

docker-compose/wallet-api2/config/_features.conf
JakeFernandes98: Left unchanged for this follow-up. The new releases will have ktor-authnz available.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
JakeFernandes98: Fixed in `f6066ed77`. Reset now runs delete first via `replaceWalletAfterSuccessfulDelete` and only creates/selects a replacement after delete succeeds. If delete fails, the current wallet stays selected. Coverage is in `WalletApi2MappingsTest` and `WalletDemoControllerTest.resetWalletKeepsSessionWhenDeleteFails`.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
JakeFernandes98: Fixed in `f6066ed77`. The callback is parsed against the pending session before `/receive/authorized`: state is compared, and the redirect target (protocol/host/port/path) must match. A mismatch returns a failure without exchanging the code or clearing the pending session. `access_denied` cancels. Coverage is in `WalletApi2AuthorizationCallbackTest`.

waltid-applications/waltid-wallet-demo-compose/webApp/Dockerfile
JakeFernandes98: Fixed in `f6066ed77`. The Dockerfile now copies `webApp/build/dist/wasmJs/productionExecutable` explicitly (not the first `find` match), falls back to `sharedUI` processed `composeResources` if needed, and asserts `/out/composeResources` is present before the image is built.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2DemoWallet.kt
JakeFernandes98: Fixed in `f6066ed77`. Empty disclosure selections are now sent as `[]` via `toDisclosureSelectionDtos()` instead of `ifEmpty { null }`, so an explicit empty choice is no longer dropped and replaced by the backend default. Mapping/encoding coverage is in `WalletApi2MappingsTest`.

JakeFernandes98:
Pushed `f6066ed77` on `feature/wal-1198` (rebased onto the latest main merge).

Addressed four of the review findings: empty disclosures are sent as `[]`, the web image copies the Wasm dist with `composeResources`, authorization callbacks are validated before code exchange, and wallet reset no longer proceeds when delete fails.

The wallet-api2 default image pin was left unchanged: published `stable` still cannot load this auth config, so local compose continues to need a snapshot/`latest` backend with auth.

szijpeter (commented):
Rechecked at `f6066ed773`. The four fixes for disclosure selection, Docker resources, callback state validation, and reset failure propagation look addressed. Two callback issues remain, detailed inline; the error-only callback routing gap was uncovered during re-review and was not introduced by the latest fix.

Backend-image compatibility is acknowledged as explicitly deferred. Until a compatible release is selected, the workaround remains building and selecting the compatible backend from this PR. Pre-authorized issuance avoids the two callback issues where available.

In the re-review, all 17 existing targeted tests passed; two additional regression probes reproduced the callback issues, while the ordinary callback control passed. The production Wasm distribution built successfully. The live journey was not rerun.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2AuthorizationCallback.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if (expected.fragment.isNotBlank()) return false
+    if (expected.protocol != callback.protocol ||
+        !expected.host.equals(callback.host, ignoreCase = true) ||
+        expected.port != callback.port ||
+        expected.normalizedPath() != callback.normalizedPath()
+    ) {
+        return false
+    }
+    val callbackBaseNames = callback.parameters.names() - callbackParameterNames.toSet()
```
szijpeter: **[P2] Allow unrecognized authorization response parameters**

A valid callback containing `code`, matching `state`, and an additional parameter such as `session_state` is rejected as a redirect mismatch. A focused Wasm test reproduces this. [OAuth 2.0 section 4.1.2](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2) requires clients to ignore unrecognized response parameters. Preserve validation of the redirect target and its configured query parameters without rejecting additional response parameters.

waltid-applications/waltid-wallet-demo-compose/webApp/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/web/Main.kt
```diff
@@ -1,18 +1,126 @@
...
+            pinStore = InMemoryDemoPinStore(),
+            skipPin = true,
+            issuanceRedirectUri = redirectUri,
+            signingProtectionMode = WalletDemoSigningProtectionMode.Disabled,
         )
     }
+
+    LaunchedEffect(controller) {
+        val href = window.location.href
+        if (href.contains("code=")) {
```
szijpeter: **[P2] Route authorization error callbacks to the handler**

This guard only dispatches callbacks containing `code=`, so `?error=access_denied&state=…` never reaches the new cancellation handling and leaves issuance pending. `WalletDeepLinkScheme.parse` also requires `code`, so both entry points need to accept error responses and pass them through session validation. A focused classification test reproduces the failure.

waltid-applications/waltid-wallet-demo-compose/sharedLogic/src/commonMain/kotlin/id/walt/walletdemo/compose/logic/walletapi2/WalletApi2AuthorizationCallback.kt
JakeFernandes98: Fixed in `6ca5a0a6e`. Redirect matching now requires the configured query parameters and ignores additional authorization response parameters such as `session_state`. Coverage is in `WalletApi2AuthorizationCallbackTest.unrecognizedResponseParametersAreIgnored`.

waltid-applications/waltid-wallet-demo-compose/webApp/src/wasmJsMain/kotlin/id/walt/walletdemo/compose/web/Main.kt
JakeFernandes98: Fixed in `6ca5a0a6e`. HTTP(S) callbacks with `error=` are now classified as authorization callbacks, the web host dispatches them the same way as `code=`, and they still go through session validation. `access_denied` cancels pending issuance. Coverage is in `WalletDeepLinkSchemeWebTest` and `WalletDemoControllerTest.httpErrorAuthorizationCallbackIsDispatched`.

JakeFernandes98:
Pushed `6ca5a0a6e` on `feature/wal-1198`.

- Merged `main` and resolved the `WalletScreen` conflict (kept the web authorization opener and sign-out path; adopted the session-based credentials tab).
- Authorization callbacks now ignore unrecognized response parameters such as `session_state`.
- Error-only HTTP callbacks (`error=access_denied`) are routed through session validation and cancel pending issuance.

The Google x509 fixture expiry is left to the independent fix.

szijpeter (approved):
The latest changes address the review findings. Approved.

Non-blocking finding: denying issuer authorization in Compose web returns to the wallet but shows “Receive failed: received credentials are not available locally.” Cancellation should instead clear the pending issuance and return to Receive with a neutral message such as “Credential offer declined.”

Reproduced using a local test issuer: Receive → Continue to sign in → Deny authorization. The shared Compose controller treats cancellation as an empty issuance result and then checks for received credentials. This behavior predates this PR and does not block approval.

Workaround: dismiss the error and start another offer; cancellation clears the pending issuance and adds no credential.

JakeFernandes98:
Follow-up to the leftover cancellation note: `ef5f1f61b` now treats `WalletDemoIssuanceOutcome.Cancelled` in the shared Compose controller as a declined offer.

Denying issuer authorization (and other cancelled issuance outcomes) clears the pending receive flow and returns to Receive with “Credential offer declined”, instead of “Receive failed: received credentials are not available locally.”

Also merged current `main` so this branch includes the x509 fixture and demo issuer metadata-signing CI fixes.

szijpeter:
Found a Coil initialization-order crash in the combined Android DC API suite: the 10 sharing tests pass, then the first issuance test crashes because provider UI has already initialized Coil before the main wallet registers its image-loader factory. Prepared [#2226](https://github.com/walt-id/waltid-identity/pull/2226), targeting this branch, to register the factory at host startup. The same full suite passes locally with the fix: **13/13, no failures or skips**.

PR #2208, opened 2026-09-10
fix: address review findings on x509 CSR/certificate builder

philpotisk:
## Summary

Fixes the 4 findings from the automated review of #2152 (posted [here](https://github.com/walt-id/waltid-identity/pull/2152#issuecomment-5621125843)):

- `X509CertificateUtil.createCertificate(issuerKey: Crypt1Key, ...)` now resets `builder.issuerDnRaw` after `block.invoke(builder)`, matching the `Key`-based overload — prevents a caller's block from producing a certificate whose issuer DN doesn't match the actual issuing certificate.
- The new `require()` exclusivity checks for subject-public-key-info now have accurate error messages instead of copy-pasted ones describing the wrong condition.
- `BouncyPkcs10CertificateSigningRequestSigner` no longer conditionally omits the PKCS#9 extensionRequest attribute for CSRs with zero extensions — restores parity with `SignumCertificateSigner`, which always includes it.
- `IsoDocumentSignerX509CertificateProfile`'s crypto2 `Key`-based `profileDocumentSignerCertificate` overload now reuses the shared no-arg helper, removing the last duplicated copy of the validity/extension setup.

### DIFF
--- a/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/X509CertificateUtil.kt
+++ b/waltid-libraries/crypto/waltid-x509/src/commonMain/kotlin/id/walt/certificate/x509/X509CertificateUtil.kt
@@ -115,7 +115,7 @@ sealed class X509CertificateUtil(val services: X509CertificateServices) {
         require(subjectPublicKeyInfo.key != null || subjectPublicKeyInfo.spki != null) {
             "Certificate subject public key missing"
         }
-        require(subjectPublicKeyInfo.crypto1key == null) { "For subject public key info key or SPKI must be set" }
+        require(subjectPublicKeyInfo.crypto1key == null) { "Certificate subject public key must not mix Crypto1Key with key or SPKI" }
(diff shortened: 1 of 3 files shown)


philpotisk:
Update: found and fixed a real regression in my own change — removing the `isNotEmpty()` guard around the Bouncy CSR extensionRequest attribute broke every CSR path without extensions (BouncyCastle's `ExtensionsGenerator.generate()` throws on an empty extension set). Restored the guard with a comment explaining why it's required; the Bouncy/Signum divergence flagged in review turned out to be a necessary platform workaround, not a bug.

Verified this time: `:waltid-libraries:crypto:waltid-x509:jvmTest` passes, and — with `waltid-identity` on this branch — the full `waltid-identity-enterprise` x509 integration test suite (208/208, including `X509ServiceCsrAndGenericHappyPathIntegrationTest` and `X509CertificateServiceIntegrationTest::shouldCreateCertificateFromCsr`) passes too.

hokorn:
> I have also tried the same

hokorn (approved):

PR #2217, opened 2026-09-14
More convenient Java API for x509 certificate util (WAL-1402) and crypto2 key gneration

hokorn:
## Summary

Adds Java-friendly facades for two Community Stack crypto libraries — `waltid-x509`'s `X509CertificateUtil` and `waltid-crypto2-java`'s software-key generation — so Java (and other non-Kotlin JVM) callers can create/parse/validate certificates and CSRs and generate software keys without touching `suspend` functions or Kotlin's `Builder.() -> Unit` lambda-with-receiver DSLs (WAL-1402).

### `waltid-x509`: `JavaX509CertificateUtil`
- New `JavaX509CertificateUtil` class wraps `X509CertificateUtil` as plain blocking methods (`runBlocking` under the hood) configured with `java.util.function.Consumer` callbacks instead of Kotlin builder lambdas, and accepts `byte[]` in place of `ByteString` for DER input.
- Covers the full surface: parsing PEM/DER certificates and CSRs, issuing self-signed and CA-issued certificates, CSR/chain validation, and `restoreSubjectPublicKey`.
- `JavaX509CertificateUtil.getDefault()` gives a ready-to-use BouncyCastle-backed instance; `configure(...)` builds a custom instance (e.g. custom trust store or chain validators) from a base instance.

### `waltid-crypto2-java`: `JavaSoftwareKeys`
- New `JavaSoftwareKeys` object is the software-key counterpart to the existing `JavaManagedKeyProvider` SPI: generates software keys from a `CompletionStage`-returning API and adapts the resulting key to a plain-Java `JavaSoftwareKey` facade (`signer()`/`verifier()` as `CompletionStage`-based callbacks), or optionally returns the real `SoftwareKey` (`generateKey`) for Java callers that need to hand it into other crypto2-consuming Kotlin APIs such as `waltid-x509`'s certificate builders.
- New `JavaGenerateSoftwareKeyRequest` / `JavaSoftwareKey` Java types and their Kotlin conversion glue in `JavaSoftwareKeys.kt`.
- `JavaProviderAdapter`'s Kotlin↔Java type converters (`SignatureAlgorithm.toJava()`, `KeySpec.toJava()`, etc.) are widened from `private` to `internal` so `JavaSoftwareKeys.kt`, in the same module, can reuse them instead of duplicating the mapping.
- `JavaX509CertificateUtil.restoreSubjectPublicKey(certificate, cryptoRuntime)` added as the blocking Java entry point for `X509Certificate.restoreSubjectPublicKey`, completing the loop with keys generated via `JavaSoftwareKeys`.
- `.api` file updated for the new public surface.

## Architecture Notes

- Both facades follow the same shape already established by `JavaManagedKeyProvider`: a thin adapter layer under `jvmMain` that never changes the underlying Kotlin API, only wraps it — Kotlin call sites are unaffected and should keep using `X509CertificateUtil` / the Kotlin crypto2 APIs directly.
- `JavaSoftwareKeys` deliberately offers both `generate` (Java-only facade) and `generateKey` (real `SoftwareKey`) so Java callers can choose between staying fully in Java types or interoperating with Kotlin-side consumers like `waltid-x509`.

## Breaking

- None. These are additive facades; no existing public API changed shape (only two internal-visibility widenings in `JavaProviderAdapter.kt`, which are additive from a binary-compatibility standpoint).


waltid-libraries/crypto/waltid-x509/src/jvmMain/kotlin/id/walt/certificate/x509/JavaX509CertificateUtil.kt
```diff
@@ -0,0 +1,158 @@
...
+        issuerKey: Key,
+        signatureAlgorithm: SignatureAlgorithm,
+        block: Consumer<X509CertificateDataBuilder> = Consumer {}
+    ): X509Certificate = runBlocking {
+        delegate.createSelfSignedCertificate(issuerKey, signatureAlgorithm) { block.accept(this) }
+    }
+
+    @JvmOverloads
+    fun createSelfSignedCertificate(
+        issuerKey: Crypto1Key,
```
philpotisk, line 81: How would this work with key from waltid-crypto2?

hokorn: With the method above:

```
    @JvmOverloads
    fun createSelfSignedCertificate(
        issuerKey: Key,
        signatureAlgorithm: SignatureAlgorithm,
        block: Consumer<X509CertificateDataBuilder> = Consumer {}
    ): X509Certificate = runBlocking {
        delegate.createSelfSignedCertificate(issuerKey, signatureAlgorithm) { block.accept(this) }
    }
```

philpotisk (approved):

PR #2218, opened 2026-09-14
Update workflows with latest action versions and improved configs

dinkar-jain:


JakeFernandes98 (approved):

JakeFernandes98 (changes requested):
See comments

.github/actions/checkout-repos/action.yml
```diff
@@ -79,31 +79,31 @@ runs:
         resolve_ref "walt-id/waltid-identity-enterprise-license" true enterprise_license
 
     - name: Checkout waltid-unified-build
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
```
JakeFernandes98, line 82: WAL-1320's first requirement is that a PR editing an in-repo composite must run **this PR's** copy.

This `checkout@v6` bump is not what CI runs. Every caller still pins an older SHA that uses `checkout@v4`:

- `d123dc03` — `android-device-tests.yml:49`, `gradle.yml:72`, `crypto2-platforms.yml:15`, plus `deployment.yml:86` and `docker.yml:44` (untouched here)
- `6cf69001` — `gradle-macos.yml:74`, `mobile-sdk-docs.yml:50`

Both of those commits still use `checkout@v4`. `gradle-macos.yml:60` documents that only `checkout-repos` stays pinned — the pin was not then moved onto a commit that contains this v6 bump.

Every job on `d689d1e` that goes through `checkout-repos` still annotates:

> Node.js 20 is deprecated… `actions/checkout@v4`

`setup-java` / `upload-artifact` / `cache` warnings are gone; checkout is the leftover ticket item.

Minimal fix: retarget every `checkout-repos@…` pin, including `deployment.yml` and `docker.yml`, to a commit already on this branch that contains `checkout@v6` (this head `d689d1e773d6918cc42299b3cf9563a9ebce3ca0` works) and unify the two pin SHAs.

Ticket-aligned fix: bootstrap with `actions/checkout@v6` (sparse to the composite) then `uses: ./.github/actions/checkout-repos`, so the next composite edit cannot go stale again.

After the change, re-run and confirm the `checkout@v4` annotations disappear. That is the acceptance check for ticket item 1.

dinkar-jain: Updated SHA refs.

Didn't use `./.github/actions/checkout-repos` directly because the composite itself needs to be present in the runner workspace before it can be loaded. That would require an additional checkout step.

JakeFernandes98 (approved):

PR #2225, opened 2026-09-15
Add data mapping support for mDoc credentials

SuperBatata:
This pull request enhances the handling of mdoc credential data mapping and adds comprehensive testing to ensure that data mapping correctly overrides and processes namespace data before signing. The main improvements focus on ensuring that any provided data mapping is merged into the credential payload, with support for function evaluation and proper context, and that this logic is thoroughly tested.

**Enhancements to mdoc credential data mapping:**

* Introduced the `mergeMdocPayloadWithMapping` extension function on `JsonObject` to merge the provided `dataMapping` into the credential payload, evaluating functions and using a context containing `issuerId` and `display` information. This ensures mapped values (including function results) override the original namespace data before signing.
* Updated the credential issuance flow in `MdocCredentialHandler` to use the merged credential data (`mappedCredentialData`) instead of the original `credentialData` at all relevant stages, ensuring the mapping is consistently applied. [[1]](diffhunk://#diff-8ddf8afb36a66213beb2c87a1a7cb843cf6884290307a57f08ed08bac1da3665R78-R86) [[2]](diffhunk://#diff-8ddf8afb36a66213beb2c87a1a7cb843cf6884290307a57f08ed08bac1da3665L90-R99) [[3]](diffhunk://#diff-8ddf8afb36a66213beb2c87a1a7cb843cf6884290307a57f08ed08bac1da3665R131-R136) [[4]](diffhunk://#diff-8ddf8afb36a66213beb2c87a1a7cb843cf6884290307a57f08ed08bac1da3665L134-R145)

**Testing improvements:**

* Added a new test, `data mapping overrides mdoc namespace data and evaluates functions before signing`, to verify that the data mapping logic correctly overrides namespace values and evaluates functions, including substitution of dynamic values like `<issuerId>`.
* Refactored the test fixture to allow issuing credentials with custom `credentialData` and `dataMapping`, and to return the full `IssuerSigned` object for more detailed assertions. [[1]](diffhunk://#diff-0fda74720441a4e61d0a995d3293e7dca201c94e41b9011a3bfe0fc65d62bd53R117-R132) [[2]](diffhunk://#diff-0fda74720441a4e61d0a995d3293e7dca201c94e41b9011a3bfe0fc65d62bd53L99-R142) [[3]](diffhunk://#diff-0fda74720441a4e61d0a995d3293e7dca201c94e41b9011a3bfe0fc65d62bd53L112-R154)

**Dependency and import updates:**

* Added necessary imports for JSON processing and mapping utilities to support the new merging and function evaluation logic. [[1]](diffhunk://#diff-8ddf8afb36a66213beb2c87a1a7cb843cf6884290307a57f08ed08bac1da3665R22-R33) [[2]](diffhunk://#diff-0fda74720441a4e61d0a995d3293e7dca201c94e41b9011a3bfe0fc65d62bd53R26)


waltid-libraries/protocols/waltid-openid4vci/src/jvmTest/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialValidityTest.kt
```diff
@@ -69,6 +71,45 @@ class MdocCredentialValidityTest {
...
+                    put("expiry_date", "2024-10-20")
+                    put("issuing_authority", "request-authority")
+                    putJsonArray("administrative_number") {
+                        add("request-value")
+                    }
+                }
+            },
+            dataMapping = buildJsonObject {
+                putJsonObject(namespace) {
+                    put("issue_date", "2026-09-15")
```
JakeFernandes98, line 91: **Major:** This test hard-codes `issue_date` / `expiry_date` as static strings. WAL-1392 is about mapping *functions* dynamically calculating those dates, which is how SD-JWT examples work (`<timestamp>`, `<timestamp-in:365d>`). The only function asserted here is `<issuerId>`.

Please add a test that maps `issue_date` / `expiry_date` with the documented date functions, then signs with the standard mDL `mDocNameSpacesDataMappingConfig` (`stringToFullDate` on those elements).

JakeFernandes98 (commented):
## Review notes (major + minor)

Issuer2 / `waltid-openid4vci` applying `dataMapping` before mDoc signing is the right direction for WAL-1392. The replace-merge in the second commit is also the right array semantics.

The remaining gaps are inline. One cannot be anchored on this diff:

**Major (v1 issuer):** `IssuerEnterpriseService.doGenerateMDoc` still builds namespaces from `credentialData` only and never applies `issuanceRequest.mapping`. If WAL-1392 traffic is on the v1 issuer, this PR does not close the ticket and needs a counterpart there (or an explicit Issuer2-only resolution).

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
```diff
@@ -152,6 +166,71 @@ class MdocCredentialHandler(
...
+    }
+
+    private suspend fun mergeMdocJsonElement(
+        original: JsonElement?,
+        mapping: JsonElement,
+        context: Map<String, JsonElement>,
+        functionHistory: MutableMap<String, JsonElement>,
+    ): JsonElement = when (mapping) {
+        is JsonPrimitive -> when {
+            mapping.isString && mapping.isTemplate() -> getTemplateData(
```
JakeFernandes98, line 226: **Major:** After this merge, `"issue_date": "<timestamp>"` evaluates to an Instant string such as `2026-09-15T10:06:00.123Z`. mDL issuance then runs `LocalDate.parse(s)` via `stringToFullDate`, which rejects that form.

SD-JWT can keep Instant strings; ISO 18013-5 `issue_date` / `expiry_date` are `full-date`. Either add an mDoc-safe function (or truncate Instant → `YYYY-MM-DD` before CBOR conversion), or document that mDoc mappings must already be full-date strings. Otherwise the support repro moves from “mapping ignored” to “issuance fails in `stringToFullDate`”.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
```diff
@@ -67,14 +78,15 @@ class MdocCredentialHandler(
                     )
                 )
             }
```
JakeFernandes98, line 80: **Major:** `mergeMdocPayloadWithMapping` runs once before `computeCredentialResult` loops `verifiedProofs`. The Crypto2 overload has the same pre-loop merge.

SD-JWT evaluates mapping inside `SdJwtVcCredentialSigner.generateSdJwtVC`, so each credential gets a fresh `<uuid>`, `<timestamp>`, or `<webhook>`. Move the merge into the `issue` callback on both overloads so mDoc matches that contract.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
```diff
@@ -152,6 +166,71 @@ class MdocCredentialHandler(
         CredentialResponseResult.Failure(e.toCredentialHandlerError())
     }
 
+    private suspend fun JsonObject.mergeMdocPayloadWithMapping(
+        dataMapping: JsonObject?,
+        issuerId: String,
+        display: List<CredentialDisplay>?,
+    ): JsonObject {
+        val mapping = dataMapping ?: return this
```
JakeFernandes98: **Minor:** Context includes `issuerId`, `issuerDid`, and `display`, but not `subjectDid`. SD-JWT puts the holder DID in context. A mapping of `<subjectDid>` will throw `Cannot find in context: subjectDid`.

Add the holder DID from the verified proof when evaluating inside the per-proof `issue` path, or document that mDoc mappings cannot use `<subjectDid>`.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
```diff
@@ -152,6 +166,71 @@ class MdocCredentialHandler(
         CredentialResponseResult.Failure(e.toCredentialHandlerError())
     }
 
+    private suspend fun JsonObject.mergeMdocPayloadWithMapping(
```
JakeFernandes98, line 167: **Minor:** This merger is a private copy of `CredentialDataMergeUtils` with different array semantics. Prefer `CredentialDataMergeUtils.mergeMdocPayloadWithMapping` (or a shared replace-merge) so Issuer2, tests, and any later v1 fix share one implementation.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
JakeFernandes98, line 226: **Minor:** `JsonPrimitive.isTemplate()` calls `content.first()` before the `length > 2` check. An empty-string mapping value therefore throws instead of being copied through. Pre-existing, now reachable on this path. Guard `isNotEmpty()` / `length > 2` first.

waltid-libraries/protocols/waltid-openid4vci/src/commonMain/kotlin/id/walt/openid4vci/handlers/credential/MdocCredentialHandler.kt
```diff
@@ -152,6 +166,71 @@ class MdocCredentialHandler(
...
+        display: List<CredentialDisplay>?,
+    ): JsonObject {
+        val mapping = dataMapping ?: return this
+        val context = mapOf(
+            "issuerId" to issuerId,
+            "issuerDid" to issuerId,
+            "display" to Json.encodeToJsonElement(display ?: emptyList()).jsonArray,
+        ).filterValues {
+            when (it) {
+                is JsonElement -> it !is JsonNull && (it !is JsonObject || it.isNotEmpty()) && (it !is JsonArray || it.isNotEmpty())
```
JakeFernandes98, line 185: **Minor:** `display` is encoded from `CredentialDisplay`, whose logo/background fields are `uri`. The shared `<display>` function reads `logo["url"]` and `background_image["url"]`. Same encoding as SD-JWT, so this is not new, but any mDoc mapping that uses `<display>` with a logo will throw. Align names in `dataFunctions` or normalize the context object.

PR #2227, opened 2026-09-15
feat(wallet-sdk): configure transaction-data profiles locally

JakeFernandes98:
## Summary

This adds an opt-in Wallet SDK catalog of OpenID4VP `transaction_data` types so Compose Android, Compose iOS, and native iOS wallets can declare what they accept at bootstrap, instead of fetching profiles from Wallet API at startup.

Tracked in [WAL-1408](https://linear.app/walt-new/issue/WAL-1408/configure-mobile-transaction-data-profiles-locally-in-the-wallet-sdk). Native wallets were treating a remote Wallet API document as their allowlist, which raced DC API / Credential Manager entry points against the network, failed open to an empty list, and had already drifted across platforms.

### Wallet SDK
- Adds `MobileWalletTransactionDataProfiles` (KMP) and `WalletTransactionDataProfile` well-known values (Swift) covering payment authorization, account access, EUDI SCA payment, and `payment_card`.
- Keeps `transactionDataProfiles` empty by default. Integrators must pass `all` or a subset.
- Documents the catalog in the KMP README and Swift Getting Started page.

### Demo apps
- Compose Android, Compose iOS, and native iOS pass the catalog into SDK config at wallet create, including independently launched Digital Credential / document-provider entry points.
- Removes the HTTP GET of `/wallet-api/transaction-data-profiles`, the native iOS blocking semaphore fetch, fetch-failure warnings, and `TRANSACTION_DATA_PROFILES_URL` / `-PtransactionDataProfiles.url` plumbing.

### Tests and CI
- Device/integration tests and E2E fixtures use the local catalog instead of fetching profile fields from Wallet API.
- Android CI scripts no longer pass the remote profile URL.

## Architecture Notes

Accepted `transaction_data` types are a local capability / policy allowlist, not a server discovery document. Wallet API and Verifier2 `GET /transaction-data-profiles` remain the right surface for **server** wallets and verifier UIs; this change does not move HTTP discovery into the SDK.

## Caveats and Follow-Ups

- On-device presentation with `transaction_data` is not re-run here. Host tests cover the empty default and catalog contents; live verifier E2E still depends on CI devices.
- Web wallet / web portal profile fetching and Enterprise profile CRUD are unchanged.

## Breaking

- Demo apps no longer read `TRANSACTION_DATA_PROFILES_URL` or `-PtransactionDataProfiles.url`. Passing those values has no effect.
- SDK default remains an empty allowlist. Apps that previously relied on a successful public-demo fetch must now pass `MobileWalletTransactionDataProfiles.all` / `WalletTransactionDataProfile.all` (the walt.id demos do this).


szijpeter (commented):
The move to local configuration has clear benefits. I think it is worth weighing those against what the endpoint approach provided, particularly around consistency and configuration ownership.

### Current `main`: fetch profiles from Wallet API

**Pros**

- Gives the demo apps a centrally maintained catalog.
- Allows configuration changes without rebuilding the apps.
- Lets demos target a particular deployment through the configurable endpoint.
- Keeps deployment/demo-specific definitions outside the SDK.

**Cons**

- Makes transaction-data availability depend on a network request during wallet creation.
- The current native iOS implementation blocks initialization while fetching.
- The backend catalog can change independently of the installed wallet’s capabilities. Fetching a type identifier does not itself establish that the wallet understands that type.
- Tests still carry separate definitions, so this approach did not eliminate all drift.

### This PR: bundle profiles locally

**Pros**

- Makes the configured allowlist predictable and available without the startup fetch.
- Removes the blocking iOS path and its associated failure handling.
- Keeps the SDK empty by default, with apps explicitly selecting profiles.
- Provides reusable named profiles and corrects the existing iOS test field mismatch.

**Cons**

- Bundled definitions can become stale relative to a deployment; changing them requires rebuilding the app.
- Removing the endpoint override reduces runtime configurability for demo deployments.
- There are separate definitions in KMP, Swift, and both fixture libraries.
- Making the catalog public SDK API creates a maintenance commitment for a collection mixing demo, EUDI, and interoperability types.

### Failure behavior

My reading is that **both approaches fail closed for unrecognized transaction-data types**.

On `main`, a failed profile fetch installs an empty allowlist, so transaction-data requests are rejected while ordinary wallet flows remain available. In this PR, the demos use a bundled allowlist, so endpoint availability no longer affects which types they accept. Types outside that list remain rejected, and the SDK default remains empty.

Could we adjust the “failed open” wording accordingly? This improves availability and predictability rather than fixing a fail-open fallback. Here, “fail closed” refers specifically to type allowlisting, rather than complete validation of each profile’s payload.

### Catalog ownership and consistency

Could we separate **where definitions are maintained** from **when the wallet obtains them**?

A single authoritative definition, shared or generated into consumers at build time, could preserve consistency while avoiding the startup dependency. Separate bundled definitions could also work if contract tests enforce their agreement.

For SDK ownership, I can see either option making sense:

- If these are intentionally maintained, supported product profiles, an explicitly opt-in SDK catalog seems reasonable.
- If they primarily describe the public demos, demo-support code seems a better fit, using the existing SDK configuration hook.

The generic profile model and configuration hook clearly belong in the SDK. Could we clarify the intended ownership of this concrete catalog before establishing it as public API? I realize the ticket explicitly asks for an SDK catalog, so this is a design consideration rather than an implementation deviation.

### Tests and scope

The new catalog tests mostly compare `.all` against the same constants used to construct it. Could we strengthen them to cover every identifier and field set across the platform catalogs and fixtures? The fixture checks against their own local required-field definitions might also be simplified in favor of those contract tests.

Two scope clarifications may help:

- The mobile/web distinction is really about where the wallet executes. A browser wallet that validates and signs locally would benefit from the same ownership principle; a server-wallet UI can appropriately fetch metadata from its backend.
- The shared iOS document-provider configuration still uses an empty profile list. Its current ISO Annex C path does not establish an OpenID4VP regression, but the claim that these entry points receive the catalog could be narrowed.

Overall, I support removing the mandatory startup fetch. My main considerations are preserving a clear source of truth, preventing future platform drift, and deciding whether this catalog represents SDK-supported profiles or demo configuration.

PR #2236, opened 2026-09-17
Fix the verifier OutOfMemoryError: bounded session store and base64 byte encoding

waltkb:
## 1. More efficient storing of large byte strings

`MdocsSchemaMappingFunction` emitted `BYTES` as a JSON array with an entry per byte, costing a `JsonLiteral` + `String` + backing array ~on the order of 80 bytes of heap per byte of payload. A 230.000 Byte portrait became ~460,000 objects and tens of megabytes of tree.

Byte strings are now a single base64url string. (base64url because that is what OpenID4VP and ISO 18013-5 use elsewhere).

The array form is still accepted on read, so persisted credential data and existing clients keep working. This is a change to the `credentialData` shape that API consumers see.

Risk here: an mdoc's digests are computed over the CBOR, so a byte string returning as a CBOR text string would still parse, still look right in JSON, and fail signature verification only for credentials with binary fields.
The round trip is asserted on both the CBOR type and the bytes and mutation-verified (making the reverse mapping produce `CborString` fails the tests).

Deliberately not changed: the schema-less v1 path (`JsonElementConversions`), where `toDataElement()` maps any string to `StringElement`, so emitting base64 there would silently produce text strings. That path needs an explicit conversion hint (`StringToCborElementConverter` already exists for this).

## 2. Automatic cleanup of stored entries

Retention gives a used session `DEFAULT_RETENTION_YEARS` (ten) years, and the only eviction was the lazy one in `get` for the single id being looked up.

Now expired sessions are swept on an amortised schedule, and an access-ordered ceiling (`DEFAULT_MAX_IN_MEMORY_SESSIONS` = 2,000) drops the least recently used session with a warning, since an in-memory store cannot honour a ten-year retention promise.

Mutation-verified (removing the ceiling fails tests, removing the sweep too).


dinkar-jain (approved):

szijpeter:
The expiry sweep and CBOR byte round-trip tests look useful. A few suggestions and considerations from the code:

### Verifier parsing path

[`CredentialParser.handleMdocs`](https://github.com/walt-id/waltid-identity/blob/c954e382ba4dfb3b10a382b92b84c571f0d75b1c/waltid-libraries/credentials/waltid-digital-credentials/src/commonMain/kotlin/id/walt/credentials/CredentialParser.kt#L202) still calls `toSerializedJsonElement`, whose [`CborByteString` branch](https://github.com/walt-id/waltid-identity/blob/c954e382ba4dfb3b10a382b92b84c571f0d75b1c/waltid-libraries/crypto/waltid-crypto/src/commonMain/kotlin/id/walt/crypto/utils/JsonUtils.kt#L32) returns a numeric JSON array. The repository also still [serializes and deserializes the entire session](https://github.com/walt-id/waltid-identity/blob/c954e382ba4dfb3b10a382b92b84c571f0d75b1c/waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/VerificationSessionRepository.kt#L221).

This leaves the parser/storage path unchanged by the `schemafulToJsonElement` update. A portrait-bearing test through that path would help establish whether the allocation issue is addressed end to end. If the returned JSON shape changes, coordinating consumers that currently expect arrays would also be useful.

### Capacity eviction

The [access-ordered eviction](https://github.com/walt-id/waltid-identity/blob/c954e382ba4dfb3b10a382b92b84c571f0d75b1c/waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/VerificationSessionRepository.kt#L116) bounds the entry count, but it does not distinguish completed sessions from sessions waiting for the wallet or undergoing verification. Under capacity pressure, an unfinished session can therefore be removed.

One option would be to prefer eligible completed records and protect processing claims, with explicit overload behavior when no suitable eviction candidate remains. A test covering this situation would help clarify the intended behavior.

### Smaller validation suggestions

- In [`decodeSchemaBytes`](https://github.com/walt-id/waltid-identity/blob/c954e382ba4dfb3b10a382b92b84c571f0d75b1c/waltid-libraries/credentials/waltid-mdoc-credentials2/src/commonMain/kotlin/id/walt/mdoc/schema/MdocsSchemaMappingFunction.kt#L32), checking `JsonPrimitive.isString` would keep the new branch limited to base64 strings; `.content` also accepts booleans and numbers.
- A `require(maxSessions > 0)` guard would catch invalid configuration early. Currently, a nonpositive capacity allows creation to return a snapshot even though the new entry is immediately removed.

szijpeter:
One non-blocking consideration: replacing the serialization copy with `copy()` means snapshots now share nested objects with storage. Top-level assignments remain isolated, but changing a mutable nested object can change stored state without a repository update or version increment.

Some nested values are still mutable—for example, `CredentialPolicyResult.result`, `PolicyRunResult.results`, and `redirects.successRedirectUri`. That makes the comment’s claim that everything reachable is immutable or a top-level `var` too broad.

There is also an existing caller affected by this: the [PII purge route](https://github.com/walt-id/waltid-identity/blob/b6e79766ef209293588238816740702d44d4d534/waltid-services/waltid-verifier-api2/src/main/kotlin/id/walt/verifier2/OSSVerifier2Service.kt#L187) calls [`deletePII()`](https://github.com/walt-id/waltid-identity/blob/b6e79766ef209293588238816740702d44d4d534/waltid-libraries/protocols/waltid-openid4vp-verifier/src/commonMain/kotlin/id/walt/verifier2/data/Verification2Session.kt#L183), which modifies nested policy results before `compareAndSet`. If a concurrent update causes that check to reject the purge, those nested changes have already reached storage.

A focused improvement could be to have the purge construct replacement result objects while continuing to share immutable payloads. More generally, it would help to clarify whether snapshots guarantee isolation or whether callers must avoid all nested mutation; the shallow copy currently relies on the latter convention.

PR #2246, opened 2026-09-18
fix: emit OpenID4VCI proof iss for client-bound grants

JakeFernandes98:
## Summary

This adds the OpenID4VCI 1.0 JWT proof `iss` claim so Wallet2 can pass WeBuild ITB CS-01 authorization-code issuance (VCI-001/002/003/005, and the same class of failure on VCI-004). After the authorization-server redirect, the ITB issuer rejected our credential request with `invalid_proof` because the proof JWT omitted `iss`.

Wallet2 now writes `iss` as the session OAuth `client_id` for authorization-code and other client-bound token requests, and still omits it for anonymous pre-authorized access. Issuer2 now requires that match on client-bound grants instead of treating `iss` as optional when present.

Tracks [WAL-495](https://linear.app/walt-new/issue/WAL-495/webuild-test-bed-compliance). Enterprise counterpart: [walt-id/waltid-identity-enterprise#686](https://github.com/walt-id/waltid-identity-enterprise/pull/686).

### Proof builder
- `ProofOfPossessionBuilder` / `Crypto2ProofOfPossessionBuilder` accept an optional `clientId`.
- `JwtProofBuilder` writes `iss` only when `clientId` is set, and rejects a blank value.
- 4-arg `JwtProofBuilder` overloads remain so existing callers keep compiling and still omit `iss`.

### Wallet2 issuance
- Authorization-code proofs always send the request `client_id` as `iss`.
- Pre-authorized proofs send `client_id` unless the token request used anonymous access.
- `WalletIssuanceSessionService` records that anonymous decision at token exchange so proof building does not re-run attestation.
- `SignProofRequest` can optionally supply `clientId` for the same claim.

### Issuer2 verifier
- Anonymous pre-authorized access still forbids `iss`.
- Client-bound grants now require `iss` and require it to equal the access-token `client_id`.

## Architecture Notes

- The wallet uses the token-request anonymity already computed for the grant, not a hardcoded ITB client id (`eudiw-abca` remains the Wallet2 default).
- Session persistence is unchanged: token exchange and proof build happen in the same `complete()` call, so the anonymous flag does not need to be stored.

## Caveats and Follow-Ups

- This does not add an ITB/GITB runner or close WAL-495. Re-run CS-01 (and VCI-004) on the Aegean portal after deploy.
- Enterprise `LicenseOID4VCIClient` still compiles against the 4-arg builder overload and continues to omit `iss` (anonymous).

## Breaking

- None for existing 4-arg `JwtProofBuilder` callers. New `clientId` on the proof-builder interfaces is optional.
- `SignProofRequest` gains an optional `clientId`; JSON without the field still deserializes.
- Issuer2 now rejects client-bound credential proofs that omit `iss` or send a mismatched value. Wallets that previously omitted `iss` on authorization-code or non-anonymous pre-authorized grants will receive `invalid_proof`.


waltid-libraries/protocols/waltid-openid4vci-wallet/src/commonMain/kotlin/id/waltid/openid4vci/wallet/proof/ProofOfPossessionBuilder.kt
```diff
@@ -49,6 +51,7 @@ interface ProofOfPossessionBuilder {
         audience: String,
         nonce: String?,
         binding: ProofKeyBinding,
+        clientId: String? = null,
```
szijpeter: Non-blocking compatibility observation: adding a defaulted parameter replaces the existing abstract method signature in both proof-builder interfaces. An existing custom implementation compiles against the base but fails here with `overrides nothing`; a previously compiled caller through either interface throws `NoSuchMethodError`. The overloads retained on `JwtProofBuilder` protect concrete calls, but not these interface contracts. The same binary issue affects the old two-argument `ProofBuilderUtils.validateProofParameters` call. These APIs are present in v1.0.0. Please assess whether to preserve the previous signatures alongside the new entry points or explicitly document that custom implementations must be updated and consumers recompiled.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
```diff
@@ -357,11 +357,17 @@ data class SignProofRequest(
     val key: DirectSerializedKey? = null,
     val keyId: String? = null,
     val did: String? = null,
+    /**
+     * OAuth `client_id` written as the proof `iss` claim for client-bound token requests.
+     * Leave unset for anonymous pre-authorized access (OpenID4VCI 1.0 Appendix F.1).
+     */
+    val clientId: String? = null,
```
szijpeter, line 364: Non-blocking compatibility observation: the optional JSON field remains compatible, but adding it to this published data class changes its JVM constructor and generated `copy` signatures. I reproduced `NoSuchMethodError` when running consumers compiled against the base that construct `SignProofRequest` with default arguments or call `copy(nonce = ...)` against this head. This model is already present in v1.0.0. Please assess whether the release should preserve the previous binary entry points or explicitly document the recompilation requirement; the current Breaking section only describes JSON compatibility.

szijpeter (approved):
Approved with two non-blocking compatibility observations below; please assess whether preserving the published APIs or documenting the required consumer migration is appropriate for this release.

waltid-libraries/protocols/waltid-openid4vci-wallet/src/commonMain/kotlin/id/waltid/openid4vci/wallet/proof/ProofOfPossessionBuilder.kt
JakeFernandes98: Good catch — addressed in ba0242f8a7.

I reverted the `ProofOfPossessionBuilder` / `Crypto2ProofOfPossessionBuilder` abstract methods back to their v1.0.0 arities (no `clientId`) and restored the original two-arg `ProofBuilderUtils.validateProofParameters`. The `iss`/`clientId` support is now exposed only as additive concrete overloads on `JwtProofBuilder` (plus a separate three-arg `validateProofParameters`), so the interface contracts and the two-arg util keep their original JVM descriptors.

Every internal call site uses the concrete `JwtProofBuilder`, so no functionality is lost while existing external interface implementations and precompiled callers stay binary compatible.

waltid-libraries/protocols/waltid-openid4vc-wallet/src/commonMain/kotlin/id/walt/wallet2/handlers/WalletIssuanceHandler.kt
JakeFernandes98: Agreed, and this one cannot be made binary compatible. `SignProofRequest` is a `@Serializable data class`, and the new field has to live in the primary constructor for kotlinx.serialization to (de)serialize it, so the generated constructor / `copy` / synthetic `$default` descriptors change unavoidably while keeping the feature.

I have documented the recompilation requirement in the PR Breaking section: consumers that construct `SignProofRequest` (including with default arguments) or call `copy(...)` against a v1.0.0 binary must recompile against this version, otherwise they hit `NoSuchMethodError`. Happy to consider an alternative mitigation if you would prefer one.

chsavvaidis (approved):

PR #2248, opened 2026-09-18
feat(credentials): shared KMP status-list codecs (WAL-593)

JakeFernandes98:
## Summary

This adds a Kotlin Multiplatform `waltid-status-list` library so encode and decode for IETF Token Status List, W3C BitstringStatusList, StatusList2021, and RevocationList2020 live in one place. `waltid-verification-policies2` now decodes through those codecs instead of keeping a second expansion implementation.

Ticket: [WAL-593](https://linear.app/walt-new/issue/WAL-593/credential-status-refactor)

Related PRs:
- [walt-id/waltid-identity#2248](https://github.com/walt-id/waltid-identity/pull/2248)
- [walt-id/waltid-identity-enterprise#687](https://github.com/walt-id/waltid-identity-enterprise/pull/687)
- [walt-id/waltid-unified-build#64](https://github.com/walt-id/waltid-unified-build/pull/64)

### Shared status-list library
- New KMP module with bit packing/reading, gzip/zlib expect-actual compression, string encodings, and four codecs.
- JVM compression uses `java.util.zip` so IETF Token Status List Appendix C vectors stay byte-identical.
- JS and Apple use korlibs for the same codec API.

### Policies2
- Expansion algorithms decode via the shared codecs.
- Constructors still accept the unused `Base64UrlHandler` argument for compatibility.
- `waltid-verification-policies` (v1) is unchanged on purpose.

## Architecture Notes

- Encode and decode both live in OSS so enterprise can drop its private bitstring encoder and still share decode with policies2.
- Storage, signing, Mongo, and publishing stay in the enterprise counterpart.

## Caveats and Follow-Ups

- This PR does not cut a Maven/npm release; publishing follows the existing library pipeline.
- Policies v1 still has its own expansion path.

## Breaking

- None for published OSS APIs. Expansion algorithm constructors still accept the unused handler argument.


waltid-libraries/credentials/waltid-status-list/src/jvmMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.jvm.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    actual fun decompress(data: ByteArray): ByteArray {
+        val inflater = Inflater()
+        inflater.setInput(data)
+        val result = ByteArrayOutputStream()
+        val buffer = ByteArray(1024)
+        while (!inflater.finished()) {
+            val count = inflater.inflate(buffer)
+            if (count == 0 && inflater.needsInput()) break
```
szijpeter: **[P1] Reject incomplete or dictionary-dependent zlib streams**

Please handle `needsDictionary()` and require the stream to finish successfully. `TokenStatusListCodec.decode("eLsADQAHYwQjAAAsAA0")` loops indefinitely here: `inflate()` returns zero, but neither `finished()` nor `needsInput()` is true. Removing the checksum trailer from a valid stream also returns decoded bytes instead of failing. The previous decoder rejects both inputs, and policies2 now reaches this implementation. A bounded negative test for each case would help; please also release the inflater in `finally`.

waltid-libraries/credentials/waltid-status-list/src/commonMain/kotlin/id/walt/statuslist/bit/BitReader.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun readStatus(
+        input: ByteArray,
+        index: Int,
+        bitSize: Int,
+        bitOrder: BitOrder,
+        reverseBits: Boolean = false,
+    ): UInt {
+        val bits = readBits(input, index, bitSize, bitOrder)
+        val ordered = if (reverseBits) bits.reversed() else bits
+        if (ordered.isEmpty()) return 0u
```
szijpeter: **[P2] Reject indexes outside the decoded list**

`TokenStatusListCodec.readStatus(byteArrayOf(0), index = 8, bits = 1)` returns `0u` (VALID), although that byte only contains indexes 0–7. Please validate the full requested bit range and reject missing bits instead of returning a status. This concerns the new public codec API; policies2 still uses its existing reader. A test at the first index beyond the list would catch it.

waltid-libraries/credentials/waltid-status-list/src/commonMain/kotlin/id/walt/statuslist/bit/BitReader.kt
```diff
@@ -0,0 +1,37 @@
...
+object BitReader {
+    fun readBits(
+        input: ByteArray,
+        index: Int,
+        bitSize: Int,
+        bitOrder: BitOrder,
+    ): List<Boolean> {
+        val bitStartPosition = index.toULong() * bitSize.toUInt()
+        val byteStart = (bitStartPosition / 8u).toInt()
+        val bytesToRead = (bitSize - 1) / 8 + 1
```
szijpeter: **[P2] Account for the starting offset when reading across bytes**

A three-bit W3C entry at index 2 starts at bit 6 and needs two bytes, but this calculation reads only one. With decoded bytes `[0x01, 0xc0]`, `BitstringStatusListCodec.readStatus(decoded, 2, 3)` returns `1u` instead of `3u`. Please include the starting bit offset in the byte-span calculation and add a cross-byte test; `take(bitSize)` currently hides the short read.

Correction: the expected value in this fixture is `3u`, not `7u` as I originally wrote. Thanks for catching that! Verified fixed in `f952daf28d`.

szijpeter (commented):
Nice refactor! I reviewed this together with [Enterprise #687](https://github.com/walt-id/waltid-identity-enterprise/pull/687) and [Unified Build #64](https://github.com/walt-id/waltid-unified-build/pull/64). Four reproducible codec edge cases below need addressing; I found no additional issues in the companion diffs.

waltid-libraries/credentials/waltid-status-list/src/korlibsMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.korlibs.kt
```diff
@@ -0,0 +1,10 @@
+package id.walt.statuslist.compression
+
+import korlibs.io.compression.deflate.ZLib
+import korlibs.io.compression.compress
+import korlibs.io.compression.uncompress
+
+actual object ZlibCompressor {
+    actual fun compress(data: ByteArray): ByteArray = ZLib.compress(data)
```
szijpeter: **[P2] Handle empty input in the JavaScript compressors**

Both new `CompressionRoundTripTest` cases fail under `jsNodeTest` for an empty byte array. The JS gzip and zlib encoders emit the wrapper without a valid deflate block, so their output cannot be decompressed; nonempty controls pass. Please handle this case in both compressors while retaining the common empty-input tests. JVM and iOS passed these tests in the review.

waltid-libraries/credentials/waltid-status-list/src/jvmMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.jvm.kt
JakeFernandes98: Fixed on `f952daf28d`. JVM inflate now treats `needsDictionary()`, zero-output (not finished), and missing Adler-32 as errors, and the inflater is always `end()`ed in `finally`. Added `ZlibCompressorNegativeTest` for `eLsADQAHYwQjAAAsAA0` and a truncated trailer.

waltid-libraries/credentials/waltid-status-list/src/commonMain/kotlin/id/walt/statuslist/bit/BitReader.kt
JakeFernandes98: Fixed. `readBits` now requires the full bit range to fall inside the decoded list and throws `IndexOutOfBoundsException` instead of returning `0u`. Covered with `TokenStatusListCodec.readStatus(byteArrayOf(0), index = 8, bits = 1)`.

waltid-libraries/credentials/waltid-status-list/src/commonMain/kotlin/id/walt/statuslist/bit/BitReader.kt
JakeFernandes98: Fixed: byte span is `(bitOffset + bitSize + 7) / 8`, so a 3-bit value at index 2 reads two bytes.

Packing value `7` at that slot round-trips to `7u`. For the `[0x01, 0xC0]` fixture specifically, MSB-first bits 6–8 are `011` (`3u`); the old one-byte read returned `1u` because it only saw `01`. Happy to adjust if you intended a different bit numbering.

waltid-libraries/credentials/waltid-status-list/src/korlibsMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.korlibs.kt
JakeFernandes98: Fixed. korlibs JS still emits an invalid stored block for empty input even with a larger `outputSizeHint`, so empty gzip/zlib now use RFC stored-block payloads. `CompressionRoundTripTest` empty cases pass under `jsNodeTest`.

waltid-libraries/credentials/waltid-status-list/src/korlibsMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.korlibs.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+private val EMPTY_ZLIB = byteArrayOf(
+    0x78, 0x01,
+    0x01, 0x00, 0x00, 0xff.toByte(), 0xff.toByte(),
+    0x00, 0x00, 0x00, 0x01,
+)
+
+actual object ZlibCompressor {
+    actual fun compress(data: ByteArray): ByteArray =
+        if (data.isEmpty()) EMPTY_ZLIB
```
szijpeter: **[P2] Return a fresh array for each empty compression**

Could you return `EMPTY_ZLIB.copyOf()` here and do the same for `EMPTY_GZIP`? Both are mutable singletons: `ZlibCompressor.compress(byteArrayOf()).fill(0)` corrupts every later empty compression, including `TokenStatusListCodec.encode(byteArrayOf())`. I reproduced the same failure for gzip on JS; Apple shares this implementation. A test that mutates one result and verifies the next round trip would cover it.

szijpeter (commented):
Thanks for the fixes! All four original findings are addressed; one small shared-buffer issue remains in the new empty-input handling.

waltid-libraries/credentials/waltid-status-list/src/korlibsMain/kotlin/id/walt/statuslist/compression/ZlibCompressor.korlibs.kt
JakeFernandes98: Fixed on `1b4a615f90`. Empty gzip/zlib now return `copyOf()` of the stored-block bytes. Added common tests that mutate one empty compress result and still round-trip the next; `jvmTest` and `jsNodeTest` both passed.

szijpeter (commented):
Thanks, the fresh-array fix and regression tests address the last finding; I found no further issues in the update. All 61 JVM/JS/iOS simulator codec tests pass locally; CI is still pending.

szijpeter (approved):

PR #2251, opened 2026-09-18
ci: fail hung Linux Gradle jobs instead of running for six hours

JakeFernandes98:
## Summary
- Linux Gradle jobs were hanging after ~45 minutes of real work and sitting until the 6-hour job timeout. Cap that job at 90 minutes, add 10-minute per-test / 25-minute per-task timeouts, and set `KLOGGING_MIN_LOG_LEVEL=INFO` so TRACE test logs stop filling the 76MB GitHub log cap.
- Restore PR Gradle to `build allTests` with the setup-gradle daemon. Keep `clean cleanAllTests --rerun-tasks --no-daemon` for main / `ci:cacheless` only.
- This does not fix the suspected hung verifier worker from #2236; it makes that hang fail in minutes instead of occupying runners for hours.

## Test plan
- [ ] Confirm this PR’s `gradle-build / gradle (ubuntu-22.04)` either finishes or fails within 90 minutes (not 6 hours)
- [ ] Confirm a passing run still reaches `BUILD SUCCESSFUL` (recent main runs were 36–45 minutes)
- [ ] If a test hangs, confirm JUnit/Gradle timeout is what fails the job rather than the 90-minute wall clock
- [ ] Spot-check job logs: TRACE session dumps from issuer/verifier/wallet tests should no longer dominate output


.github/workflows/gradle.yml
```diff
@@ -50,7 +50,12 @@ on:
 jobs:
   gradle:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 360
+    timeout-minutes: 90
+    env:
+      # Test klogging.conf defaults to TRACE and fills the 76MB GitHub log cap.
+      KLOGGING_MIN_LOG_LEVEL: INFO
+      KLOGGING_MIN_DIRECT_LOG_LEVEL: INFO
```
szijpeter: Non-blocking: these variables do not suppress application TRACE/DEBUG logs. `KLOGGING_MIN_LOG_LEVEL` controls Klogging’s internal diagnostics, while `KLOGGING_MIN_DIRECT_LOG_LEVEL` controls direct versus asynchronous delivery. I reproduced this with Klogging 0.11.8 and the verifier’s existing `klogging.conf`: with both variables set to `INFO`, TRACE and DEBUG messages still appear. Please configure the application logging ranges used by CI, or remove the claim that this change suppresses TRACE output (the same applies to the conformance job). This leaves an existing logging problem unresolved rather than introducing a functional regression. Reference: https://klogging.io/docs/internals/environment-variables/

szijpeter (approved):
Approved with a non-blocking logging comment. Please also update the PR summary: the final diff no longer contains the advertised 25-minute per-task timeout; it retains the 10-minute JUnit default and 90-minute job limit.

.github/workflows/gradle.yml
JakeFernandes98: Replaced those env vars with `KLOGGING_CONFIG_PATH` pointing at `.github/ci/klogging.conf`, which sets application ranges to INFO–WARN on stdout and ERROR+ on stderr. Klogging loads that file before classpath `klogging.conf`, so the TRACE test configs no longer apply in Gradle or conformance CI.

